PR #2, opened 2020-06-24
separate dropshot and dropshot_endpoint for publishing to crates.io

davepacheco:
**Background:** most of dropshot is (naturally) in the "dropshot" crate.  Since procedural macros must be in separate crates, "dropshot_endpoint" is also needed.  This is basically an implementation detail of "dropshot" (consumers would never use "dropshot_endpoint" directly), but as far as I can tell, the only way to publish "dropshot" is to publish "dropshot_endpoint" as well.

This change moves the "dropshot"-specific source, examples, and tests into a new top-level directory and creates a virtual manifest at the top level that just defines a Cargo workspace.  This is the same pattern we see for [serde](https://github.com/serde-rs/serde) and [schemars](https://github.com/GREsau/schemars).

@ahl can you give this a sanity check?  I don't think there's anything controversial here.  I did add a few metadata tags to dropshot_endpoint.

@steveklabnik @cbiffle if either of you have thoughts on the pattern used here for an internal dependency (used for a proc macro), let me know.

**Test notes:** besides the usual style and `cargo test` checks, I did `cargo publish --dry-run` in "dropshot_endpoint", and that works and includes what I expect.  See:

```
dropshot_endpoint $ cargo package --list
.cargo_vcs_info.json
Cargo.toml
src/lib.rs
```

For the "dropshot" subdirectory:

```
dropshot $ cargo package --list
.cargo_vcs_info.json
Cargo.lock
Cargo.toml
examples/basic.rs
src/api_description.rs
src/config.rs
src/error.rs
src/handler.rs
src/http_util.rs
src/lib.rs
src/logging.rs
src/router.rs
src/server.rs
src/test_util.rs
tests/test_config.rs
tests/test_demo.rs
```

`cargo publish --dry-run` fails in "dropshot" because "dropshot_endpoint" doesn't exist yet.

Once this lands, I'll publish "dropshot_endpoint", then "dropshot" to crates.io.


dropshot_endpoint/Cargo.toml
```diff
@@ -12,6 +15,6 @@ proc-macro2 = "1"
 quote = "1"
 serde = { version = "1.0", features = [ "derive" ] }
 serde_derive_internals = "0.25"
-serde_tokenstream = { git = "https://github.com/oxidecomputer/serde_tokenstream" }
+serde_tokenstream = { version = "0.1.0", git = "https://github.com/oxidecomputer/serde_tokenstream" }
```
ahl, line 18: I could imagine just having the version here and delete the git part; I don't think there's a strong argument either way. If it's just version then it will make development a little tricky for interlocking changes, but then we haven't had those in a while. If we have git in there too we'll just need to remember to publish that version if we want to update dropshot which might be kind of a pain.

davepacheco: According to the cargo output, the git version is removed before it gets published.  From a previous commit where I didn't have the "version" field:

```
dropshot_endpoint $ cargo publish --dry-run
    Updating crates.io index
error: all dependencies must have a version specified when publishing.
dependency `serde_tokenstream` does not specify a version
Note: The published dependency will use the version from crates.io,
the `git` specification will be removed from the dependency declaration.
dropshot_endpoint $ 
```


So I don't think having the "git" field there makes things much easier than if you added your own "patch" section to Cargo.toml or replace the "version" with a "git" URL while you're working on it.  Given that, I'm tempted to just drop the git part.  What do you think?

steveklabnik: I would drop the git part.

Cargo.toml
```diff
@@ -1,42 +1,8 @@
...
-
-[dev-dependencies]
-libc = "0.2.71"
-
-[dependencies.slog]
-version = "2.5"
-features = [ "max_level_trace", "release_max_level_debug" ]
-
+#
+# Dropshot is comprised of two crates.  Consumers only need to know about
```
steveklabnik, line 2:
```suggestion
# Dropshot is comprised of two packages.  Consumers only need to know about
```

This is not the most important difference, but while I'm here...

Cargo.toml
```diff
@@ -1,42 +1,8 @@
...
+# "dropshot".  "dropshot" depends on "dropshot_endpoint", a procedural macro
```
steveklabnik, line 3: This pattern is good. *technically* this should be called `dropshot-endpoint`, but that convention isn't followed very strictly, so....

dropshot/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+serde_urlencoded = "0.6"
+slog-async = "2.4"
+slog-bunyan = "2.2"
+slog-json = "2.3"
+slog-term = "2.5"
+tokio = { version = "0.2", features = [ "full" ] }
+toml = "0.5.6"
+uuid = { version = "0.8", features = [ "serde", "v4" ] }
+
+[dependencies.slog]
```
steveklabnik, line 33: So, I prefer this style, but we should pick either one or the other. That is, this and, say, the UUID example above. They do the same things two different ways. Consistency is more important than which one is picked.

dropshot/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+
+[dependencies]
+async-trait = "0.1.24"
+bytes = "0.5.4"
+chrono = { version = "0.4", features = [ "serde" ] }
+dropshot_endpoint = { version = "0.1.0", path = "../dropshot_endpoint" }
+futures = "0.3.1"
+http = "0.2.0"
+hostname = "0.3.0"
+hyper = "0.13"
```
steveklabnik, line 18: For consistency's sake, IMHO we should either do

    "0.13"

or

    "0.13.0"

consistently across all deps. They're equivalent, so which one we choose doesn't matter much.

dropshot_endpoint/Cargo.toml
```diff
@@ -1,8 +1,11 @@
 [package]
 name = "dropshot_endpoint"
+description = "macro used by dropshot consumers for registering handlers"
 version = "0.1.0"
 authors = ["Adam H. Leventhal <ahl@oxide.computer>"]
 edition = "2018"
+license = "Apache-2.0"
```
steveklabnik, line 7: I mentioned this on chat, but this means this package is GPL incompatible. Standard Rust licensing is `Apache-2.0/MIT` for this reason. I don't know how folks like @bcantrill feel about that.

dropshot_endpoint/Cargo.toml
ahl: yes, let's drop it. If it turns out there are a bunch of interlocking changes then we can reconsider.

dropshot_endpoint/Cargo.toml
ahl: I'd suggest we decouple licensing from this PR.

It's currently documented here https://github.com/oxidecomputer/meta/blob/51da81dafc71623412f526ac1a7778e712ef044e/engineering/code.md#open-source-licensing. @steveklabnik if you think we should reconsider, I'd suggest we revisit the topic in chat. It's unlikely that we're going to get a barrage of contributors from whom we'd need to extract permission to relicense dropshot if we change our position after that discussion.

steveklabnik: great!

Cargo.toml
davepacheco: Interesting!  Thanks for pointing it out.  I just looked up the definitions in the [Cargo book glossary](https://doc.rust-lang.org/cargo/appendix/glossary.html), and now it feels awfully confusing that it's called `crates.io`!

dropshot_endpoint/Cargo.toml
davepacheco: Fixed in 584593e8a05e7a80e9b1de1d6dd7022883156dab.

Cargo.toml
davepacheco: Fixed in 584593e8a05e7a80e9b1de1d6dd7022883156dab.

dropshot/Cargo.toml
davepacheco: Will look into this in a follow up (this was there before this change).

dropshot/Cargo.toml
davepacheco: Will look into this in a follow up (this was there before this change).

davepacheco:
Thanks both for taking a look!

dropshot/Cargo.toml
davepacheco: https://github.com/oxidecomputer/dropshot/pull/3

dropshot/Cargo.toml
davepacheco: It wasn't obvious to me that those would be equivalent, so I went looking for confirmation in a primary source.

[Cargo says it uses SemVer](https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field).  The [detailed docs on specifying dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) describe semvers with carets, tildes, and wildcards -- none of which are part of the [SemVer spec](https://semver.org/), though they're obviously widely used in similar contexts (thanks to npm, I believe).  The [SemVer BNF](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions) says a semver *must* have a major, minor, and patch number -- you can't leave out the patch or minor numbers -- but as we said, the spec doesn't actually describe what can go in this field.

[The semver package that npm uses documents that in the specific case of hyphen ranges, you can leave off the trailing parts of the version number (it calls these "partial versions") and they'll be filled in with zeros. ](https://www.npmjs.com/package/semver#hyphen-ranges-xyz---abc)  I think there are cases where npm supports these partial versions that aren't documented there (e.g., as a plain package dependency, just like we're talking about here, not inside an explicit range).

I've obviously gone overboard here but I can't find anything that says these are the same.  I'm going to prefer the full written out version for consistency.

steveklabnik: I'll move my reply to #3 :)

PR #3, opened 2020-06-25
change Cargo.toml dep style

davepacheco:
@steveklabnik is this what you had in mind in terms of the style?  It feels a lot less readable to me.  The alternative I considered was to stick with the inline table style, but that gets pretty wide for some of the deps and it sounded like you preferred this one?

### DIFF
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -7,32 +7,77 @@ edition = "2018"
 license = "Apache-2.0"
 repository = "https://github.com/oxidecomputer/dropshot/"
 
-[dependencies]
-async-trait = "0.1.24"
-bytes = "0.5.4"
-chrono = { version = "0.4", features = [ "serde" ] }
-dropshot_endpoint = { version = "0.1.0", path = "../dropshot_endpoint" }
-futures = "0.3.1"
-http = "0.2.0"
-hostname = "0.3.0"
-hyper = "0.13"
-indexmap = "1"
-openapiv3 = "0.3"
-schemars = "0.7"
-serde = { version = "1.0", features = [ "derive" ] }
-serde_json = "1.0"
-serde_urlencoded = "0.6"
-slog-async = "2.4"
-slog-bunyan = "2.2"
-slog-json = "2.3"
-slog-term = "2.5"
-tokio = { version = "0.2", features = [ "full" ] }
-toml = "0.5.6"
-uuid = { version = "0.8", features = [ "serde", "v4" ] }
+[dependencies.async-trait]
+version = "0.1.24"
+
+[dependencies.bytes]
+version = "0.5.4"
+
+[dependencies.chrono]
+version = "0.4"
+features = [ "serde" ]
+
+[dependencies.dropshot_endpoint]
+version = "0.1.0"
+path = "../dropshot_endpoint"
+
+[dependencies.futures]
+version = "0.3.1"
+
+[dependencies.http]
+version = "0.2.0"
+
+[dependencies.hostname]
+version = "0.3.0"
+
+[dependencies.hyper]
+version = "0.13"
+
+[dependencies.indexmap]
+version = "1"
+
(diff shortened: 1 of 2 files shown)


ahl:
I value the consistency we're shooting for here, but--inconsistent as it is--I like the pragmatic choice you had to begin with: mostly preferring the terse, 1-line syntax except when the list of features ran off the edge of the screen. To be clear, I'm not arguing that some option is *better* in absolute terms; rather, I'm expressing a preference for terseness and convenience over consistency.

davepacheco:
In case it wasn't obvious: the reason "slog" was previously done differently from all the others is because if done inline, it exceeded our standard 80 column limit.  (i.e., it wasn't arbitrary -- not to say it was a good reason)

steveklabnik:
Ah! This is not *exactly* what I meant. What I personally do is:

1. everything that depends on just a version goes under dependencies
2. everything that's more complex gets its own line

an example:

```toml
[dependencies]
foo = "1.2.3"
bar = "2.3.4"

[dependencies.special]
version = "3.4.5"
features = ["a", "b"]

[dependencies.second-special]
version = "4.5.6"
features = ["c", "d"]
```

*or*

```toml
[dependencies]
foo = "1.2.3"
bar = "2.3.4"
special = { version = "3.4.5", features = [ "a", "b" ] }
second-special = { version = "4.5.6", features = [ "c", "d" ] }
```

but *not*

```toml
[dependencies]
foo = "1.2.3"
bar = "2.3.4"
special = { version = "3.4.5", features = [ "a", "b" ] }

[dependencies.second-special]
version = "4.5.6"
```

... does that make any sense?

dropshot/Cargo.toml
```diff
@@ -7,32 +7,77 @@ edition = "2018"
...
+version = "0.2.0"
+
+[dependencies.hostname]
+version = "0.3.0"
+
+[dependencies.hyper]
+version = "0.13"
+
+[dependencies.indexmap]
+version = "1"
```
steveklabnik, line 37: Continuing from https://github.com/oxidecomputer/dropshot/pull/2#discussion_r445747402

To not toot my own horn too much, I am the maintainer of the semver package Cargo uses, *and* a maintainer of the semver spec. This doesn't mean I'm infallible of course...

The issue here is exactly what you identified, @davepacheco :

> but as we said, the spec doesn't actually describe what can go in the field we're talking about.

What goes here is a *range*, not a semver *version*. The Rust semver package currently calls them "VersionReq"s, but I'm changing that soon. Anyway, they *do* have a completely different grammar and semantic from versions, which must contain all three numbers. 

I actually have a personal TODO this evening to review https://github.com/semver/semver/pull/584, which is an attempt to specify ranges properly. Since I haven't reviewed it, I can't cite it just yet :)

For another empiric test, you can go to https://semver.npmjs.com/ and try both `^1` and `^1.0.0`, and see that they match the same versions. (Cargo, unlike npm, assumes ranges with no operation are `^` not `=`.)

steveklabnik: oh and anyway, since they're equivalent, it doesn't matter which form is used, but for consistency's sake, I personally write out all three numbers. but dropping trailing zeros would also be consistent, and consistency is the high order bit here imho

davepacheco:
> Ah! This is not exactly what I meant

Ah, got it.  That makes more sense.  Yes -- it sounds like we were both being consistent, but with a different criteria for switching to the expanded form...so inconsistent from the other's point of view!  Revised in 9b1251dee1858db581ba66fc1bdd1403dc5c93f2, where I've also added trailing minor and patch numbers to all version numbers.

dropshot/Cargo.toml
davepacheco: > (Cargo, unlike npm, assumes ranges with no operation are `^` not `=`.)

That's precisely what led to my confusion.  I'm glad to hear it's being specified!  In the latest commit I added trailing minor and patch numbers everywhere.

PR #4, opened 2020-06-26
Get clippy clean

steveklabnik:
This PR fixes a number of warnings caused by Clippy. I've split them into individual commits so that we can talk about them individually. I didn't add clippy to CI, but we may want to consider it.

### DIFF
--- a/dropshot/src/api_description.rs
+++ b/dropshot/src/api_description.rs
@@ -130,6 +130,17 @@ pub struct ApiDescription {
     /** In practice, all the information we need is encoded in the router. */
     router: HttpRouter,
 }
+pub struct OpenApiOptions<'a> {
+    title: &'a dyn ToString,
+    description: Option<&'a dyn ToString>,
+    terms_of_service: Option<&'a dyn ToString>,
+    contact_name: Option<&'a dyn ToString>,
+    contact_url: Option<&'a dyn ToString>,
+    contact_email: Option<&'a dyn ToString>,
+    license_name: Option<&'a dyn ToString>,
+    license_url: Option<&'a dyn ToString>,
+    version: &'a dyn ToString,
+}
 
 impl ApiDescription {
     pub fn new() -> Self {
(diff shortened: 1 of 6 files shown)


steveklabnik:
Oh, I should also mention that I'd like to change 5a7119f further, but I focused purely on fixing lints, rather than larger changes, since there's enough of them here already!

ahl:
@davepacheco and I looked through the kibitzing from clippy. We discussed whether maintaining clippy cleanliness was a goal and decided that it wasn't (your input here would be very valuable). We fixed the clippy lint that seemed beneficial and left the stuff that didn't. For example, I left the `starts_with(" ")` rather than converting it to a char because of the `starts_with("  ")` on the next line, and I left the parameters to `print_openapi` because adding a struct didn't seem to have any tangible benefit (though it may as we learn common patterns for callers).

If we decide to maintain clippy cleanliness, then we should bake that into CI. Our "for now" decision had been to run clippy periodically to look for useful guidance (there was plenty in that first pass--you can see in the git history).

steveklabnik:
> We discussed whether maintaining clippy cleanliness was a goal and decided that it wasn't (your input here would be very valuable). 

Historically, I have found clippy to be not ideal, but over time, they've tweaked things, and I've warmed to it. The above is all stuff I would actually merge, personally, and so lately I've been more pro keeping clippy clean.

> For example, I left the starts_with(" ") rather than converting it to a char because of the starts_with(" ") on the next line

Yeah I mean, this gets into the weeds a lot, but the char version optimizes a bit better: https://godbolt.org/z/8C_rWD

It really depends on what you're going for. I personally choose the "use the smallest thing that works", that would be, char, given that it's a single character, whereas you're choosing "visual consistency." (part of the reason to use the smallest (I'm not sure I like using this word this way, I hope it makes sense) choice is for optimization/performance reasons, but I also just generally think it's a good habit.)

>  I left the parameters to print_openapi because adding a struct didn't seem to have any tangible benefit (though it may as we learn common patterns for callers).

Yeah, so in general, the end goal here, IMHO, would be convert to the builder pattern. This is probably a similar kind of thing, I just don't tolerate this many arguments to functions as a general rule, because it's far too easy to make mistakes. Having names to go with that wall of `None`s is super helpful to me. It seems like it's just not for you, which is also fine, but that's the root of it.

davepacheco:
Thanks for this, @steveklabnik, and sorry it took so long to get around to this.  I thought it would be easier to re-do this (I'm not sure that turned out to be true) and wound up integrating a bunch of this under #84.  The pieces here were very helpful, e.g., for the warning in the handler macros about depending on order of evaluation rules.

Issue #9, opened 2020-06-26
support for paginated endpoints


ahl:
We should have first-class support for paginated endpoints, i.e. endpoints that enumerate lists of items and return a page at a time of content. We want to make it easy for developers to write these paginated endpoints without too much additional boiler plate.

In addition, it would be nice to have a generic manifestation of these paginated endpoints in the openapi doc. While OpenAPI doesn't have first-class support for pagination, we can potentially provide hints in the extensions (e.g. `x-dropshot-paginated`) that we could use in reference doc generation as well as client library code gen (both would require modifications to existing tools).

davepacheco:
Adding a bunch of my own notes on this.

- The endpoints that do this today all return an `HttpResponseOkObjectList`, which is intended to encapsulate this sort of thing.  It might need more context to fully do that (e.g., how many items are in the paginated set, how to construct the marker, etc.)
- The endpoints that do this today all consume a `Query<PaginationParams<...>>` and the pagination field is usually `ApiName`.  (One paginates by uuid today.)  This too is intended to encapsulate the input side of this use case.
- Internally in oxide-api-prototype, these endpoints are implemented using a common set of function (`collection_page()` and related functions).

The net result is that a lot of this is factored to be first-class already, though it maybe isn't quite there.

On my TODO list around this includes:

- (probably) swap NDJSON for an actual JSON object with fields for the marker, total_size, etc.
- if we can make this more first-class, it would be good to enforce limits on page size from a central place
- should the marker be an opaque token?  should it be encrypted?  ([Google does do this.](https://cloud.google.com/apis/design/design_patterns#list_pagination))  Adam and I talked about having a v1 that's not encrypted.  It would be pretty easy to re-do this provided that we put some recognizable version number in the representation (e.g., prepend "v1:" prior to base64-encoding).

davepacheco:
See #14.

By the way, I read the Google example too quickly -- they only encrypt when there's sensitive information in the token.

PR #14, opened 2020-07-16
Pagination

davepacheco:
This PR is a *very rough* draft implementation of pagination for Dropshot.  Briefly: pagination is the API design pattern by which API endpoints (e.g., HTTP resources) that list large collections of objects limit the number of items you can list at a time, forcing clients to make several requests to fetch the complete list.  There are lots of good reasons to do this.  These are documentation in dropshot/src/pagination.rs under the heading "Background: Why paginate HTTP APIs?".

This has taken a while and it's a lot more complex than I expected.  There's definitely more work to be done, but before I go much deeper, I'd like some feedback on the Rust API.  Whether it's a thumbs-up, thumbs-down, or concrete suggestions for improvement, I'd like to know what you think!

----

An important opening question is: why put any of this in Dropshot at all?  As far as I can tell, other popular Rust frameworks don't have pagination as a first-class thing.  The current Oxide control plane prototype implements paginated resources without any specific support from Dropshot.  But this leads to a few issues:

- There's nothing to ensure (or even help people ensure) that they paginte _at all_.  In an ideal world, any time somebody wanted to return a collection of things, they'd reach for a Dropshot abstraction that makes it paginated by default.  This is necessary for building scalable, highly-available APIs, and getting this wrong early can require breaking API changes to fix, so it's helpful if we can guide people early.
- There's nothing to ensure (or even help people ensure) that the paginated endpoints look the same in terms of query parameters and returned structures.  Again, getting this wrong either becomes cruft a client has to know about or cruft the server has to deal with for compatibility reasons (or both).
- There's nothing to guarantee that the various manually-paginated endpoints behave correctly with respect to pagination.  For example, there'd be nothing to prevent one of them from forgetting to check "limit" and returning some huge number of results.  For a consumer that wants to make sure this works correctly, they'd have to test this behavior for every single paginated API.

Now, at the REST level, there are many different ways of doing pagination.  These too are discussed in the block comment in dropshot/src/pagination.rs, under the heading "Patterns for pagination."  There's also a range of complexity needed by clients: the simplest case is a collection of items with a single field you can sort by (e.g., "id", a unique integer, or "name", a unique string).  A more complicated case would be a collection of items with several combinations (e.g., "name", a unique string, in either ascending or descending order; or a combination of "mtime" and "name").  I think these more complicated cases are likely more common than they sound.

----

That background aside, here's how I'd probably go about looking at this.  In this PR I've implemented two pagination APIs: a general one and a simpler "marker" one.

- Start with `examples/pagination-marker.rs`.  This uses the "marker" interface to paginate a collection of Projects.
- Check out `examples/pagination-multi.rs`.  This is an example of the more complicated case.  The documentation at the top of this file also goes into detail about how everything looks to an HTTP client (e.g., the query params and responses).
- Check out `examples/pagination-multi-resources.rs`.  This is a more realistic example of an API with multiple resources with similar pagination endpoints.  This use-case exposed some ugliness in the previous revisions of this interface.
- Check out `examples/pagination-basic.rs` (again, sorry for the dubious naming here).  This uses the general API to paginate in the same way as the marker example does.  I included this to show what the simple one is doing under the hood.

If you want, check out `dropshot/src/pagination.rs` to see how most of it's implemented.  There are also some bits in `dropshot/src/handler.rs`.

Here are my own thoughts about this right now:

- I think this works and does what I want, and I think it's close to a reasonable level of boilerplate / ergonomics.  Can we do better?
- Does the pattern in examples/pagination-multi-resources.rs make sense?  Should callers use a procedural macro where we currently define a macro rule?
- of particular note for @ahl : the openapi generation does not seem to include any of the information about these query parameters. I haven't dug into why not yet.

I'm interested in any and all feedback, especially on those questions.  I really want this to be something easy for people to reach for whenever they build an API that returns a list of things so that pagination becomes the default.  But I also want to ensure as much as possible that the resulting APIs are correct, robust, use scalable patterns, etc.

----

History and status:

2020-07-16 (commit 1791d7a9c2afe1d466feff1cdfb368fb92f0cffd): initial PR

2020-07-17 (commit 1e91d3c7bbca159385d32d6746086946656b8a14):
* renamed "simple" interface to "marker" and simplified the interface a bunch.  Renamed the examples to match.
* removed `page_selector_for` from the `PaginatedResource` trait.  Instead, callers pass a callback where this would go.  This seems much more convenient for both simple and more complicated consumers.

2020-07-17 (commit f4d5e850c6bcdea2ba5378b459136ba983fed67c): Remove `PaginatedResource` altogether.  This removed some boilerplate from the more complicated examples.

2020-07-31 (commit fa8865a5ea484c9542ee7c3e16131b4211abec5c): Lots of changes, largely to simplify the interface while also making it more flexible.  See [my comment on the changes](https://github.com/oxidecomputer/dropshot/pull/14#issuecomment-667422536).

2020-08-05 (commit 9d6347b4745ae4f85141473b5d693e8f7025cbbc): Added automated testing and removed `Incoming` pattern from examples.

2020-08-05 (commit 7ba2ee2f409833bda1a909b96880be0a8781f1e2): Sync'ed up with "master" branch (force push, unfortunately), added an openapi test, and fixed style.


davepacheco:
Regarding the openapi issue that I mentioned: take the example of "pagination-multi.rs":

- the first page in any scan has an optional "list_mode" querystring parameter which may have values "by-name-ascending", "by-name-descending", "by-mtime-ascending", and "by-mtime-descending".  In Rust, this corresponds to the `ProjectScanMode` type, which is an enum.
- subsequent pages in the scan have a required "page_token" querystring parameter.  This is a string.  In Rust, this corresponds to `PageToken`, though this struct is defined with `#[serde(try_from = "String")]`, which is how we deserialize a struct from a String.
- all pages in the scan have an optional "limit" parameter of type `NonZeroU64`

I modified the example to dump the openapi spec:

```
diff --git a/dropshot/examples/pagination-multi.rs b/dropshot/examples/pagination-multi.rs
index 1feaf21..5bef672 100644
--- a/dropshot/examples/pagination-multi.rs
+++ b/dropshot/examples/pagination-multi.rs
@@ -346,6 +346,8 @@ async fn main() -> Result<(), String> {
         .map_err(|error| format!("failed to create logger: {}", error))?;
     api.register(example_list_projects).unwrap();
+    api.print_openapi(&String::from("example"), None, None, None, None, None,
+        None, None, &String::from("1.0.0"));
     let mut server = HttpServer::new(&config_dropshot, api, ctx, &log)
         .map_err(|error| format!("failed to create server: {}", error))?;
     let server_task = server.run();
```

and here's what it looks like:

```json
{
  "openapi": "3.0.3",
  "info": {
    "title": "example",
    "contact": {},
    "version": "1.0.0"
  },
  "paths": {
    "/projects": {
      "get": {
        "description": "API endpoint for listing projects\n\n This implementation stores all the projects in a BTreeMap, which makes it\n very easy to fetch a particular range of items based on the key.",
        "operationId": "example_list_projects",
        "parameters": [
          {
            "in": "query",
            "name": "page_params",
            "description": "Specifies either how the client wants to begin a new scan or how to\n resume a previous scan.  This field is flattened by serde, so see the\n variants of [`WhichPage`] to see which fields actually show up in the\n serialized form.",
            "required": true,
            "schema": {
              "type": "string"
            },
            "allowReserved": true,
            "style": "form"
          },
          {
            "in": "query",
            "name": "limit",
            "description": "Optional client-requested limit on page size.  Consumers should use\n [`RequestContext::page_limit()`] to access this value.",
            "required": true,
            "schema": {
              "type": "string"
            },
            "allowReserved": true,
            "style": "form"
          }
        ],
        "responses": {
          "200": {
            "description": "TODO: placeholder",
            "content": {
              "application/json": {
                "schema": {
                  "description": "Client's view of a page of results from a paginated API",
                  "type": "object",
                  "properties": {
                    "items": {
                      "description": "list of items on this page of results",
                      "type": "array",
                      "items": {
                        "$ref": "#/components/schemas/Project"
                      }
                    },
                    "next_page": {
                      "description": "token to be used in pagination params for the next page of results",
                      "type": "string"
                    }
                  },
                  "required": [
                    "items"
                  ]
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Project": {
        "description": "Item returned by our paginated endpoint\n\nLike anything returned by Dropshot, we must implement `JsonSchema` and `Serialize`.  We also implement `Clone` to simplify the example.",
        "type": "object",
        "properties": {
          "mtime": {
            "type": "string",
            "format": "date-time"
          },
          "name": {
            "type": "string"
          }
        },
        "required": [
          "mtime",
          "name"
        ]
      }
    }
  }
}
```

There are a couple of problems here:

- "limit" should probably be some numeric type instead of "string".  Ideally, it would say it's a positive integer.  I'm not sure if that's because I used `NonZeroU64` instead of `u64`?
- both "limit" and the other query parameters should be optional
- instead of either "list_mode" or "page_token", we have "page_params".  This is probably because of the way I've implemented `PaginationParams`: `page_params` _is_ the right field in the Rust struct, but it's got `#[serde(flatten)]`, so it's not present on serialization/deserialization.

ahl:
I'm pretty sure the openapi oddness is related to some shortcuts I took in the parameter processing code. I think we're likely falling into this case: https://github.com/oxidecomputer/dropshot/blob/master/dropshot_endpoint/src/lib.rs#L186

dropshot/src/pagination.rs
```diff
@@ -0,0 +1,569 @@
...
+     * serialized form.
+     */
+    #[serde(flatten)]
+    pub page_params: WhichPage<ScanMode, PageSelector>,
+
+    /**
+     * Optional client-requested limit on page size.  Consumers should use
+     * [`dropshot::RequestContext::page_limit()`] to access this value.
+     */
+    pub(crate) limit: Option<NonZeroU64>,
```
ahl, line 349: should the limit have a limit? We can express this in OpenAPI. I do wonder if there's a single limit that makes sense for all paginated items (100? 200? 1000?) or if we'd want to let different items have different limits (seems like just more rope to hang ourselves with).

dropshot/src/pagination.rs
```diff
@@ -0,0 +1,569 @@
...
+     * `PageSelector` type.  Critically, this must provide enough information
+     * for the consumer to resume a scan.
+     */
+    NextPage { page_token: PageToken<PageSelector> },
+
+    /**
+     * Indicates that the client is beginning a new scan.  The client may
+     * provide `list_mode`, a consumer-specified type.
+     */
+    FirstPage { list_mode: Option<ScanMode> },
```
ahl, line 387: I don't have an opinion, but why did you choose to have these as named elements rather than the tuple form i.e.
```rust
NextPage(PageToken<PageSelector>),
FirstPage(Option<ScanMode>),
```
Is it to affect the encoding?

dropshot/src/pagination.rs
```diff
@@ -0,0 +1,569 @@
...
+                        "failed to serialize marker: {}",
+                        e
+                    ))
+                })?;
+
+            base64::encode_config(json_bytes, URL_SAFE)
+        };
+
+        /*
+         * TODO-robustness is there a way for us to know at compile-time that
```
ahl, line 507: hah.. this TODO is exactly the question I was going to ask. How would we feel about just applying a SHA hash to larger tokens automatically? Is that too cute?

dropshot/src/handler.rs
```diff
@@ -806,43 +856,83 @@ impl<T: JsonSchema + Serialize + Send + Sync + 'static>
...
 {
-    fn from(list_wrap: HttpResponseOkObjectList<T>) -> HttpHandlerResult {
-        let list = list_wrap.0;
-        let buffer_list = list.iter().map(serialize_json_stream_element);
-        let mut bytebuf = BytesMut::new();
-        for maybe_buffer in buffer_list {
-            let buffer = maybe_buffer?;
-            bytebuf.put(buffer);
-        }
+    pub fn new_for_marker<F, S>(
```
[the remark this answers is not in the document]
davepacheco: Pretty much -- you basically skip having to define a one-item `ScanMode` enum and a one-item `PageSelector` enum, and your function for mapping an item to a page selector is a tiny bit simpler (doesn't take an unused ScanMode argument).

dropshot/src/handler.rs
```diff
@@ -806,43 +856,83 @@ impl<T: JsonSchema + Serialize + Send + Sync + 'static>
...
+            items,
+        }))
+    }
 
-        Ok(Response::builder()
-            .status(HttpResponseOkObjectList::<T>::STATUS_CODE)
-            .header(http::header::CONTENT_TYPE, CONTENT_TYPE_NDJSON)
-            .body(bytebuf.freeze().into())?)
+    pub fn new_with_paginator<F, ScanMode, PageSelector>(
+        items: Vec<I>,
```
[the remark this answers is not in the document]
davepacheco: Yeah, it'd be neat to just accept an iterator, although we might need to `::collect()` it into a Vec in order to serialize it.

I like the idea of putting the `take(limit)` in here, but I think the most common use case is probably that the list endpoint is backed by a database query and the consumer will want to put the limit into the SQL query.  Given that, is it worth trying to put this in here?  I'm not sure.  There is a question of how we want to handle the case where the caller provided more items than the limit.  We could panic with an explicit message, or produce a 500 server error, or silently drop the extra items.  If we put the `take(limit)` in here, then on the plus side, it will "just work" for examples like the ones in the repo and it won't be wrong for consumers that have already applied the limit by the time we get here.  On the other hand, it might be pretty confusing behavior that would be pretty tough to debug.

dropshot/Cargo.toml
```diff
@@ -25,6 +25,10 @@ slog-json = "2.3.0"
 slog-term = "2.5.0"
 toml = "0.5.6"
 
+[dependencies.schemars]
+version = "0.7.0"
+features = [ "chrono", "uuid" ]
```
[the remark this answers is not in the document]
davepacheco: Yes, I think that's right.  (Maybe it would be more correct to use these features only as a dev-dependency.  I'm not sure if that's possible.)

dropshot/src/pagination.rs
davepacheco: That'd be neat.  Any idea how to represent that in the source?

dropshot/src/pagination.rs
davepacheco: Yes.  This is how `list_mode` winds up being the name of the query parameter.  I don't think it's important for the `NextPage` variant because of the `#[serde(flatten)]`.

Related to this: the way these structures are set up feels a bit elaborate because of the way we're building on `Query` + serde to do all of the deserialization.  The requirements to make the query params feel idiomatic (i.e. `list_mode` or `page_token`) results in having this untagged serde enum where the order of fields matters, and it also results in consumers dealing with this extra `page_token` field when they pull the value out of `NextPage`, and that thing uses `deserialize_from`.

Another option would be to do more of this explicitly instead of relying on all these serde attributes.  Specifically, we could modify `Query` so that instead of taking a `Deserialize`, it consumes a new trait `FromQuerystring`.  Then we could `impl<T> FromQueryString for T: Deserialize` to cover all of the existing use cases today that use `Deserialize`.  For `PaginationParams`, we could implement a `FromQuerystring` that explicitly parses the string and fills in the structures we want.  This could let us change these variants to use the tuple form, which would be more concise when destructuring it, as well as eliminate the `page_token` indirection from the interface.

dropshot/src/pagination.rs
davepacheco: I don't follow how we'd use that (because there's no way to reverse the SHA hash on the way back in, right?)

davepacheco:
Thanks for taking a look!

dropshot/src/pagination.rs
```diff
@@ -0,0 +1,569 @@
...
+ * at least that there's a constant upper bound), then it may be possible to use
+ * a simple timeout scheme to cancel requests that are taking too long, as might
+ * happen if a malicious client finds some way to cause requests to hang or take
+ * a very long time.
+ *
+ * [1]: https://cloud.google.com/apis/design/design_patterns#list_pagination
+ * [2]: https://www.citusdata.com/blog/2016/03/30/five-ways-to-paginate/
+ */
+
+use crate as dropshot; // XXX needed for ExtractedParameter below
```
davepacheco, line 266: @ahl This was another thing I ran into that you might have an idea about.  I needed to use the `ExtractedParameter` proc macro in this file.  However, that macro spits out code that explicitly uses `dropshot` by name -- here's some `cargo expand` output:

```
    impl dropshot::ExtractedParameter for MarkerScanMode {
        fn metadata(
            _in: dropshot::ApiEndpointParameterLocation,
        ) -> Vec<dropshot::ApiEndpointParameter> {
            ::alloc::vec::Vec::new()
        }
    }
```

so that's why I've got `use crate as dropshot` here, but that feels weird.  Is there some other pattern for referring to this that doesn't require that the consumer has already `use`d the right thing?

dropshot/src/handler.rs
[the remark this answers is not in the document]
ahl: That's a good point that the SQL query is probably the right place to apply the limit. I agree that we should leave it as-is, and maybe keep an eye out as we build consumers for common patterns that we can simplify.

dropshot/Cargo.toml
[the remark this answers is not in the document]
ahl: I had the same thought, and no idea how that works...

dropshot/src/pagination.rs
ahl: We can do it with a custom `impl JsonSchema` like I did for `ApiName` here: https://github.com/oxidecomputer/oxide-api-prototype/blob/master/src/api_model.rs#L183

This is for a string rather than an integer, but I think you get the point.

I'd be happy to contribute that if it would be helpful.

dropshot/src/pagination.rs
ahl: Good point; I didn't think it through.

dropshot/src/pagination.rs
ahl: I see. While it leads to some ugly cases, it does seem like a nice short cut in terms of implementation *and* education to lean on serde here regardless.

dropshot/src/pagination.rs
ahl: So I built something to handle this:
https://github.com/oxidecomputer/dropshot/blob/master/dropshot_endpoint/src/lib.rs#L142

Here's an examples of its use:
https://github.com/oxidecomputer/dropshot/blob/master/dropshot/src/api_description.rs#L878

I think you can do e.g.
```rust
#[derive(Debug, Deserialize, ExtractedParameter)]
#[serde(rename_all = "kebab-case")]
#[dropshot(crate = crate)]
pub enum MarkerScanMode {
    ByMarkerAsc,
}
```

It's going to be a little annoying to do for all uses. As I've mentioned, I think `ExtractedParameter` is not going to end up being the right solution (e.g. as we handle some of the stuff it's punting on now, non-required parameters)

dropshot/Cargo.toml
```diff
@@ -9,14 +9,14 @@ repository = "https://github.com/oxidecomputer/dropshot/"
 
 [dependencies]
```
ahl, line 10: bump to version 0.3.0? (I bumped to 0.2.0 when I added support for `tags` in the `endpoint` metadata

dropshot/src/pagination.rs
davepacheco: Ah okay.  Right now it feels like `use crate as dropshot` is a cleaner workaround for this file -- does that seem reasonable?

I'd hoped [`$crate`](https://doc.rust-lang.org/1.5.0/book/macros.html#the-variable-crate) could help here, but when I tried it, I ran into the behavior [described here](https://stackoverflow.com/questions/48050155/is-it-possible-to-get-the-full-namespace-of-a-struct-in-a-custom-derive#comment83076899_48052098).

ahl: Yes; it feels a little janky, but less annoying than sprinkling that magic everywhere..

dropshot/Cargo.toml
davepacheco: Done.  (Looks like I'll need to bump it again when I merge with #16.)

dropshot/src/pagination.rs
davepacheco: For the record: we discussed this in person and decided it's not really part of the API contract, but a runtime parameter of the server.

dropshot/src/pagination.rs
davepacheco: FWIW, I ended up finding a middle ground here.  I created `RawPaginationParams` and `RawWhichPage` which use the serde trickery that we were doing before to parse stuff right off the wire.  Then I've implemented `PaginationParams` in a way that causes serde to first parse it as a `RawPaginationParams` and then use `TryFrom` to convert it into a `PaginationParams`.  This way, we still use the serde trickery that we had previously (so we don't have to implement `Deserialize` by hand), but we can expose a much cleaner structure to consumers.  More on this in another comment.

dropshot/Cargo.toml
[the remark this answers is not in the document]
davepacheco: I think I made this work correctly by duplicating `schemaras` it in dev-dependencies with a different set of features.

dropshot/src/handler.rs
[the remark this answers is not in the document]
davepacheco: I've deferred the question of enforcing the limit.  I will file an issue for it so we don't forget about it.

dropshot/examples/pagination-multi-resources.rs
```diff
@@ -0,0 +1,422 @@
...
+        impl HasIdentity for $T {
+            fn id(&self) -> &Uuid {
+                &self.id
+            }
+            fn name(&self) -> &String {
+                &self.name
+            }
+        }
+    };
+}
```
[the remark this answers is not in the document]
davepacheco: I did this in the oxide-api-prototype integration of these new dropshot interfaces.  More in my summary comment.

dropshot/examples/pagination-multi.rs
```diff
@@ -0,0 +1,483 @@
...
+            ProjectScanPageSelector::MtimeName(Ascending, ..) => {
+                ProjectScanMode::ByMtimeAscending
+            }
+            ProjectScanPageSelector::MtimeName(Descending, ..) => {
+                ProjectScanMode::ByMtimeDescending
+            }
+        },
+    };
+
+    let iter = match &pag_params.page_params {
```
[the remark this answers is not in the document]
davepacheco: Thanks.  It's surprisingly tricky to write code that deals with two different possible iterator types (e.g., an optional `map()` step).  I'm not sure how to do it without either a `Box` or a separate function to return the iterator reference.

davepacheco:
I just pushed a pretty big set of changes based on the feedback and more ideas I had to simplify the interface.  I don't think there's anything controversial here.

Below is a rough summary of changes between commits f4d5e85..fa8865a

On the input side:

- Previously, the only parameter consumers could accept for the first request in a scan was the "list_mode", which was expected to be an enum of things like "name-ascending".  Adam pointed out that you probably need to be able to apply filters here as well.  So `ScanMode` became `ScanParams`, and it's expected that consumers would use a structure type for this rather than an enum.
- I streamlined the `PaginationParams` / `WhichPage` interface a bit.  `WhichPage::{FirstPage,NextPage}` are now `{First, Next}`, and they're both tuple variants.  The consumer doesn't need to know about "page_token" any more.  (Internally, these structures are now decoupled from exactly what comes in from the querystring, so we have some flexibility to structure them however we want.)
- I got rid of the "Marker" interface altogether.  See `examples/pagination-basic.rs` for the simplest case now.  This eliminated the awkwardness around `HttpResponseOkPage::new_with_paginator/new_with_marker`.
- I added a test case where the consumer accepts non-pagination-related query parameters (e.g., "debug=true").  I expected this to "just work" by embedding `PaginationParams` inside a bigger structure and using `#[serde(flatten)]`, but this failed due to nox/serde_urlencoded#33, which is really serde-rs/serde#1183.  After an hour or so of panic thinking the whole design was sunk, I found that a consumer can simply have their handler function accept two different `Query` arguments.  (This will parse the query string twice, but that's arguably the most correct thing to do anyway.)

On the output side:

- I realized that consumers should be able to return their own structure that might include the list of returned items and a page token.  (For example, they might want to include other metadata like the total number of results.)  That's when I realized that we shouldn't conflate the `HttpResponseTyped` type with the actual data returned.  So I got rid of `HttpResponseOkPage` altogether and moved that functionality to a new type called `ResultsPage`.  Now, you return an `HttpResponseOkObject<ResultsPage<T>>`.  If you want, you can embed `ResultsPage` in your own type that has more stuff around it (and I've tested that).  This also eliminated the awkwardness that `HttpResponseOkPage` was the only one of the `HttpResponseTyped` structs that had a constructor.

Other stuff:

- I've added rustdoc comments for all the new stuff.
- I prototyped applying this to oxide-api-prototype.  It's a lot of boilerplate to support three different kinds of scan, but I'm otherwise happy with it (and critically the boilerplate is O(number of types of scan), not O(number of endpoints that use each type of scan)).  You can see this [here](https://github.com/oxidecomputer/oxide-api-prototype/compare/master...davepacheco:pagination).

What's still remaining:

- I want to write a complete battery of automated tests for all this.
- I need to sync up with other changes in master.
- There's a bunch more to wrap up on the oxide-api-prototype change, but none of it feels like a blocker to me.

There's other stuff I'm not planning to do now, but I will file issues about:

- We may want to provide a version of `ResultsPage` that includes the total result set size.  This should be a separate type so that the OpenAPI spec can reflect whether it's present or not.
- We still need to fix the openapi spec output for the query parameters.
- The "next_page" token currently shows up in `ResultsPage" as long as there are _any_ results on the page.  That's wrong -- it should only show up if there's a _complete_ list of results.  Unfortunately, we don't currently have access to the limit that was used here.  This would be good to flesh out, but it doesn't seem urgent and I'm getting itchy to get this integrated.
- Similarly, it would be good to have some policy around what to do when the consumer provides more results than we expected and then enforce that policy.  Example policies could include: panic, return a 500 error, log a warning, silently truncate the result set, allow it anyway.  This is blocked on the same thing: having the limit in the result set.
- We should provide some way for consumers to test the format of the page token.  This is an interface that they might not realize because it doesn't appear in the OpenAPI spec.
- The max and default page sizes should be really configurable, instead of hardcoded values. 

@ahl do you want to take another look?

dropshot/Cargo.toml
ahl: Meh... I now think we should just bump it when we push to crates.io, but up to you

ahl:
> I'm pretty sure the openapi oddness is related to some shortcuts I took in the parameter processing code. I think we're likely falling into this case: https://github.com/oxidecomputer/dropshot/blob/master/dropshot_endpoint/src/lib.rs#L186

I'll be interested to know if post-merging with #16 this changes.

dropshot/src/pagination.rs
```diff
@@ -0,0 +1,487 @@
...
+ *
+ * `PageSelector` describes the information your endpoint needs for requests
+ * after the first one.  Typically this would include an id of some sort for the
+ * last item on the previous page so that your function can resume the scan with
+ * the first item after that.  If your endpoint supports filters or multiple
+ * sort options, you probably want to include these, too.  The entire
+ * `PageSelector` will be serialized to an opaque string and included in the
+ * [`ResultsPage`].  The client is expected to provide this string as the
+ * `"page_token"` querystring parameter in the subsequent request.
+ * `PageSelector` must implement both [`Deserialize`] and [`Serialize`].
```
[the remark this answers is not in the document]
davepacheco: Updated in e7db0d83f304e20c0f241b6a80f9b396b600ba3f.  (It still gets deserialized indirectly from the querystring, of course.)

dropshot/src/pagination.rs
```diff
@@ -0,0 +1,487 @@
...
+ *
+ * `ScanParams` describes the set of querystring parameters that your endpoint
+ * accepts for the _first_ request of the scan (typically: filters and sort
+ * options).  This must be deserializable from a querystring.
+ *
+ * `PageSelector` describes the information your endpoint needs for requests
+ * after the first one.  Typically this would include an id of some sort for the
+ * last item on the previous page so that your function can resume the scan with
+ * the first item after that.  If your endpoint supports filters or multiple
+ * sort options, you probably want to include these, too.  The entire
```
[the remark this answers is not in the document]
davepacheco: Your edit sounds good.  I think a large fraction of our APIs are likely to have no filtering or sorting options, but that the remaining ones will probably have several, and they'll be important.

dropshot/src/lib.rs
```diff
@@ -275,6 +270,178 @@
...
+ * ) -> Result<HttpResponseOkObject<ResultsPage<String>>, HttpError>
+ * {
+ *  # unimplemented!();
+ *  /* ... */
+ * }
+ * ```
+ *
+ * You might expect that instead of doing this, you could define your own
+ * structure that includes a `PaginationParams` using `#[serde(flatten)]`, and
+ * this ought to work, but it currently doesn't due to serde_urlencoded#33,
```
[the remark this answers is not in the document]
davepacheco: Tell me about it!

dropshot/src/lib.rs
```diff
@@ -275,6 +270,178 @@
...
+ * }
+ *
+ * #[endpoint {
+ *     method = GET,
+ *     path = "/list_stuff"
+ * }]
+ * async fn my_list_api(
+ *     rqctx: Arc<RequestContext>,
+ *     pag_params: Query<PaginationParams<MyScanParams, MyPageSelector>>,
+ *     extra_params: Query<MyExtraQueryParams>,
```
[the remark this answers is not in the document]
davepacheco: I can definitely see how it's a little strange.  On the other hand, it would seem a little strange to me if it _didn't_ work -- that would imply that extractors somehow mutate the request or are otherwise intertwined instead of being self-contained things that construct a higher-level view of something in the request.

With the serde bug, I don't know another way to support taking non-pagination query parameters, which seems pretty useful.

I thought about whether we can avoid parsing the querystring twice.  Unfortunately, we don't know until we've parsed the querystring whether it's the `WhichPage::First` case, in which case _all_ of the query parameters should be deserialized into the user's struct.  But once we've parsed it to determine we're in the `WhichPage::First` case, we don't have a way to take the representation we've got and parse it into the user's type.  The only way we have to get to the user's struct is to parse the querystring again.  Alternatively, we could define an intermediate representation (e.g., `BTreeMap<String, String>`) and define a serde Deserializer for that representation that can reconstitute the user's type, but this doesn't feel worth the work right now just to avoid a second parse.

davepacheco:
> There seems to be a common pattern of the ExScanParamsIncoming turning into ExScanParams (with the difference being Optional<..>). Is there something clever we can do to obviate the need for that?

Consumers can do a few things:

1. Just use the `Incoming` version (they don't have to call it `Incoming` in that case) and deal with the fact that every where it appears, they have this `Option` that they need to either deal with or make assumptions about.  This is kind of annoying because there are a bunch of places you use these `ScanParams` and it's a lot cleaner and more robust if you've filled in the default once and you know you have a valid value everywhere else.

2. Do as I did in this example and define a separate type for `Incoming`, convert it to the non-optional type _once_ (filling in defaults at that point), and use that everywhere else.  This is kind of annoying for the extra type and the conversion. 

3. Implement `Default` for their `ScanParams` type or whatever members of it are optional.  This works fine but I'm a little skeptical of having a general `Default` implementation if we only want to apply default values when deserializing.  I think that's the only case where it comes up today, but if it came up in another case, I'd rather force us to think about whether we want Default values or whether the values should be coming from somewhere else explicitly.

4. Use `#[serde(default)]` to define a function that produces a default value when deseriailizing.  I did this in the oxide-api-prototype integration here:
https://github.com/davepacheco/oxide-api-prototype/blob/58ddf0a243555d169c66cf5e42d356e9e86f547f/src/http_pagination.rs#L69-L70

   This is kind of annoying for the extra attribute and function (but no moreso than an implementation of `Default` or the extra `Incoming` type).

They basically all seem fine and also imperfect.  I think (1) opens the door to various runtime bugs.  (2), (3), and (4) all get the job done, at the cost of some extra boilerplate, though that code has to exist _somewhere_.

I like (4) best, and I can do that in the example if that seems better.

davepacheco:
> Statically, I think we've moved to these endpoints being a little harder to identify.

I think that's right.  The only thing we'd know about them now is that they take a `Query<PaginationParams<...>>` (or, if that serde bug gets fixed, a `Query<X>` where X somehow includes a `PaginationParams)` and they return a `Y` that somehow includes a `ResultsPage`.  It would be nice if this were easier to identify.  From talking offline, maybe we can defer this for now.

> I think it might be nice to be able to provide some sort of compile-time check for these paginated endpoints to confirm that inputs and outputs jived.

Yeah, I'm interested in this as well.  I tried a bunch of things (described below for future context) and went in a bunch of circles as I tried to validate them against my design goals.  There may well be a way to do this with less boilerplate and/or better compile-time checking, and we should definitely iterate as we find better ways to do it.  From talking offline, we'll defer this for now.

----

The rest of this describes my goals and some of the things I tried for reference if we revisit this.  Some goals:

- support for multiple consumer-defined ways to paginate a resource (e.g., different sort orders, possibly on different combinations of fields)
- arbitrary consumer-defined first-page parameters (e.g., for filtering)
- arbitrary consumer-defined next-page parameters to support whatever approach the consumer wants for the marker
- as little boilerplate as possible for the simple case (no scan options, paginated by one marker field)
- as little boilerplate as possible for each new resource that shares pagination behavior with an existing one (e.g., if you already have N resources paginated by "name" in ascending order with no filtering params, you shouldn't need much boilerplate -- if any -- for another one that works that way too)
- an opaque token for the consumer that encapsulates whatever the server needs for the next page.  (This was modeled loosely after the Google pagination guidelines linked in #9, after surveying several cloud infrastructure APIs)
- support for consuming other querystring parameters not related to pagination (e.g., "debug=true" or something)
- support for returning other output in addition to the results (e.g., to include other metadata with the results, like maybe the total count of items or a request id)

My early intuition was that the page of results could be parametrized by the ScanMode and PageSelector types so that you could tell statically if, say, an endpoint told Dropshot to generate a page selector that was different than the one it accepted on input.  Similarly, it seems like there's some kind of checking you could do to make sure the page selector makes sense for the scan mode.

The very first version I had looked similar to what you describe:
https://github.com/davepacheco/dropshot/blob/1b7813e1df4cc23592ecde6bcf46016d4aaa3d27/dropshot/src/handler.rs#L862-L882

This was before the interface supported multiple different scan modes, so `for<'a> &'a ItemType: Into<MarkerFields>,` is analogous to your `PageSelector: From<(ItemType, ScanParams>`.  One of the next commits removed the `MarkerFields` (the analog at that time for `PageSelector`) as a type parameter, and I remember that being important, but I don't remember why :(

I don't think this necessarily helps match up the input and output unless there's something to make sure that the PageSelector incoming is the same one you use in the `ResultsPage`.  At one point, I had a trait `PaginatedResource`:
https://github.com/davepacheco/dropshot/blob/1e91d3c7bbca159385d32d6746086946656b8a14/dropshot/src/pagination.rs#L194

that was just the combination of ScanMode, PageSelector, and Item types.  When you called `new_with_paginator` to construct the results page, it's parametrized by `PaginatedResource`.  At least this way, you can't easily wind up using the wrong `PageSelector` because you'd have to pass an `impl` of `PaginatedResource`, which means you'd either have to implement a new one for no reason or you were passing one from a different handler function (e.g., copy/paste).  But if you passed a valid one from a different handler, its ScanMode and Item types would likely be different, and since you had to pass a scan_mode and list of items, and the item type also appears in handler function's signature, it just seemed unlikely you'd get all those things wrong and not notice.

Something I really didn't like about this approach was that you need a new `PaginatedResource` impl for every single paginated resource in the API.  That seems like a non-starter, particularly for an API with lots of resources that all paginate the same way (e.g., by "id" or "name").

davepacheco:
I've pushed a few more changes: one adds automated tests for most of the surface area here, and the other replaces the `Incoming` pattern in the examples with defaults provided by one-off functions (option 4 in [this comment](https://github.com/oxidecomputer/dropshot/pull/14#issuecomment-668152665).)

My plan next is to sync up, retest, potentially add some more openapi tests, and wrap up.

davepacheco:
I filed #19 #20 #21 #22 and #23.  I did not end up filing a ticket for an alternate version of `ResultsPage` with a field for the total number of results.  I figured consumers can build this very easily, and we can incorporate it into Dropshot if/when we find it useful.

PR #17, opened 2020-08-03
formatting with rustfmt-2.0.0-rc.2

ahl:
Not sure why I got this itch, but wanted to see what rustfmt 2 was looking like. I think it's basically sensible, but I'm not a fan of the change in `handler.rs`. @davepacheco let me know what you think and I can file a bug against rustfmt

To be clear: I don't intend to merge this PR.

### DIFF
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -570,7 +570,7 @@ where
  */
 trait GetMetadata {
     fn metadata(loc: ApiEndpointParameterLocation)
-        -> Vec<ApiEndpointParameter>;
+    -> Vec<ApiEndpointParameter>;
 }
 
 impl<ParamType> GetMetadata for ParamType
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -570,7 +570,7 @@ where
  */
 trait GetMetadata {
     fn metadata(loc: ApiEndpointParameterLocation)
-    -> Vec<ApiEndpointParameter>;
+        -> Vec<ApiEndpointParameter>;
 }
 
 impl<ParamType> GetMetadata for ParamType
--- a/dropshot/src/router.rs
+++ b/dropshot/src/router.rs
@@ -842,9 +842,11 @@ mod test {
             "project_id"
         ]);
         assert_eq!(result.variables.get("project_id").unwrap(), "p12345");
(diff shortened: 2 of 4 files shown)


dropshot/src/handler.rs
```diff
@@ -570,7 +570,7 @@ where
  */
 trait GetMetadata {
     fn metadata(loc: ApiEndpointParameterLocation)
-        -> Vec<ApiEndpointParameter>;
+    -> Vec<ApiEndpointParameter>;
```
steveklabnik, line 573: this.... seems like a bug to me

dropshot/src/router.rs
```diff
@@ -842,9 +842,11 @@ mod test {
...
         ]);
         assert_eq!(result.variables.get("project_id").unwrap(), "p12345");
-        assert!(router
-            .lookup_route(&Method::GET, "/projects/p12345/child")
-            .is_err());
+        assert!(
+            router
+                .lookup_route(&Method::GET, "/projects/p12345/child")
+                .is_err()
+        );
```
steveklabnik, line 849: I'm guessing this is the change youre not a fan of? because the other one actually looks good to me!

dropshot/src/handler.rs
ahl: This is the change I'm not a fan of; agreed that it probably should be filed as a bug.

dropshot/src/router.rs
ahl: Meh. I'm neutral on this one.

steveklabnik: ah, i mis-read `handler.rs` as `router.rs`, oops 😓 

(I am also neutral tbh)

dropshot/src/handler.rs
ahl: https://github.com/rust-lang/rustfmt/issues/4366

ahl: fixed: https://github.com/rust-lang/rustfmt/issues/4366

davepacheco:
Sorry I forgot about this one!  Looks like you got the handler.rs change fixed upstream -- nice!  Yeah, this basically seems fine.

ahl:
I'm going to close this PR since we're not actually switching to rustfmt 2

PR #18, opened 2020-08-04
try to give the user better failure messages

ahl:
I took a crack at producing less inscrutable error messages when people don't specify the right types for endpoints. I could use some input as to whether this is beneficial or just different.

Thanks.

### DIFF
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -30,7 +30,7 @@ version = "0.4.0"
 features = [ "serde" ]
 
 [dependencies.dropshot_endpoint]
-version = "0.2.0"
+version = "0.3.0"
 path = "../dropshot_endpoint"
 
 [dependencies.serde]
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -59,6 +59,7 @@ lazy_static = "1.4.0"
 libc = "0.2.71"
 serde_with = "1.4.0"
 subprocess = "0.2.4"
+trybuild = "1.0.31"
 
 [dev-dependencies.schemars]
 version = "0.7.0"
--- a/dropshot/tests/fail.rs
+++ b/dropshot/tests/fail.rs
@@ -1,0 +1,7 @@
+// Copyright 2020 Oxide Computer Company
+
+#[test]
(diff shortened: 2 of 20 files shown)


.github/workflows/rust.yml
```diff
@@ -17,11 +17,11 @@ jobs:
     - uses: actions/checkout@28c7f3d2b5162b5ddd3dfd9a45aa55eaf396478b
     - uses: actions-rs/toolchain@b223206e28798aa3c3668bdd6409258e6dc29172
       with:
-        toolchain: nightly-2020-06-02
+        toolchain: nightly-2020-07-26
```
davepacheco, line 20: Nit: do we need to change this, or is there some reason to do it now?  It'd just be nice to not have to change this all the time if we've got something that works.

dropshot/Cargo.toml
```diff
@@ -54,4 +54,6 @@ features = [ "uuid" ]
 
 [dev-dependencies]
 libc = "0.2.71"
-difference = "2.0.0"
+difference = "2.0"
```
davepacheco, line 57: Nit: I think in a previous PR @steveklabnik mentioned that "x.y.0" was the same as "x.y", which I think is because [Cargo implicitly uses the "^" syntax](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio).  I thought we decided to be consistent and always specify a patch version, even if it's zero.  WDYT?

steveklabnik: yes, these two lines are equivalent. I personally always put all three numbers, but the tools will support either of these until the end of all time

.github/workflows/rust.yml
ahl: I accidentally upgraded locally like a numpty. I just did it for me for the moment.

dropshot/Cargo.toml
ahl: did not know! thank you.

ahl:
@davepacheco this is a great checklist; thanks

ahl:
pausing on https://github.com/dtolnay/trybuild/issues/83

PR #31, opened 2020-08-07
try to shake out failure

ahl:


ahl:
fix from @davepacheco seems to have chased away the problem at least

PR #48, opened 2020-08-18
would like support for freeform request bodies

### LINKED ISSUE #44: would like support for freeform request bodies
I have an API where I am just accepting a freeform string in a POST.  I was able to read a capped amount of bytes, thus:

```rust
use hyper::body::HttpBody;

async fn read_body(rqctx: &Arc<RequestContext>, max: usize) -> Result<String> {
    let mut req = rqctx.request.lock().await;
    let body = req.body_mut();
    let mut out: Vec<u8> = Vec::new();

    while let Some(x) = body.data().await.transpose()
        .map_err(|e| anyhow!("body read: {}", e))?
    {
        if out.len().saturating_add(x.len()) > max {
            bail!("request size exceeded {} bytes", max);
        }

        out.append(&mut x.to_vec());
    }

    Ok(String::from_utf8(out)?)
}
```

It'd be neat if there was some way to get either a UTF-8 `String` or a slice or vector of `u8` (I think both will be useful at different times) through an extractor in the way that you can get a specific `Deserializable` today.

davepacheco:
This fixes #44.

I've added a new extractor called `UntypedBody`, which exposes:

- `as_bytes() -> &[u8]`: returns a byte slice
- `as_str() -> Result<&str, HttpError>`: returns a (UTF-8) `str` from the byte slice.  If the string is not valid UTF-8, then this bails with a 400 response saying that the body should be a UTF-8 string.

This approach is straightforward and seems convenient to me -- what do you think @jclulow and @ahl?  Note that from this, we cannot statically determine if the endpoint is _expecting_ a UTF-8 string or just an octet stream, so the JSON schema for this just says that it's a set of binary bytes.  If we want that, we'd need to separate this into two types, maybe `BodyBytes` and `BodyUtf8`?  I wasn't sure that was worth it here.

In terms of the implementation: I've piggy-backed off the existing `ApiEndpointParameterName::Body` to stash the expected content-type rather than more serious surgery to the api_description.rs.  I could separate this into something that doesn't overload the "name" if that seems better, or maybe just rename this to `ApiEndpointParameterMetadata`?  I defer to @ahl on this.

The OpenAPI schema looks like this:
https://github.com/oxidecomputer/dropshot/blob/95e1bfa92303f127779c4a4c229e9d603d5e3113/dropshot/tests/test_openapi.json#L12-L22

Edit: I forgot to mention that I chose that schema based on the suggestions in the [Considerations for File Uploads](https://swagger.io/specification/#considerations-for-file-uploads) section in the Swagger spec.

### DIFF
--- a/dropshot/src/api_description.rs
+++ b/dropshot/src/api_description.rs
@@ -12,6 +12,7 @@ use crate::router::HttpRouter;
 use crate::router::PathSegment;
 use crate::Extractor;
 use crate::CONTENT_TYPE_JSON;
+use crate::CONTENT_TYPE_OCTET_STREAM;
 
 use http::Method;
 use http::StatusCode;
@@ -77,7 +78,7 @@ impl<'a> ApiEndpoint {
  */
 #[derive(Debug)]
 pub struct ApiEndpointParameter {
-    pub name: ApiEndpointParameterName,
+    pub metadata: ApiEndpointParameterMetadata,
     pub description: Option<String>,
     pub required: bool,
     pub schema: ApiSchemaGenerator,
@@ -94,12 +95,12 @@ impl ApiEndpointParameter {
         examples: Vec<String>,
     ) -> Self {
         Self {
-            name: match loc {
+            metadata: match loc {
                 ApiEndpointParameterLocation::Path => {
-                    ApiEndpointParameterName::Path(name)
+                    ApiEndpointParameterMetadata::Path(name)
                 }
                 ApiEndpointParameterLocation::Query => {
-                    ApiEndpointParameterName::Query(name)
+                    ApiEndpointParameterMetadata::Query(name)
                 }
             },
             description,
@@ -110,13 +111,14 @@ impl ApiEndpointParameter {
     }
 
     pub fn new_body(
+        content_type: ApiEndpointBodyContentType,
         description: Option<String>,
         required: bool,
         schema: ApiSchemaGenerator,
         examples: Vec<String>,
     ) -> Self {
         Self {
-            name: ApiEndpointParameterName::Body,
+            metadata: ApiEndpointParameterMetadata::Body(content_type),
             description,
             required,
             schema,
@@ -132,10 +134,27 @@ pub enum ApiEndpointParameterLocation {
 }
 
 #[derive(Debug, Clone)]
-pub enum ApiEndpointParameterName {
+pub enum ApiEndpointParameterMetadata {
     Path(String),
     Query(String),
-    Body,
+    Body(ApiEndpointBodyContentType),
+}
+
+#[derive(Debug, Clone)]
+pub enum ApiEndpointBodyContentType {
+    /** application/octet-stream */
+    Bytes,
+    /** application/json */
+    Json,
+}
+
+impl ApiEndpointBodyContentType {
+    fn mime_type(&self) -> &str {
+        match self {
+            ApiEndpointBodyContentType::Bytes => CONTENT_TYPE_OCTET_STREAM,
+            ApiEndpointBodyContentType::Json => CONTENT_TYPE_JSON,
+        }
+    }
 }
 
 /**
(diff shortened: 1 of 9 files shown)


dropshot/src/handler.rs
```diff
@@ -888,6 +894,78 @@ where
...
+        Ok(UntypedBody {
+            content: body_bytes,
+        })
+    }
+
+    fn metadata() -> Vec<ApiEndpointParameter> {
+        let schema = SchemaObject {
+            instance_type: Some(InstanceType::String.into()),
+            format: Some(String::from("binary")),
+            ..Default::default()
```
ahl, line 955: neat.. I didn't know this syntax

dropshot/src/handler.rs
```diff
@@ -888,6 +894,78 @@ where
...
+        )
+        .await?;
+        Ok(UntypedBody {
+            content: body_bytes,
+        })
+    }
+
+    fn metadata() -> Vec<ApiEndpointParameter> {
+        let schema = SchemaObject {
+            instance_type: Some(InstanceType::String.into()),
```
ahl, line 953: well that's a lot simpler than what I did here: https://github.com/oxidecomputer/oxide-api-prototype/blob/master/src/api_model.rs#L182

love it

dropshot/src/api_description.rs
```diff
@@ -135,7 +137,24 @@ pub enum ApiEndpointParameterLocation {
 pub enum ApiEndpointParameterName {
     Path(String),
     Query(String),
-    Body,
+    Body(ApiEndpointBodyContentType),
```
ahl, line 140: "name" was already probably a stretch. If you can think of a better identifier here I think that would be an improvement.

dropshot/src/lib.rs
```diff
@@ -187,13 +187,14 @@
  *      [query_params: Query<Q>,]
  *      [path_params: Path<P>,]
  *      [body_param: TypedBody<J>,]
+ *      [body_param: UntypedBody<J>,]
```
ahl, line 190: nit:
```suggestion
 *      [body_param: UntypedBody,]
```

dropshot/src/api_description.rs
davepacheco: `ApiEndpointParameterMetadata`?

dropshot/src/lib.rs
davepacheco: Nice catch.  Fixed in eccf949da8c52d3d56278bb1ba8478d102695e82.

dropshot/src/handler.rs
davepacheco: Yeah, I saw this in one of the implementations in `schemars`.  I normally avoid `into()`.

dropshot/src/handler.rs
davepacheco: I'd forgotten it myself, but they use it in one of the implementations in schemars.

davepacheco:
Better compile-time checking for multiple body parameters would be good.  I'd like to defer that if that seems okay to you.

dropshot/src/api_description.rs
ahl: A nice catchall word! I'm good w/ it. Better than `ApiEndpointParameterThingamajigger`

ahl:
> Better compile-time checking for multiple body parameters would be good. I'd like to defer that if that seems okay to you.

Yes of course -- just an idea I had that I wanted to run by you.

davepacheco:
Commit 9a16c346c57d2178d52cd990037ebd9c74b9afa6 adds a runtime check for attempted use of multiple body extractors.  It's not quite as nice as compile-time, but assuming people `unwrap()` the return value of `register()`, this will fail pretty loudly at runtime.

dropshot/src/api_description.rs
davepacheco: Fixed in 409e18a44093ddf6f7a9b0efa53f2f2ad75f38bd.

davepacheco:
Commit 409e18a44093ddf6f7a9b0efa53f2f2ad75f38bd changes `ApiEndpointParameterName` as discussed.  @ahl want to take another look?

ahl:
Looks good!

davepacheco:
The latest changes are just a sync up with master.  As part of doing this, I squashed the commits up to this point, since I had to force-push anyway to sync up.

davepacheco:
The new change should fix #85.

PR #50, opened 2020-08-29
fill in OpenAPI responses description

### LINKED ISSUE #11: openapi: implement response descriptions
OpenAPI gives us several places where we can put relevant information related to a response:

```json
    "/hardware/racks/{rack_id}": {
      "get": {
        "description": "**1** Retrieve information about a rack specified by ID.",
        "operationId": "api_hardware_racks_get_rack",
        "parameters": [
          { "...": "..."}
        ],
        "responses": {
          "200": {
            "description": "**2** Successfully retrieved information about the rack.",
            "content": {
              "application/json": {
                "schema": {
                  "description": "**3** Information that describes a rack",
                  "type": "object",
                  "properties": {
                    "id": {
                      "type": "string",
                      "format": "uuid"
                    }
                  },
                  "required": [
                    "id"
                  ]
                }
              }
            }
          }
        }
      }
    },
```

`**1**` is for the description of the OpenAPI "operation" i.e. the endpoint. Dropshot derives this field from the doc comment on the handler function
`**2**` is for the description of the OpenAPI "response". In this context, we mostly case about successful responses as error codes will be handled more-or-less generically (see #7). Dropshot doesn't have any content in this field today.
`**3**` is for the description of the OpenAPI "schema" i.e. the datatype that the response produces. Dropshot derives this today from the doc comment on the structure type.

Question 1: Do we need all three? 1 seems useful as it describes the purpose of the interface. 2 and 3 seem a bit redundant, but I'd argue that we may want both: the datatype may be used in several places and we may want to customize the response description. For example, `POST`, `GET`, and `PUT` (create, view, update) may return the same structure but we may want the description for 2 or 3 to differ. Alternatively, we may just want 3, leaving 2 implied.

Question 2.1: if we want to get rid of one of 2 or 3, which should we keep? Should the text continue to come from the doc comment on the type?

Question 2.2: if we want to keep all 3, where should the description for 2 come from? One could imagine asking the developer to specify this in the `#[endpoint {...}]` attribute, but this seems like it might be kind of a nuisance. Alternatively, we could use the `HttpTypedResponse` implementations to have an automatically generated, semi-tailored description. For example, `HttpResponseCreated` would say something like "creation successful"; `HttpResponseDeleted`, "deletion successful"; `HttpResponseOkObjectList`, "list of entities".

ahl:
fixes #11 as described in that discussion.

This leaves as a TODO the case of the freeform `Response<Body>` as OpenAPI still, technically, requires a description.

We also may want to consider in the future letting consumers specify a custom description field. After further consideration, perhaps specifying the description for the success case in the `#[endpoint ... ]` macro isn't that bad if developers want to specify it for freeform return types or customize it for typed responses.

### DIFF
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -925,6 +925,7 @@ impl HttpResponse for Response<Body> {
         ApiEndpointResponse {
             schema: None,
             success: None,
+            description: None,
         }
     }
 }
@@ -947,6 +948,7 @@ pub trait HttpTypedResponse:
 {
     type Body: JsonSchema + Serialize;
     const STATUS_CODE: StatusCode;
+    const DESCRIPTION: &'static str;
 
     /**
      * Convenience method to produce a response based on the input
@@ -978,6 +980,7 @@ where
         ApiEndpointResponse {
             schema: Some(ApiSchemaGenerator::Gen(T::Body::json_schema)),
             success: Some(T::STATUS_CODE),
+            description: Some(T::DESCRIPTION.to_string()),
         }
     }
 }
@@ -999,6 +1002,7 @@ impl<T: JsonSchema + Serialize + Send + Sync + 'static> HttpTypedResponse
 {
     type Body = T;
     const STATUS_CODE: StatusCode = StatusCode::CREATED;
+    const DESCRIPTION: &'static str = "successful creation";
 }
 impl<T: JsonSchema + Serialize + Send + Sync + 'static>
     From<HttpResponseCreated<T>> for HttpHandlerResult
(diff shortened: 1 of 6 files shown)


dropshot/src/handler.rs
```diff
@@ -1022,6 +1026,7 @@ impl<T: JsonSchema + Serialize + Send + Sync + 'static> HttpTypedResponse
 {
     type Body = T;
     const STATUS_CODE: StatusCode = StatusCode::ACCEPTED;
+    const DESCRIPTION: &'static str = "successful operation";
```
davepacheco, line 1029: I think this one typically just means "successfully enqueued operation" or the like.

dropshot/src/handler.rs
```diff
@@ -1080,6 +1087,7 @@ pub struct HttpResponseUpdatedNoContent();
 impl HttpTypedResponse for HttpResponseUpdatedNoContent {
     type Body = ();
     const STATUS_CODE: StatusCode = StatusCode::NO_CONTENT;
+    const DESCRIPTION: &'static str = "successful operation";
```
davepacheco, line 1090: "resource updated"?

ahl:
> (edit: sorry, I completely missed that one of the changes _was_ the openapi test)

Yeah, I'm just leaning on that one test to do a lot of this openapi testing, but I think it's got decent coverage.

PR #51, opened 2020-08-31
petstore and OpenAPI pagination descriptions

ahl:
some stuff I bumped into prepping for the ASC talk


dropshot/src/pagination.rs
```diff
@@ -209,6 +212,9 @@ pub struct PaginationParams<ScanParams, PageSelector> {
      * Consumers should use [`dropshot::RequestContext::page_limit()`] to access
      * this value.
      */
+    #[schemars(
```
davepacheco, line 215: Nice -- thanks for fixing this.

PR #61, opened 2020-11-13
upgrade schemars to 0.8

mikeyhew:


davepacheco:
@ahl any objection here?  Here's the change log: https://github.com/GREsau/schemars/blob/master/CHANGELOG.md

There are a bunch of breaking changes here but CI passed on all platforms so I believe that means they don't affect us.

davepacheco:
(thanks @mikeyhew!)

ahl:
No objection! If CI is happy, I'm happy!

davepacheco:
Thanks again.

PR #63, opened 2020-11-26
feat: mark endpoint const as pub

baetheus:
The endpoint macro did not create a pub const and thus could
not be used in modules outside of main. I updated the macro
but feel that the visibility of the entities that it creates
should be pulled from the visibility of the handler function
that it wraps.

I also updated the basic example to show the module namespacing
that can now be achieved.

### DIFF
--- a/dropshot_endpoint/src/lib.rs
+++ b/dropshot_endpoint/src/lib.rs
@@ -100,6 +100,7 @@ fn do_endpoint(
     let name = &ast.sig.ident;
     let name_str = name.to_string();
     let method_ident = format_ident!("{}", method);
+    let visibility = &ast.vis;
 
     let description_text_provided = extract_doc_from_attrs(&ast.attrs);
     let description_text_annotated = format!(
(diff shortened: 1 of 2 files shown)


davepacheco:
Thanks for this.  For what it's worth, in the prototypes we've built, we _do_ put the endpoint handler functions in separate modules, and they're generally _not_ `pub`, but we include a public `register` function that registers them.  This way the specific endpoints don't need to be exposed outside the module.

This is a cool idea though.  @ahl what do you think?  It does seem nice if the visibility were determined by the visibility of the function that the user defined.

CI appears to have failed -- see the "details" link above for more.

Instead of changing basic.rs, could you add a new example for this, like maybe "modules.rs"?  basic.rs is supposed to be pretty minimal.

baetheus:
Sounds good. I'll work on the following when time permits:

* Resolve test failures
* Revert the basic.rs example and implement one that shows public module changes
* Revisit the endpoint macro and tokenize the visibility of the endpoint function and apply it to the macro output.

dropshot_endpoint/src/lib.rs
```diff
@@ -212,7 +212,7 @@ fn do_endpoint(
         // ... a constant of type `#name` whose identifier is also #name
         #[allow(non_upper_case_globals, missing_docs)]
         #description_doc_comment
-        const #name: #name = #name {};
+        pub const #name: #name = #name {};
```
ahl, line 215: How about we look at the function and mirror its visibility? This should be accessible through `ast.vis`

ahl (changes requested):
Thanks for this. As @davepacheco said, a different example would be great.

As noted in the code, matching the visibility specified by the user might be the least surprising approach.

dropshot_endpoint/src/lib.rs
baetheus: Sorry for the delay. I've implemented the mirror in visibility, cleaned up the resulting test errors, and updated the related example.

ahl:
Thanks for making the changes. I think this came out very nicely. Much appreciated.

PR #66, opened 2020-12-22
WIP: update to tokio 0.3

davepacheco:
See #60.

The only breaking build-time change I've found is that `tokio::time::delay_for` has been renamed to `tokio::time::sleep` (with no alias or anything).  At runtime, things still fail like this:

```
$ RUST_BACKTRACE=full cargo test 
    Finished test [unoptimized + debuginfo] target(s) in 0.09s
     Running target/debug/deps/dropshot-a4cd805ac67f9af3

running 55 tests
test from_map::test::test_lone_literal ... ok
test api_description::test::test_empty_struct ... ok
test from_map::test::test_deep ... ok
test from_map::test::test_missing_data1 ... ok
test from_map::test::test_missing_data2 ... ok
test from_map::test::test_types ... ok
test handler::test::test_metadata_flattened ... ok
test api_description::test::test_badpath1 ... ok
test api_description::test::test_badpath3 ... ok
test api_description::test::test_badpath2 ... ok
test handler::test::test_metadata_flattened_enum ... ok
test handler::test::test_metadata_pagination ... ok
test handler::test::test_metadata_simple ... ok
test api_description::test::test_garbage_barge_structure_conversion ... ok
test logging::test::test_config_bad_file_no_level ... ok
test logging::test::test_config_bad_file_no_file ... ok
test logging::test::test_config_bad_terminal_bad_level ... ok
test logging::test::test_config_bad_log_mode ... ok
test logging::test::test_config_bad_terminal_no_level ... ok
test pagination::test::test_results_page ... ok
test logging::test::test_config_bad_file_bad_path_type ... ok
test logging::test::test_config_bad_file_path_exists_fail ... ok
test pagination::test::test_page_token_serialization ... ok
test pagination::test::test_pagparams_parsing ... ok
test router::test::test_embedded_non_variable ... ok
test router::test::test_empty_variable ... ok
test router::test::test_error_cases ... ok
test router::test::test_iter ... ok
test router::test::test_iter2 ... ok
test router::test::test_iter_null ... ok
test logging::test::test_config_stderr_terminal ... ok
test router::test::test_router_basic ... ok
test logging::test::test_config_file ... ok
test router::test::test_duplicate_route1 ... ok
test router::test::test_duplicate_route2 ... ok
test router::test::test_variable_name_bad_start ... ok
test router::test::test_duplicate_varname ... ok
test router::test::test_variables_basic ... ok
test router::test::test_inconsistent_varname ... ok
test router::test::test_variables_multi ... ok
[...]

PR #72, opened 2020-12-30
Upgrade to tokio 1.0 and hyper 0.14

### LINKED ISSUE #60: Upgrade to tokio 0.3
Tokio 0.3 came out in mid-october (https://tokio.rs/blog/2020-10-tokio-0-3), and they're now on 0.3.3, so it seems like it should be stable enough to upgrade to now. I haven't done an upgrade from 0.2  - 0.3, but just speculating, it looks like the main thing to do is use the new feature flag names, and call the `.compat()` method from `tokio-compat-02` to wrap futures from crates that are still on 0.2.

nshalman:
There are some minor code changes needed to complete the upgrade to tokio 1.0 and hyper 0.14 which I've made here.
Additionally hyper upgrades to bytes 1 and that change seems to be needed here as well.

I fell down this rabbit whole as I was planning to start writing my own service based on dropshot, but haven't written it yet so I don't have any downstream codebases to test against. I got no test failures on my machine, but some were skipped so I figured I'd see what your CI does with this.

Let me know how to rewrite the summary, if you want a squashed branch, etc. or, since I've checked the "Allow edits by maintainers" box, feel free to do with this PR as you wish! 

Fixes #60 
Duplicate of #66

### DIFF
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/oxidecomputer/dropshot/"
 [dependencies]
 async-trait = "0.1.24"
 base64 = "0.12.3"
-bytes = "0.5.4"
+bytes = "1"
 futures = "0.3.1"
 hostname = "0.3.0"
 http = "0.2.0"
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -14,7 +14,6 @@ bytes = "0.5.4"
 futures = "0.3.1"
 hostname = "0.3.0"
 http = "0.2.0"
-hyper = "0.13.0"
 indexmap = "1.0.0"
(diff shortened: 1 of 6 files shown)


ahl (approved):
This looks good to me. @davepacheco can you merge if you're happy?

davepacheco:
Thanks for doing this!  I went through the changelogs for Bytes, Tokio, and [Hyper](https://github.com/hyperium/hyper/blob/master/CHANGELOG.md).  It looks basically fine.  I'm not a fan of [logging when we fail to set socket options](https://github.com/hyperium/hyper/issues/2359) and I'm not sure if "hyper no longer emits log records automatically" affects us.  Regardless, unless we're going to switch to something else, moving forward here is the right change.  Thanks again!

Issue #68, opened 2020-12-22
improve ergonomics of OpenAPI definition generation


jclulow:
The current [print_openapi()](https://docs.rs/dropshot/0.3.0/dropshot/struct.ApiDescription.html#method.print_openapi) is a little unwieldy; it contains a non-optional argument after a sea of optionals, and even if you intend to provide all of the optionals it's a lot of strings that you have to put in the right order.

A simple call, with no optionals, looks like this today:

```rust
api.print_openapi(
    &mut f,
    &"Keeper API".to_string(),
    None,
    None,
    None,
    None,
    None,
    None,
    None,
    &"1.0".to_string(),
)?;
```

I suspect it'd be better as a struct, like:

```rust
api.print_openapi(
    &mut f,
    OpenApiDetails {
        name: "Keeper API".to_string(),
        version: "1.0".to_string(),
        ..Default::default()
    },
)?;
```

It'd be easier and clearer to then incrementally add additional fields; e.g.,

```rust
api.print_openapi(
    &mut f,
    OpenApiDetails {
        name: "Keeper API".to_string(),
        version: "1.0".to_string(),
        license_name: Some("CDDL").to_string(),
        ..Default::default()
    },
)?;
```

ahl:
Agreed that this is unwieldy, but I don't like the idea of picking defaults for fields that truly require the caller to specify something e.g. the title or version which are mandated by the spec. I'm inclined to do something like this:

```rust
api.print_openapi(
    &mut f,
    OpenApiDetails::new(title, version)
        .license_name("CDDL")
);
```

... but I'm worried that's not very rusty and would welcome suggestions for how it might be improved.

jclulow:
That sounds great!  The point about defaults for values that need something specified is a good one, and the Default pattern doesn't have a good answer there.

I believe what you're describing is indeed quite idiomatic, and is generally caller the [builder pattern](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html).  See also `std::process::Command` or `std::fs::OpenOptions`.

jclulow:
Actually even better might be `api.printer(name, version).licence(...).description(...).write(stream)?`.

ahl:
Yeah, I'm a pretty big fan of the builder pattern in Java, I just haven't seen employed much in Rust.

Agreed that what I had proposed wouldn't actually work, and that last version looks good.

jclulow:
I'll take a swing at an implementation?

[...]

PR #75, opened 2021-01-01
improve ergonomics of OpenAPI definition generation (#68)

jclulow:
**NOTE:** This is a preliminary review.  If the shape and premise are acceptable I will go and write documentation comments and tests.

I left the existing `print_openapi()` routine as-is as part of this change, though if we don't actually care about it I am happy to remove it.

The new API works, in practice, like this:

```rust
    api.register(enrol).unwrap();
    api.register(report_start).unwrap();
    api.register(report_output).unwrap();
    api.register(report_finish).unwrap();

    if let Some(s) = p.opt_str("S") {
        let mut f = std::fs::OpenOptions::new()
            .create_new(true)
            .write(true)
            .open(&s)?;
        api.openapi("Keeper API", "1.0")
            .description("report execution of cron jobs through a \
                mechanism other than mail")
            .contact_name("Joshua M. Clulow")
            .contact_url("https://github.com/jclulow/keeper")
            .write(&mut f)?;
        return Ok(());
    }
```

The `license` property in the definition document requires a `name`, so I have tried to arrange the methods on the builder so that one can either set the `name` _and_ the `url`, or just the `name`, but not accidentally only set the `url`.  I have also added a `.json()` routine to the builder that will give the caller a `serde_json::Value` back that they might store and provide as the result of an API endpoint like you might do with a WSDL file etc.


dropshot/src/api_description.rs
```diff
@@ -252,11 +252,22 @@ impl ApiDescription {
...
+    {
+        OpenApiConfig::new(self, title.as_ref(), version.as_ref())
+    }
+
     /**
      * Emit the OpenAPI Spec document describing this API in its JSON form.
      */
-    // TODO: There's a bunch of error handling we need here such as checking
-    // for duplicate parameter names.
     pub fn print_openapi(
```
ahl, line 271: Should we mark this deprecated?

jclulow: Yes, I think so!  I have added a note to the documentation and attached the deprecated attribute.

jclulow:
Alright, I reckon this is probably about ready.  I've added some documentation (and confirmed that it looks OK in the browser) and some very basic tests (which appear to pass!).

davepacheco:
LGTM.  @ahl are you good with the latest version?

dropshot/src/api_description.rs
```diff
@@ -252,11 +252,30 @@ impl ApiDescription {
...
+        S1: AsRef<str>,
+        S2: AsRef<str>,
+    {
+        OpenApiDefinition::new(self, title.as_ref(), version.as_ref())
+    }
+
     /**
      * Emit the OpenAPI Spec document describing this API in its JSON form.
+     *
+     * This routine is deprecated in favour of the new openapi() builder
```
ahl, line 275: I love your colourful spelling!

ahl (approved):
ship it!

PR #78, opened 2021-01-19
update docs to explain intended pattern for common code (#58)

### LINKED ISSUE #58: update docs to explain intended pattern for common code
Good Day,

I'm currently learning dropshot,  and am curious on how to configure a handler to run on all requests.

Similar to http://restify.com/docs/server-api/#use  (https://github.com/joyent/sdc-cloudapi/blob/master/lib/app.js#L426) 

Or perhaps there is an alternative pattern I should use while using DropShot?

Thanks,
Bruce

davepacheco:
Fixes #58.


README.adoc
```diff
@@ -107,3 +107,13 @@ and then uses it as though it had just been created), and `"fail"` (which causes
...
... these handlers is to share code between handlers, what if we rely instead on existing mechanisms -- i.e., function calls.  The big risk is that it's easy for someone to accidentally forget some important function call, like the one that authentices or authorizes a user.  We haven't gotten far enough in a complex implementation to need this yet, but the plan is to create a pattern of utility functions that return typed values.  For example, where in Node.js you might add an early authentication handler that fills in `request.auth`, with Dropshot you'd have an authentication function that _returns_ an `AuthzContext` struct.  Then anything that needs authentication consumes the `AuthzContext` as a function argument.  As an author of a handler, you know if you've got an `AuthzContext` available and, if not, how to get one (call the utility function).  This composes, too: you can have an authorization function that returns an `AuthnContext`, and the utility function that returns one can consume the `AuthzContext`.  Then anything that requires authorization can consume just the `AuthnContext`, and you know it's been authenticated and authorized (possibly with details in that structure).
```
ahl, line 117:
```suggestion
With Dropshot, we wanted to try something different: if the primary purpose of these handlers is to share code between handlers, what if we rely instead on existing mechanisms -- i.e., function calls.  The big risk is that it's easy for someone to accidentally forget some important function call, like the one that authenticates or authorizes a user.  We haven't gotten far enough in a complex implementation to need this yet, but the plan is to create a pattern of utility functions that return typed values.  For example, where in Node.js you might add an early authentication handler that fills in `request.auth`, with Dropshot you'd have an authentication function that _returns_ an `AuthzContext` struct.  Then anything that needs authentication consumes the `AuthzContext` as a function argument.  As an author of a handler, you know if you've got an `AuthzContext` available and, if not, how to get one (call the utility function).  This composes, too: you can have an authorization function that returns an `AuthnContext`, and the utility function that returns one can consume the `AuthzContext`.  Then anything that requires authorization can consume just the `AuthnContext`, and you know it's been authenticated and authorized (possibly with details in that structure).
```

README.adoc
```diff
@@ -107,3 +107,13 @@ and then uses it as though it had just been created), and `"fail"` (which causes
...
+
+It's early, and we may find we need richer facilities in the framework.  But we're hopeful this approach will make it faster and smoother to iterate on complex API servers.
```
ahl, line 119:
```suggestion
It's early, and we may find we need richer facilities in the framework.  But we're hopeful this approach will make it faster and smoother to iterate on complex API servers. We welcome feedback from folks using or interested in using Dropshot.
```

ahl (approved):
typo (I think?) and a very mild suggestion that you should feel free to reject

PR #81, opened 2021-01-28
Restructure HttpServer's API

smklein:
TL;DR: Split HttpServer in twain to make the state-machine logic a little better.

Old API: 
❌ [wait_for_shutdown](https://docs.rs/dropshot/0.3.0/dropshot/struct.HttpServer.html#method.wait_for_shutdown)
 required an external "join_handle" - it never actually acted on the `self` object of the server, and there are no guarantees the provided `JoinHandle` matches the one from `run`.
❌ [run](https://docs.rs/dropshot/0.3.0/dropshot/struct.HttpServer.html#method.run) can be invoked multiple times, which would cause a panic on the second invocation.

New API:
✔️ Invoking `start` consumes `self`, and may only be invoked once, creating a new `HttpServer` object which represents the running server.
✔️ The running HttpServer may only be terminated once by calling `close` (taking no arguments), which signals the server to stop immediately, and waits for it to exit.
✔️ The HttpServer implements `Future`, and may be `await`-ed to query the completion of the server.


dropshot/src/server.rs
```diff
@@ -173,6 +151,62 @@ impl HttpServer {
...
+
+    pub fn app_private(&self) -> Arc<dyn Any + Send + Sync + 'static> {
+        Arc::clone(&self.app_state.private)
+    }
+
+    /**
+     * Signals the currently running server to terminate (if it hasn't
+     * stopped already) and waits for it to exit.
+     */
+    pub async fn terminate(self) -> Result<(), String> {
```
davepacheco, line 182: Summarizing for myself here: before this change, there were two interfaces that were part of shutdown:

- close(), which initiates shutdown but does not wait for it
- wait_for_shutdown(), which waits for the shutdown to finish (but doesn't initiate it)

It took me a while reviewing this to remember the reason for decoupling these: there are consumers where it makes sense to `wait_for_shutdown()` without calling `close()` -- namely, a standard server program, including all the examples in this repo.  There are also potentially use cases where `close()` might be invoked a different context than the one that created and waited on the server.  I'm not sure if this second case worked before, and it's not important to support that now if we don't use it already.

In the new code, we have `terminate()` and `with_graceful_shutdown()`.  `terminate()` initiates shutdown and waits for it, which is fine.  But for the standard server program case that I mentioned above, we still need a way to wait-until-shutdown without actually starting a shutdown.  I gather that's why `with_graceful_shutdown()` exists, but this interface feels awkward to me.  If I remember correctly, @cbiffle and I both found `with_graceful_shutdown()` in hyper to be confusing, both in its name and behavior.  I was trying to create a more intuitive interface in Dropshot.  One thing that's awkward here: if you're looking at this API, it's not obvious why a caller wouldn't just wait for the future themselves and then call terminate() -- this function doesn't seem to add anything that the caller couldn't do themselves.

If we end up keeping `terminate()`, I'd like the name to be a bit more explicit that it waits for shutdown to complete.  What about `terminate_and_wait`?  (Or `close_and_wait`.  Why the change from `close` to `terminate`?)

dropshot/tests/test_pagination.rs
```diff
@@ -227,7 +227,7 @@ async fn test_paginate_errors() {
     struct ErrorTestCase {
         path: String,
         message: &'static str,
-    };
+    }
```
davepacheco, line 230: I'm curious what's the story behind this?  I would have expected rustfmt to always put the semicolon here or always leave it off, but presumably we were style-clean both before and after this change.

dropshot/examples/basic.rs
```diff
@@ -58,15 +58,15 @@ async fn main() -> Result<(), String> {
...
+    let server = HttpServer::new(&config_dropshot, api, api_context, &log)
         .map_err(|error| format!("failed to create server: {}", error))?;
     let server_task = server.run();
 
     /*
      * Wait for the server to stop.  Note that there's not any code to shut down
      * this server, so we should never get past this point.
      */
-    server.wait_for_shutdown(server_task).await
+    server_task.with_graceful_shutdown(std::future::pending()).await
```
davepacheco, line 69: This feels pretty awkward to me, especially being the "hello world" golden path.  I suggested elsewhere just dropping the `with_graceful_shutdown()` function, in which case we could presumably just remove these calls.  It might be a problem if there's nothing here because the program might exit.  (I haven't confirmed this.)  I took a look at [hyper's "hello world" guide](https://hyper.rs/guides/server/hello-world/) and it looks like they give you a way to just wait for the server to exit without having to provide a never-resolving future argument.  That seems reasonably clean.  (In hyper, the server itself impls Future, which seems too overloaded to me, but we could just have a method that does this.)

This affects the other examples too.

More on this below.

dropshot/examples/basic.rs
```diff
@@ -58,15 +58,15 @@ async fn main() -> Result<(), String> {
     /*
      * Set up the server.
      */
-    let mut server = HttpServer::new(&config_dropshot, api, api_context, &log)
+    let server = HttpServer::new(&config_dropshot, api, api_context, &log)
         .map_err(|error| format!("failed to create server: {}", error))?;
     let server_task = server.run();
```
davepacheco, line 63: Particularly in the examples, but ideally for all the in-repo callers of `run()`, can we rename this variable?  I find it pretty confusing for it to be called `server_task` now.  In fact, combining a few of the suggestions here, maybe this whole statement can become:

```
let server = HttpServerStarter::new(&config_dropshot, api, api_context, &log)
    .map_err(|error| format!("failed to create server: {}", error))?
    .run()
```

dropshot/src/server.rs
```diff
@@ -60,62 +63,47 @@ pub struct ServerConfig {
...
-     * a Future that resolves when the server is finished?
+    /**
+     * Begins execution of the underlying Http server.
      */
-    pub fn run(&mut self) -> tokio::task::JoinHandle<Result<(), hyper::Error>> {
-        let future =
-            self.server_future.take().expect("cannot run() more than once");
-        tokio::spawn(async { future.await })
-    }
+    pub fn run(self) -> RunningHttpServer {
```
davepacheco, line 82: Per the TODO that was here, and especially if we call this thing `HttpServerStarter`, I think this function is more accurately called `start()` now.

davepacheco (commented):
Thanks for taking this on!  I've got some thoughts below on the new shutdown interfaces.

Two other things to do once we're settled on the interface changes:

- This will obviously be a breaking change.  Dropshot doesn't have much of a release process yet, but we do have several consumers now, so I want to start doing one.  Could you draft a few sentences explaining to a consumer how to update their program for the new API?  You can just leave that here in the PR and I'll add that to the changelog when I create one.
- Although this is a separate repo and we do have other consumers now, the expectation is that Dropshot is still pretty tightly coupled to oxide-api-prototype.  Could you try updating oxide-api-prototype for these changes?  That will be good validation that the new interface still works where we need it.

Thanks again!

dropshot/tests/test_pagination.rs
smklein: My bad; I'll put this back! It's an artifact from my defaulting to nightly, where it'll become a warning.

dropshot/src/server.rs
smklein: > Summarizing for myself here: before this change, there were two interfaces that were part of shutdown:
> 
> * close(), which initiates shutdown but does not wait for it
> * wait_for_shutdown(), which waits for the shutdown to finish (but doesn't initiate it)

That's accurate, this is the old version:

```rust
pub fn close(self) { ... }

pub async fn wait_for_shutdown<'_>(
    &'_ mut self,
    join_handle: JoinHandle<Result<(), Error>>
) -> Result<(), String> { ... }
```

My problem with the old implementation was the really loose coupling of the `join_handle` with the actual server.

`wait_for_shutdown` never acts on `self`, so it can be invoked any number of times with totally unrelated handles.

> It took me a while reviewing this to remember the reason for decoupling these: there are consumers where it makes sense to `wait_for_shutdown()` without calling `close()` -- namely, a standard server program, including all the examples in this repo. There are also potentially use cases where `close()` might be invoked a different context than the one that created and waited on the server. I'm not sure if this second case worked before, and it's not important to support that now if we don't use it already.
> 
> In the new code, we have `terminate()` and `with_graceful_shutdown()`. `terminate()` initiates shutdown and waits for it, which is fine. But for the standard server program case that I mentioned above, we still need a way to wait-until-shutdown without actually starting a shutdown. I gather that's why `with_graceful_shutdown()` exists, but this interface feels awkward to me. If I remember correctly, @cbiffle and I both found `with_graceful_shutdown()` in hyper to be confusing, both in its name and behavior. I was trying to create a more intuitive interface in Dropshot. 
> One thing that's awkward here: if you're looking at this API, it's not obvious why a caller wouldn't just wait for the future themselves and then call terminate() -- this function doesn't seem to add anything that the caller couldn't do themselves.

I think this is a good point, and I agree with you - I've updated the API to the following:

- HttpServer implements `Future`, which is the primary mechanism to await the completion of the server. This basically punts as much control of the server lifetime as possible to the client.
- I've kept the async, self-consuming `close` function (sticking with that name instead of `terminate` to minimize changes) for the same hard-stop behavior.


> If we end up keeping `terminate()`, I'd like the name to be a bit more explicit that it waits for shutdown to complete. What about `terminate_and_wait`? (Or `close_and_wait`. Why the change from `close` to `terminate`?)

I have no strong attachment to the name `terminate` - happy to update to `close`.

It does seem quirky to have an async method named "XXX_and_wait" though - this seems like it should be implicit within the future.

I read:

`do_foo_and_wait().await;`

and

`do_foo().await;`

as doing the same thing.

With the updated version, do you still feel strongly about this?

dropshot/examples/basic.rs
smklein: Ack, updated invocation is now:

```rust
server.await
```

(see below for details)

smklein:
> Thanks for taking this on! I've got some thoughts below on the new shutdown interfaces.
> 
> Two other things to do once we're settled on the interface changes:
> 
> * This will obviously be a breaking change.  Dropshot doesn't have much of a release process yet, but we do have several consumers now, so I want to start doing one.  Could you draft a few sentences explaining to a consumer how to update their program for the new API?  You can just leave that here in the PR and I'll add that to the changelog when I create one.

Here's my attempt:

`HttpServer` has been split into two objects, `HttpServerStarter`, and `HttpServer`.
- In the old implementation, `HttpServer` represented both a pending and running server. Callers were expected to invoke `run()` to begin execution of the old server.
- In the new implementation, `HttpServerStarter` may be used to construct a server, and `HttpServer` represents the running server. Invoking `HttpServerStarter::start()` creates and `HttpServer` object, which represents the new server.

`HttpServer` now implements futures, and does not directly expose a `tokio::JoinHandle`.
- In the old implementation, `HttpServer` returned a `JoinHandle`, and callers were expected to invoked `wait_for_shutdown` to await the completion of a server.
- In the new implementation, `HttpServer` implements future, and may be `await`-ed directly.

As an example:

```rust
// Old Version:
let mut server = HttpServer::new( /* Arguments are the same between versions */ )
  .map_err(|error| format!("failed to start server: {}", error))?;

let server_task = server.run();
server.wait_for_shutdown(server_task).await;

// New Version
let server = HttpServerStarter::new( /* Arguments are the same between versions */ )
  .map_err(|error| format!("failed to start server: {}", error))?
  .start();

server.await;
```


> * Although this is a separate repo and we do have other consumers now, the expectation is that Dropshot is still pretty tightly coupled to oxide-api-prototype.  Could you try updating oxide-api-prototype for these changes?  That will be good validation that the new interface still works where we need it.
> 
> Thanks again!

Sure, I have a branch here: https://github.com/oxidecomputer/oxide-api-prototype/pull/34

Won't work on GitHub until this is merged, but locally, this passes all tests and simplifies the code slightly.

dropshot/src/server.rs
```diff
@@ -173,6 +152,85 @@ impl HttpServer {
...
+     * Signals the currently running server to stop (if it hasn't
+     * stopped already) and waits for it to exit.
+     */
+    pub async fn close(mut self) -> Result<(), String> {
+        self.close_channel
+            .take()
+            .expect("cannot close twice")
+            .send(())
+            .expect("failed to send close signal");
+        if let Some(handle) = self.join_handle.take() {
```
davepacheco, line 184: When exactly would we get here and find that this was not `Some`?

dropshot/src/server.rs
```diff
@@ -173,6 +152,85 @@ impl HttpServer {
...
+    }
+}
+
+/**
+ * Dropping the HttpServer will terminate the server, without actually waiting
+ * for it to stop.
+ *
+ * For graceful termination, use the `close` function instead.
+ */
+impl Drop for HttpServer {
```
davepacheco, line 201: What would happen if we didn't impl `Drop` here?

dropshot/src/server.rs
davepacheco: I don't feel strongly about it.

davepacheco (commented):
Thanks for doing this.  This is much cleaner than what was in master before.  This was probably some of the earliest code in Dropshot and some of my own earliest Rust code!  I had a couple of small questions but they're not blockers.

Thanks also for updating oxide-api-prototype.  That code look cleaner too.

Lastly, could you update CHANGELOG.adoc for this so that other people using Dropshot know how to update their code?

dropshot/src/server.rs
smklein: If the implementation of `Future` returns `Ready`.

Basically, in the implementation of `Future`, we poll the handle. If it completes, we drop it, to avoid re-polling a completed future (which is part of the Future API requirements: "[Once a future has finished, clients should not poll it again.](https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll)").

In other word: `self.join_handle` should be `Some` if we haven't seen the future complete, and should be `None` if it *has* completed.

dropshot/src/server.rs
smklein: If a client let an `HttpServer` object go out of scope without calling `await` *and* without calling `close`, it would be impossible to close the server. It would be running forever, and there would be no way to tell it to shut down.

I'm open to discuss this - personally, I felt "advise the server to terminate, even though we can't await when it does shut down" is better than "leak it forever", but I suppose we could delete this `Drop` impl and leak-by-default. That being said, I acknowledge this "leak when you let things go out of scope" behavior is seen elsewhere, such as with Tokio's JoinHandles.

dropshot/src/server.rs
davepacheco: If we drop it in the Future, how can we get to this point in `close` and find it `None`?

dropshot/src/server.rs
davepacheco: That's interesting.  I believe that before this change, if the server got dropped, it was shutdown.  (I ran into a bug with this a few months ago.)

I don't have a strong feeling about desired behavior here, but I also don't feel I fully grok the implications of _not_ doing this (e.g., for things like cancellation or panic unwinding).

Is it possible to log about this case?  If it's a reasonable thing to do, maybe just at "info".  If not, maybe "warn" -- or maybe even panic?

CHANGELOG.adoc
```diff
@@ -15,6 +15,42 @@ https://github.com/oxidecomputer/dropshot/compare/v0.4.0\...HEAD[Full list of co
 // configured in release.toml.  DO NOT change the format of the headers or the
 // list of raw commits.
 
+`HttpServer` has been split into two objects, `HttpServerStarter`, and
```
davepacheco, line 18: I think there are some asciidoc formatting nits here.  (The list doesn't render as a list, at least on GitHub.)

dropshot/src/server.rs
smklein: I'm not sure I see how the server would be shutdown before this change - the only connection to the server is via the tokio::task joinhandle, which explicitly does *not* stop the background task when dropped.

Regardless, added a "warn" statement to the drop implementation.

dropshot/src/server.rs
smklein: > If we drop it in the Future

Yeah, so "drop it in the Future" means (referring to my implementation of `HttpServer::poll`):

- We called `server.join_handle.take()` to access the handle,
- We polled the handle and it was not pending
- We `return poll` without "putting the handle back in `self.join_handle`", which we would have done if `poll.is_pending()` was true.

> how can we get to this point in `close` and find it `None`?

`self.join_handle` is an `Option`. In the implementation of `Future`, we **always** call: `server.join_handle.take()`, and we **only** put the handle back if `poll.is_pending()`.

So *if* we dropped it in the future (by completing the join_handle), close will *always* find it `None`, because the implementation of `poll` called `take()` without `replace()`.

Does that make sense?

CHANGELOG.adoc
smklein: Thanks for the heads up, looks like this is fixed now.

dropshot/src/server.rs
davepacheco: > I'm not sure I see how the server would be shutdown before this change - the only connection to the server is via the tokio::task joinhandle, which explicitly does not stop the background task when dropped.

That's true, but the guts of the Dropshot server is really in the Hyper server that it owns, and that may shut itself down on `drop`.

To give a little more detail about the problem I saw before: it related to this kind of test in oxide-api-prototype:
https://github.com/oxidecomputer/oxide-api-prototype/blob/d254e720b20cae2c8076ac196023a51af0f1fb6c/tests/test_instances.rs#L39

That `test_setup` function is this one:
https://github.com/oxidecomputer/oxide-api-prototype/blob/d254e720b20cae2c8076ac196023a51af0f1fb6c/tests/common/mod.rs#L46-L95

It returns a struct that owns the server and (prior to this change) the join handle.

At one point, the line I linked above (test_instances.rs L39) was instead something like `let client = test_setup("test_instances").await.external_client` (i.e., combining lines 39 and 40).  I found that when I did that, tests failed because the server was shutdown immediately, right after that call.  @cbiffle and I debugged this to see that `drop` was being invoked on the server because we weren't hanging onto the `ControlPlaneTestContext` struct or referencing the server later.  It's possible something else changed in the meantime, as I can't reproduce this by making that change to oxide-api-prototype#master.

Do you think it's worth testing the behavior now without this impl to see if we can avoid it?

davepacheco (approved):

dropshot/src/server.rs
davepacheco: Got it -- we drop the handle, not the server.  So the sequence of events that gets us here is just that someone has waited for the `Future` and then called `close()`.  Thanks for clarifying.

smklein: Exactly!

dropshot/src/server.rs
smklein: Sure, will test and report back before merging

smklein: So I tried making a couple tests:

Baseline, to make sure "the normal case works":
1. Create an `ApiDescription` with a simple callback.
2. Invoke `HttpServerStarter::new(...).unwrap().start()`.
3. Keep the server object alive
4. Make a client request --> Observe the result is OK

The drop-server case:
1. Create an `ApiDescription` with a simple callback.
2. Invoke `HttpServerStarter::new(...).unwrap().start()`.
3. Drop the server
4. Make a client request --> Expect it to fail  (the server should be terminated).

*without* the `impl Drop`, the `tokio::spawn`-ed task panics, as it reports "dropshot server shutting down without invoking close()". The client request then fails, as the server thread has panicked.

*with* the `impl Drop`, the `tokio::spawn`-ed task no longer panics - instead, it emits the warning I added ("dropped HttpServer without calling close or await"), and the client request fails.

TL;DR: I *think* the trigger for the server exiting on termination is actually the `tokio::sync::oneshot::Sender` object going out-of-scope, which causes the "graceful shutdown" to trigger with an error.

Regardless, it seems this behavior is covered, and if we consider it an error, I'll happily remove the `impl Drop` and add documentation to the `HttpServer` struct instead, clarifying that it must be `close`-ed or `await`-ed to completion (or else it'll panic).

smklein: One extra layer of complexity: if the `tokio::spawn`-ed task panics as a result of this "server-out-of-scope" issue, it actually [won't stop execution](https://github.com/tokio-rs/tokio/issues/700) if the default tokio runtime is used. I noticed this "panic in the spawned task" behavior actually wouldn't cause execution to stop / tests to fail.

To me, this is a *big* red flag - if a task can panic, but execution remains uninterrupted, it would be easy to misuse the API without receiving any meaningful notification. I've updated the implementation of drop to flag this case more explicitly to the task holding the server, and added some test cases to validate this behavior.

smklein:
Before merging, I think it might be worthwhile having a discussion about the operation of the server future.
*Currently*, the server is implemented with an internal `tokio::task` that executes *regardless* of whether or not the server is awaited. From the [documentation about futures](https://doc.rust-lang.org/std/future/trait.Future.html#runtime-characteristics), this is potentially a violation of expectations:

> Futures alone are inert; they must be actively polled to make progress, meaning that each time the current task is woken up, it should actively re-poll pending futures that it still has an interest in.

I think it might make more sense to:
- Avoid calling `tokio::spawn`, and instead just use an (internal) Boxed future to represent the server.
- Only process requests for the server when `server.await` is invoked.

Although this wouldn't explicitly change the API, it would change expectations, in that it would require callers to explicitly await the server.
Currently, callers use a pattern like:

```rust
  let server = HttpServerStarter(...).start();

  // Users expect the server to be running here, without additional prompting.
  client.make_request().await;

  server.close().await;
```

However, removing the "spawned tokio task" from the internal representation would require users to do the following:

```rust
  let server = HttpServerStarter(...).start();

  // Users should not expect the server to respond to requests unless
  // it is await-ed (as it is here).
  select! {
    _ = client.make_request().fuse() => ...,
    _ = server => ...,
  }
  server.close().await;
```

Do you have an opinion here?

dropshot/src/server.rs
davepacheco: > TL;DR: I think the trigger for the server exiting on termination is actually the tokio::sync::oneshot::Sender object going out-of-scope, which causes the "graceful shutdown" to trigger with an error.

That rings a bell -- yes, I think this was how things shut down in the test case I was describing earlier.

> Regardless, it seems this behavior is covered, and if we consider it an error, I'll happily remove the impl Drop and add documentation to the HttpServer struct instead, clarifying that it must be close-ed or await-ed to completion (or else it'll panic).

That makes the most sense to me (and it's why I added that assertion in the first place).  I also have the feeling that folks coming from C++ and maybe mainstream Rust will find it surprising.

> One extra layer of complexity: if the tokio::spawn-ed task panics as a result of this "server-out-of-scope" issue, it actually won't stop execution if the default tokio runtime is used. I noticed this "panic in the spawned task" behavior actually wouldn't cause execution to stop / tests to fail...To me, this is a big red flag - if a task can panic, but execution remains uninterrupted, it would be easy to misuse the API without receiving any meaningful notification.

I think this is an instance of a more general problem.  My strong preference would be that we treat all Rust panics as programmer errors and build all of our own software with panic=abort so that they abort(3c) and dump core on panic.

davepacheco:
> Before merging, I think it might be worthwhile having a discussion about the operation of the server future.
> _Currently_, the server is implemented with an internal `tokio::task` that executes _regardless_ of whether or not the server is awaited. From the [documentation about futures](https://doc.rust-lang.org/std/future/trait.Future.html#runtime-characteristics), this is potentially a violation of expectations:
> 
> > Futures alone are inert; they must be actively polled to make progress, meaning that each time the current task is woken up, it should actively re-poll pending futures that it still has an interest in.
> 
> I think it might make more sense to:
> 
>     * Avoid calling `tokio::spawn`, and instead just use an (internal) Boxed future to represent the server.
> 
>     * Only process requests for the server when `server.await` is invoked.
> 
> 
> Although this wouldn't explicitly change the API, it would change expectations, in that it would require callers to explicitly await the server.
> Currently, callers use a pattern like:
> 
> ```rust
>   let server = HttpServerStarter(...).start();
> 
>   // Users expect the server to be running here, without additional prompting.
>   client.make_request().await;
> 
>   server.close().await;
> ```
> 
> However, removing the "spawned tokio task" from the internal representation would require users to do the following:
> 
> ```rust
>   let server = HttpServerStarter(...).start();
> 
>   // Users should not expect the server to respond to requests unless
>   // it is await-ed (as it is here).
>   select! {
>     _ = client.make_request().fuse() => ...,
>     _ = server => ...,
>   }
>   server.close().await;
> ```
> 
> Do you have an opinion here?

Good catch.  This change feels like it would make the API unnecessarily harder to use.  In practice I think most consumers will want this to be in a separate task.  It seems annoying to make them all do that individually.

I hear what you're saying about this violating the principle of least surprise about futures, though.  Maybe this is an argument for decoupling the server from the wait-for-shutdown future.  `start()` could return the running server, but you'd have a separate function to wait for shutdown.  I'm not sure if one would consider that particular future to be inert or not, but certainly it doesn't go off and do something without you `await`ing on it.

As I understand it, this isn't so much a problem in master today because `run()` doesn't return a Future.  It just starts the server and gives you a handle that you can use for shutting it down.  (Of course, this interface still has all the problems you're solving in this PR.  I think we're close here to the best of both worlds.)

dropshot/src/server.rs
smklein: I agree with you on the goal to "treat all Rust panics as programmer errors". An `impl Drop` still exists, but only to force the main task to panic if close has not been invoked (it does not attempt to trigger a close if it hasn't happened already). I consider this acceptable, because the alternative is a panic in the background task, which may or may not be visible depending on panic=abort behavior.

Marking resolved, happy to re-open if that seems controversial.

davepacheco: I'm having trouble finding documentation about tokio's behavior here, but doesn't this behavior rely on the caller running the server itself in the "main" task?

smklein:
> > Before merging, I think it might be worthwhile having a discussion about the operation of the server future.
> > _Currently_, the server is implemented with an internal `tokio::task` that executes _regardless_ of whether or not the server is awaited. From the [documentation about futures](https://doc.rust-lang.org/std/future/trait.Future.html#runtime-characteristics), this is potentially a violation of expectations:
> > > Futures alone are inert; they must be actively polled to make progress, meaning that each time the current task is woken up, it should actively re-poll pending futures that it still has an interest in.
> > 
> > 
> > I think it might make more sense to:
> > ```
> > * Avoid calling `tokio::spawn`, and instead just use an (internal) Boxed future to represent the server.
> > 
> > * Only process requests for the server when `server.await` is invoked.
> > ```
> > 
> > 
> > Although this wouldn't explicitly change the API, it would change expectations, in that it would require callers to explicitly await the server.
> > Currently, callers use a pattern like:
> > ```rust
> >   let server = HttpServerStarter(...).start();
> > 
> >   // Users expect the server to be running here, without additional prompting.
> >   client.make_request().await;
> > 
> >   server.close().await;
> > ```
> > 
> > 
> > However, removing the "spawned tokio task" from the internal representation would require users to do the following:
> > ```rust
> >   let server = HttpServerStarter(...).start();
> > 
> >   // Users should not expect the server to respond to requests unless
> >   // it is await-ed (as it is here).
> >   select! {
> >     _ = client.make_request().fuse() => ...,
> >     _ = server => ...,
> >   }
> >   server.close().await;
> > ```
> > 
> > 
> > Do you have an opinion here?
> 
> Good catch. This change feels like it would make the API unnecessarily harder to use. In practice I think most consumers will want this to be in a separate task. It seems annoying to make them all do that individually.

 ¯\\_(ツ)_/¯

I am a little split - I think that a "needs-to-be-polled-to-do-anything" future would be more verbose, but also more explicit.

> I hear what you're saying about this violating the principle of least surprise about futures, though. Maybe this is an argument for decoupling the server from the wait-for-shutdown future. `start()` could return the running server, but you'd have a separate function to wait for shutdown. I'm not sure if one would consider that particular future to be inert or not, but certainly it doesn't go off and do something without you `await`ing on it.

I'm not sure I understand this - I was thinking that we'd have `start()` return a running server, which would *only* complete once the server shuts down.

If we split the future in two, when would the "running server" future complete? Why would that be different from the "wait for future" case?

> As I understand it, this isn't so much a problem in master today because `run()` doesn't return a Future. It just starts the server and gives you a handle that you can use for shutting it down. (Of course, this interface still has all the problems you're solving in this PR. I think we're close here to the best of both worlds.)

I think my takeaway here is going to be:
- Leave the interface as it currently exists in this PR
- If the implicitly running servers ends up being a problem later, we can deal with it in a follow-up PR. If not, then it's less verbose, so no problem.

PR #86, opened 2021-02-10
Pass caller-provided server context via Generics

smklein:
Avoid using the dynamic `Any` objects within dropshot to access state, instead relying on static types.

- This change removes the need for dropshot users to manually downcast from the `Any`
object to their own type - now, they can simply invoke `rqctx.context()`.
- This change additionally prevents a potential caller-side bug where multiple endpoints with
distinct "private context" objects are registered to a server. By expanding to concrete types,
this behavior is disallowed.
- Additionally, since dynamic casting to `Any` is no longer needed, the context object can be stored directly. This removes one layer of `Arc` for accessing server state.


dropshot/src/handler.rs
```diff
@@ -99,7 +100,14 @@ pub struct RequestContext {
     pub log: Logger,
 }
 
-impl RequestContext {
+impl<Context: ServerContext> RequestContext<Context> {
+    /**
+     * Returns the server context state.
+     */
+    pub fn context(&self) -> Arc<Context> {
```
smklein, line 107: Given this new mechanism of extracting the context, I actually wonder...

- Does it need to be wrapped in an Arc?
- When providing it back to a caller, do we need to provide **new** Arc object?

Seems like it should be possible to return either an `&Arc<Context>` or a `&Context` object directly.

cbiffle: Cloning the Arc on each dispatch has a performance impact. `&Arc` would be the smaller change -- code would retain the ability to make its own clone of the `Arc` if required.

However, I'd bet that current handler functions are dropping the `Arc` at the end, which means `&Context` would be fine -- it has the bonus of being a simpler, clearer type and exposing fewer implementation details of the server. So, that's what I'd nudge toward.

smklein: Agreed. I've replaced the usage here with `&Context` directly (in the callback, as well as in storage). Users can decide for themselves if they'd like to wrap this type in an `Arc`.

smklein:
So *functionally* I'm actually happy with this change, although I'll admit: I think the proc macro parsing-of-context needs some love before it should merge.

I've noticed that, as one example, passing type aliases for `Arc<RequestContext<Context>>` breaks the endpoint macro, as it's operating purely lexicographical and cannot expand the aliased identifier to the real time.

I'll look at the Extractor traits to see if I can massage the embedded context type out there in an alias-neutral way.

dropshot/src/handler.rs
davepacheco: I definitely like if we can drop the Arc here.  Does that work for API endpoint handler functions that are async and whose futures reference the context after they return?

smklein: > I definitely like if we can drop the Arc here. Does that work for API endpoint handler functions that are async and whose futures reference the context after they return?

This is up to the caller. If they'd like this behavior, they can supply arguments such that `Context=Arc<MyObject>`, or any other smart-pointer wrapping of choice. However, if callers don't need this behavior, they can just supply `Context=MyObject`. The `context` function will be monomorphized to return whichever variant the caller requested.

davepacheco: Makes sense.  Maybe we should have Context=Arc<Example> in the basic example, as I think that'll be the most common case by far.

smklein: I updated the "example context" to use asynchronous functions, but it *still* didn't need an Arc because the endpoint functions are also async, so the lifetime wasn't a problem. I think the case where the Arc is necessary is where the server wants to act on data **completely independently** of endpoint functions; I made another example which increments a counter on a timer.

(I basically just want the example to have a "real" reason for using the Arc. Added in `module-shared-context.rs`)

davepacheco (commented):
Nice!  Thanks for doing this.  I think the interface for consumers looks a lot cleaner.  I have some minor suggestions inline.

@ahl can you take a look at the user-facing change and the dropshot_endpoint changes?

Since this feels like kind of a big change, I'd love to get a +1 on the interface change from one or two others that have built something with dropshot like @jclulow @jessfraz or @arjenroodselaar.

I'd also love to see an update to at least one real consumer.  I can take a swing at doing this for oxide-api-prototype.  Let me know if you'd rather do it.

Thanks again for doing this!

dropshot/src/handler.rs
davepacheco: Sorry -- I was confused about the `Arc` around `RequestContext` vs. the user's context type.  This all makes more sense now!

arjenroodselaar:
Good change, this makes the interface more predictable. I actually had to wrestle for a bit a few days ago to get the private app context working in the flexlm web gui thing. If I read this PR correct that should not be the case anymore.

dropshot/src/api_description.rs
```diff
@@ -956,17 +961,17 @@ fn j2oas_object(
  * document may be generated via [`write()`](`OpenApiDefinition::write`) or
  * [`json()`](`OpenApiDefinition::json`).
  */
-pub struct OpenApiDefinition<'a> {
-    api: &'a ApiDescription,
+pub struct OpenApiDefinition<'a, Context: ServerContext> {
```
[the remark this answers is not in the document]
smklein: I *think* this is possible by breaking the `OpenApiDefinition`, and all related "needed traits" into a separate non-context-dependent trait.

However, this is currently really difficult, because `OpenApiDefinition` relies on a very concrete `HttpRouter` to do this work.

Unless folks strongly oppose, I agree, this would be better suited to a follow-up change (as it will require decoupling the notion of an "abstract endpoint" - no context, used w/openapi - from a "concrete" one)

dropshot/src/handler.rs
```diff
@@ -154,6 +162,18 @@ impl RequestContext {
     }
 }
 
+/**
+ * Helper trait for extracting the underlying Context type from an input
+ * request.
```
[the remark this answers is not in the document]
smklein: Yup, your understanding is accurate.

- Name updated to `RequestContextArgument`, with more detail in the comment.
- Re-export location moved next to `http::Method`.

dropshot/src/handler.rs
```diff
@@ -408,12 +434,15 @@ where
      * `FuncParams`, which allows us to use the type parameter below.
      */
     phantom: PhantomData<(FuncParams, ResponseType)>,
+
+    phantom2: PhantomData<Context>,
```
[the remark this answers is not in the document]
smklein: Oversight on my part - added to the earlier PhantomData. Good catch!

dropshot/tests/fail/bad_endpoint1.rs
```diff
@@ -1,5 +1,7 @@
 // Copyright 2020 Oxide Computer Company
 
+#![allow(unused_imports)]
```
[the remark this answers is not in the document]
smklein: It is. TL;DR: Unused import warnings that are omitted when compiler errors occur are frequently false positives, and in this test suite, they obscure the compiler error we're trying to repro.

Added a comment to this effect.

CHANGELOG.adoc
```diff
@@ -17,6 +17,10 @@ https://github.com/oxidecomputer/dropshot/compare/v0.4.0\...HEAD[Full list of co
 
 === Breaking Changes
 
+==== Generic Context
```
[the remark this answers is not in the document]
smklein: CHANGELOG updated with suggestions, endpoint example.

On the bright side, for a vast majority of those types, the context object will be inferred, so I think for most users, the endpoints are the only code that'll need updates.

dropshot/tests/fail/bad_endpoint1.rs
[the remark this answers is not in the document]
davepacheco: Nice.  Thanks for the comment!

jclulow:
I like the new API!  I took a swing at using the new API for a small scale dropshot server I created earlier: https://github.com/illumos/clag/compare/newdropshot#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fc

I haven't reviewed the change, but actual API looks great to me.  Thanks for doing this!

dropshot/tests/fail/bad_endpoint1.rs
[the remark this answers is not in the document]
ahl: Am I reading this correctly that the "unused import" warning only occurred in test 8? Did you change make it so that we would have seen these warnings in more/all tests?

My preference would be that the output of these tests actually reflect what users would see.

dropshot/tests/fail/bad_endpoint1.rs
```diff
@@ -1,5 +1,14 @@
 // Copyright 2020 Oxide Computer Company
 
+// These "bad endpoint" tests are intended to throw compiler errors.  When a
+// compiler error is thrown, imports which haven't yet successfully been used
+// may throw "unused import" warnings. This is often a false positive, as
+// the compiler error itself blocks intended usage.
+//
+// For this test (and other bad endpoint tests) disable the import warnings
+// so output can be more precisely compared with the expected compiler error.
+#![allow(unused_imports)]
```
ahl, line 10: It looks like previously only test 8 emitted "unused import" warnings. Did this PR cause more warnings?

I think it may be beneficial to see what actual users would see *including* unused import warnings.

dropshot_endpoint/src/lib.rs
```diff
@@ -53,6 +54,15 @@ struct Metadata {
 
 const DROPSHOT: &str = "dropshot";
 
+const ERR_MSG_USAGE: &str = "Try using the following endpoint format:
```
ahl, line 57: The wording here strikes me as overly coy. How about: "Endpoint handers must have the following signature"

dropshot_endpoint/src/lib.rs
```diff
@@ -53,6 +54,15 @@ struct Metadata {
...
+    async fn my_endpoint(
```
ahl, line 58: Should we fill in the endpoint name they used rather than calling it `my_endpoint`?

dropshot_endpoint/src/lib.rs
```diff
@@ -167,53 +195,21 @@ fn do_endpoint(
...
-                                }
-                                impl<T: ?Sized> TypeEq for T {
-                                    type This = Self;
-                                }
-                                fn need_arc_requestcontext<T>()
-                                where
-                                    T: ?Sized + TypeEq<This = #req>,
-                                {
-                                }
-                                need_arc_requestcontext::<#ty>();
```
ahl, line 184: do we have a test that skips the request context and just has extractors?

ahl (commented):
Good stuff! I left a couple of items for your consideration.

dropshot/tests/fail/bad_endpoint1.rs
smklein: Yeah, there are a couple spots where the warnings appear, where they didn't before. I believe this is due to the difference in parsing of the first argument:

- The old version of the proc macro *forced* the first argument to be `Arc<RequestContext>`, and emitted a stream from the proc_macro implementation in more cases, relying on the compiler to catch these issues after the lexicographic substitution of the macro completed.
- The new version needs to parse that first argument, and throws errors earlier for a couple cases: an endpoint without arguments or with a non-receiver argument will cause the macro to throw an error explicitly. I added two test cases to `dropshot_endpoint/src/lib.rs` to cover these cases.

So that's a long way of saying "yes, this change does make the unused import warning appear more often, but that's mostly because it's raising the error sooner".

If you view this as a problem, I can try to force the proc macro to emit tokens, even when I know they're invalid, but that feels kinda like the wrong direction IMO.

dropshot_endpoint/src/lib.rs
smklein: Oooooo apparently not, but I'll add one (I'm going to add it to `tests/fail/bad_endpoint...` since those handle output a little better for failure cases which occur after the macro finishes processing successfully).

The output is currently:

```
error[E0277]: the trait bound `dropshot::Query<QueryParams>: RequestContextArgument` is not satisfied
  --> $DIR/bad_endpoint9.rs:21:1
   |
21 | / #[endpoint {
22 | |     method = GET,
23 | |     path = "/test",
24 | | }]
   | |__^ the trait `RequestContextArgument` is not implemented for `dropshot::Query<QueryParams>`
   |
   = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

I don't *think* it's possible to catch this error within the `endpoint` macro, as it relies on type information for the first argument.

I updated the `RequestContextArgument` documentation to reiterate expected usage. If you don't think that's sufficient, I could look into emitting more structures-that-act-as-checks from the macro?

dropshot/tests/fail/bad_endpoint1.rs
ahl: That makes sense and I--agreed--don't think we should emit tokens in the face of an error.

Do you think it's desirable to suppress the warnings? I think I'd rather have the tests show what a user would see.

smklein: So, I can - but even separate from this PR, many of the `struct QueryParam` objects in the endpoint tests had the annotation `#[allow(dead_code)]`, presumably to avoid similar warnings. Should I remove those too?

dropshot_endpoint/src/lib.rs
ahl: If I understand you correctly (and I'm still kind of paging this back in), I believe the code that's deleted here was intended to test for type equality of that first argument. If for some reason this check no longer works with the more complex RequestContext type I think it's fine to leave as is.

... I did think this was a clever solution to checking for type equality, but cleverness alone isn't reason enough to keep it ;-)

dropshot_endpoint/src/lib.rs
```diff
@@ -54,14 +54,19 @@ struct Metadata {
 
 const DROPSHOT: &str = "dropshot";
 
-const ERR_MSG_USAGE: &str = "Try using the following endpoint format:
-    async fn my_endpoint(
+fn usage(err_msg: &str, fn_name: &str) -> String {
```
ahl, line 57: thanks for making this change. it looks like it boogered up the tests a tiny bit, but I think it's a bit nicer.

dropshot/tests/fail/bad_endpoint1.rs
ahl: Well. I'd argue those are dead code even if the endpoint were properly formed, but your point is well-taken and I'm happy with whatever you decide is the best path here.

dropshot_endpoint/src/lib.rs
smklein: Yeah, the problem is that the `req` parameter in the old version was hardcoded as `std::sync::Arc<RequestContext>`, but we don't actually know the intended type this time, as it has a nested generic. Previously, the code emitted a (if I'm reading this correctly) constant lambda filled with a trait, trait implementation, and function requiring that trait implementation, all to backpropagate errors about type equality. If we don't know the exact type we need, I'm a little worried the propagation of an error through this mechanism might deliver more confusion than clarity (since it's all to ask for a type that we can't know for certain), and I'm not sure of the best way to adapt it.

Besides, the current error (IMO) seems to provide just as much value, saying "the first argument needs to be a `RequestContextArgument`", which explains the problem, and elaborates in the documentation of `RequestContextArgument`.

dropshot/tests/fail/bad_endpoint1.rs
smklein: I think I stand with my original stance of continuing to `#![allow]` unused imports. If I was blocking **all** warnings, I would 100% agree with you, we would want to see the output, but frankly I'd be astonished if someone saw a "compiler error + unused import warning" and prioritized the import warning.

If we can agree on that, then I believe the warning suppression will just provide a better signal/noise ratio for the test output.

davepacheco:
I believe with https://github.com/oxidecomputer/oxide-api-prototype/pull/37 coded up, this is ready to go!

PR #87, opened 2021-02-16
attempt a CI build on illumos

jclulow:
WIP; do not merge!


papertigers:
Buildomat is probably much preferred here as I think it's capable of spinning up an illumos VM directly in AWS. My stuff will probably fill the free option for OSS out there as it's a macOS instance with virtualbox under the hood. I do plan to spend some time next week updating it. At which point moving it to the illumos org would also be preferred so we can put it in the GitHub marketplace or whatever it's called.

PR #89, opened 2021-02-25
CI could report version information

davepacheco:
This is a little redundant for the checks that already request a specific toolchain, but it's easy to do and gives more specific information.


davepacheco:
Confirmed that the checks on this PR include the expected version information.

PR #90, opened 2021-02-25
Implement std::error::Error for HttpError

smklein:
This implementation allows `HttpError` to be composed within a [thiserror](https://docs.rs/thiserror/1.0.24/thiserror/) error type.

### DIFF
--- a/dropshot/src/error.rs
+++ b/dropshot/src/error.rs
@@ -47,6 +47,8 @@
 use hyper::Error as HyperError;
 use serde::Deserialize;
 use serde::Serialize;
+use std::error::Error;
+use std::fmt;
 
 /**
  * `HttpError` represents an error generated as part of handling an API
@@ -282,5 +284,17 @@ impl HttpError {
                 .into(),
             )
             .unwrap()
+    }
+}
+
+impl fmt::Display for HttpError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "HttpError({}): {}", self.status_code, self.external_message)
+    }
+}
+
+impl Error for HttpError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        None
     }
 }


dropshot/src/error.rs
```diff
@@ -284,3 +286,15 @@ impl HttpError {
             .unwrap()
     }
 }
+
+impl fmt::Display for HttpError {
```
davepacheco, line 290: My only fear about this is that by impl'ing Display, does that make it easier for us to accidentally send this representation (which is essentially for debugging, not suitable for end users) out to end users?

This is a pretty weak fear -- if there's good reason to want to impl `Error` on this, that's fine.  Are there cases where you want `Error` so you can use this as a source of some other error?

dropshot/src/error.rs
```diff
@@ -284,3 +286,15 @@ impl HttpError {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "HttpError({}): {}", self.status_code, self.external_message)
+    }
+}
+
+impl Error for HttpError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
```
davepacheco, line 297: This might be fine for what you're doing now, but it seems like there are many cases where we could provide a more useful source error.  In oxide-api-prototype, I would expect a lot of errors could have a `source` that was an `ApiError`.

dropshot/src/error.rs
smklein: So... I'm not sure what you mean by "send this representation out to end users".

HttpError is *already* public, and all it's members are also public. This means end users can basically do whatever they want to with the object already, including modify it arbitrarily.

The implementation of "Display" is just showing some of those fields in an arbitrary way - but end users could 100% already access and print these fields themselves, if they'd like to do so.

And yeah, as mentioned in the PR description, [thiserror](https://docs.rs/thiserror/1.0.24/thiserror/) can only wrap types implementing `std::error::Error`.

dropshot/src/error.rs
smklein: While that might logically be true, it doesn't currently seem possible in the context of this object's members.

From what I can tell, [source](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) generally refers to "wrapping an sub-error" when crossing abstraction barriers - since `HttpError` doesn't contain an `ApiError`, nor does it have any way of accessing an `ApiError`, I think this would require modifying the contents (and construction) of `HttpErrors` to preserve the original error type.

dropshot/src/error.rs
smklein: If you're worried about someone depending on this crate depending on the specific display output... Yes, that's technically possible, but I'd find it pretty unlikely, since they're much more likely to directly depend on the members instead if they need that representation.

davepacheco: By "end user" here I mean a consumer of a dropshot-based HTTP API.  This might be an Oxide customer writing automation that hits the Oxide API, for example.  They don't necessarily know anything about Dropshot or Rust.  My worry is that someone working on Dropshot or a Dropshot consumer might use an HttpError in a context that causes it to be automatically formatted as a String and made part of an HTTP response, and the end user will see a message that makes no sense to them.  Today, that would be hard to do.  The `Display` trait makes it easier to use a an object in a context that expects a string.  But again, this a weak concern, and definitely outweighed by wanting

dropshot/src/error.rs
davepacheco: Yeah, I think that's right.  We could always add this later.

dropshot/src/error.rs
smklein: > By "end user" here I mean a consumer of a dropshot-based HTTP API. This might be an Oxide customer writing automation that hits the Oxide API, for example. They don't necessarily know anything about Dropshot or Rust.

I would not expect this representation to modify the HTTP representation itself; it's just a new behavior for someone already in possession of this type.  If a client isn't using Rust, this shouldn't produce anything visible.

> My worry is that someone working on Dropshot or a Dropshot consumer might use an HttpError in a context that causes it to be automatically formatted as a String and made part of an HTTP response, and the end user will see a message that makes no sense to them. Today, that would be hard to do. The `Display` trait makes it easier to use a an object in a context that expects a string. But again, this a weak concern, and definitely outweighed by wanting

If this is an extension of the concern that "we should be careful what we send over HTTP", I agree - but I don't think this is a spot where that could happen.

(Unless I'm misunderstanding something? Happy to modify the code to make this less risky, but I'm 100% sure this is what's concerning you)

smklein: Also, the HttpError constructors are *already* accepting arbitrary strings as inputs to construct messages. If we're worried about stringly-typed objects being passed around instead of structured data, that seems reasonable, but IMO that should be solved with error composition, not by avoiding the standard error types.

davepacheco: > If this is an extension of the concern that "we should be careful what we send over HTTP", I agree - but I don't think this is a spot where that could happen.

Yes, it's this.  Fair enough!

davepacheco (approved):

davepacheco:
I forgot -- we should update the changelog for this change as well.  (Obviously it's fine if that's a follow-on PR.)

PR #92, opened 2021-03-03
update CHANGELOG for recent notable changes

davepacheco:
I want to try to incorporate changelog updates with each change, but I forgot for a few of the recent ones.

CC @smklein @ahl

@ahl, I wasn't sure if I should include anything for #80 and #91.  We can always add it later if you think so.


smklein (approved):
Ah, apologies! This fell off my radar

davepacheco:
> Ah, apologies! This fell off my radar

No worries -- me too!

Issue #93, opened 2021-03-18
rename "master" branch to "main"


davepacheco:
If nothing else, it's helpful for this repo's primary branch to be consistent with other repos created with newer defaults.

If you have a local clone, these are the steps GitHub recommends to get past this change:

```
git branch -m master main
git fetch origin
git branch -u origin/main main
```

Otherwise, this should have no impact on you.

PR #94, opened 2021-03-18
update workflow for rename of default branch

davepacheco:


davepacheco:
As so often happens, I'm not sure why GitHub Actions are not running on this PR.

It doesn't seem like this change is _necessary_, even given #93.  Empirically, other PRs (on other repos) ran workflows configured for "master" even when they were created after the branch was renamed to "main".  And [this post](https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/) does say "All of your workflows that reference the old branch name will still work," though it's not totally clear what operations that's describing.

davepacheco:
The problem appears to have been the use of `$default-branch` instead of explicitly `main`.  I'm not sure why, given that the [official examples use `$default-branch` now](https://github.com/actions/starter-workflows/blob/main/ci/rust.yml), but using "main" does seem to work.

PR #100, opened 2021-03-23
use 32-bit types for pagination limits

davepacheco:
This change modifies the numeric type that's used for the "limit" in paginated endpoints.

Previously, this was a `NonZeroU64` in the externally-facing data structures, but only exposed internally as a `usize` (via `RequestContext::page_limit`).  If the conversion from `NonZeroU64` to `usize` failed, a 400 response was sent back to the client.

Why did it work this way?  I wanted `usize` internally because all of the internal consumers were backed by Rust collection iterators, which use `usize` to describe a number of elements (e.g., [take](https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.take)).  But I figured that using `usize` externally would cause the OpenAPI spec to change depending on the bitness of the server, which didn't seem right.  Still, this solution has a few issues:

* On any server where `usize::MAX` is smaller than `u64::MAX`, the OpenAPI spec could advertise a value larger than was supported.  On these servers, the error that a client would get for the range `usize::MAX` to `u64::MAX` is likely different than they would get for values above `u64::MAX`.
* When I went to add a consumer backed by a PostgreSQL-compatible database (CockroachDB, actually) I discovered that even [`u64` doesn't have a direct translation to PostgreSQL](https://docs.rs/tokio-postgres/0.7.0/tokio_postgres/types/trait.ToSql.html#types).  Indeed, [the largest native numeric type in PostgreSQL appears to be a signed 64-bit integer](https://www.postgresql.org/docs/current/datatype-numeric.html).

Neither of these is a big deal, but they're worth addressing.

This change modifies both the externally-facing and internally-facing types to `NonZeroU32`.  I chose this because it's large enough to be sufficient for all foreseeable practical purposes, but small enough to be satisfy the PostgreSQL consumer without additional validation.  This is still arbitrary, obviousy, but _any_ choice here, once it's big enough for practical purposes, is just a matter of making things most convenient for consumers.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -15,6 +15,10 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.5.1\...HEAD[Full list of commits]
 
+=== Breaking Changes
+
+* https://github.com/oxidecomputer/dropshot/pull/100[#100] The type used for the "limit" argument for paginated resources has changed.  This limit refers to the number of items that an HTTP client can ask for in a single request to a paginated endpoint.  The limit is now 4294967295, where it may have previously been larger.  This is not expected to affect consumers because this limit is far larger than practical.  For details, see #100.
+
 == 0.5.1 (released 2021-03-18)
 
 https://github.com/oxidecomputer/dropshot/compare/v0.5.0\...v0.5.1[Full list of commits]
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -132,7 +132,7 @@ impl<Context: ServerContext> RequestContext<Context> {
              * Compare the client-provided limit to the configured max for the
              * server and take the smaller one.
              */
-            .map(|limit_nzu32| min(limit_nzu32, server_config.page_max_nitems))
+            .map(|limit| min(limit, server_config.page_max_nitems))
             /*
              * If no limit was provided by the client, use the configured
              * default.
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -62,13 +62,12 @@ use serde::Serialize;
 use slog::Logger;
 use std::cmp::min;
 use std::collections::BTreeMap;
-use std::convert::TryFrom;
 use std::fmt::Debug;
 use std::fmt::Formatter;
 use std::fmt::Result as FmtResult;
 use std::future::Future;
 use std::marker::PhantomData;
-use std::num::NonZeroUsize;
+use std::num::NonZeroU32;
 use std::sync::Arc;
 
 /**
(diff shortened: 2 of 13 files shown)


davepacheco:
Blocked on #101.

davepacheco:
Note that the examples and tests still use in-memory data structures that need the limit to be a `usize`.  I have just thrown `as usize` onto the limit.  This isn't great: these consumers should probably use `usize::try_from(...).map_err(f)?`, where `f` is a function that returns a 400-level error with an appropriate message.  This didn't seem worthwhile for the tests or examples, but if people feel it'd be misleading users in the examples then I can substitute something like the above.

CHANGELOG.adoc
```diff
@@ -15,6 +15,10 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.5.1\...HEAD[Full list of commits]
 
+=== Breaking Changes
+
+* https://github.com/oxidecomputer/dropshot/pull/100[#100] The type used for the "limit" argument for paginated resources has changed.  This limit refers to the number of items that an HTTP client can ask for in a single request to a paginated endpoint.  The limit is now 4294967295, where it may have previously been larger.  This is not expected to affect consumers because this limit is far larger than practical.  For details, see #100.
```
bnaecker, line 20: I'm not sure if it would cloud the issue, but I might add a parenthetical comment to the actual limit, something like `4294967295 (or u32::MAX - 1)`. The linked issue would have this detail, but it might be helpful to know here something about the _type_ that client code could use to represent this. Just a thought, I'm really not sure it's worth it.

dropshot/src/handler.rs
```diff
@@ -129,32 +128,11 @@ impl<Context: ServerContext> RequestContext<Context> {
...
-            })?
             /*
              * Compare the client-provided limit to the configured max for the
              * server and take the smaller one.
              */
-            .map(|limit_usize| {
-                let limit_nzusize = NonZeroUsize::new(limit_usize).unwrap();
-                min(limit_nzusize, server_config.page_max_nitems)
-            })
+            .map(|limit_nzu32| min(limit_nzu32, server_config.page_max_nitems))
```
bnaecker, line 135: Seems like we could remove the type suffix here, since there's no conversion?

bnaecker (approved):
Pretty straightforward, looks great! Couple of minor points, but nothing to block a merge when you're ready.

CHANGELOG.adoc
davepacheco: Fair point.  In this context, I was really only trying to communicate enough information to let the reader determine _whether_ this might affect them.  I think if it even might affect you, you probably want to read the issue details.  What do you think of that?  (Should I make that clearer?)

dropshot/src/handler.rs
davepacheco: Good point.  Fixed in ed77606ea778827f4fd456cf622f4be5f4ea170d.

CHANGELOG.adoc
bnaecker: Yeah, this is exactly what I meant by "clouding the issue," so I'm on board with pointing consumers to more detail in the linked issue :)

davepacheco: 👍 Thanks for taking a look!

smklein (approved):
FYI, I patched the openAPI tests to use u32 instead of u64.

LGTM.

davepacheco:
Thanks for the reviews!

PR #103, opened 2021-04-13
implement graceful shutdown on Drop

davepacheco:
Previously if the Dropshot server was dropped prior to closing it, it would panic.  This was intended to guide people to using `close()` for graceful shutdown.  I've now run into this a lot more in failing test suites, where it's way more disruptive than it is helpful because the program panics while panicking, and you're left with very little debug information.  So I'm proposing that we do attempt a graceful shutdown in `Drop`.  I don't consider this a breaking change because previously, a program that ran into this behavior would panic anyway.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -15,6 +15,8 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.5.1\...HEAD[Full list of commits]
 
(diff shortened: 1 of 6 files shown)


smklein (approved):
LGTM, though I think you might need to hit the tests with a `TRYBUILD=overwrite` hammer

PR #105, opened 2021-04-16
fix #104 duplicated inline response type schemas

### LINKED ISSUE #104: top-level types are always inline, so duplicated in the schema
## The problem

In oxidecomputer/omicron#64 @david-crespo reported that when running openapi-generator for TypeScript on the Omicron API schema, he was getting duplicate type definitions like `ApiRackView` and `ApiRackView1`.  The types were exactly the same.  This API has two different endpoints that use this type: one returns an `ApiRackView`, and the other returns a list of `ApiRackView`.  It appears that the dropshot-generated OpenAPI spec inlines the top-level type in the response, but it generates references for subtypes.

## Test case

I was able to replicate this by modifying test_openapi.rs with this change:

```diff
diff --git a/dropshot/tests/test_openapi.rs b/dropshot/tests/test_openapi.rs
index 97daf80..3af7e5f 100644
--- a/dropshot/tests/test_openapi.rs
+++ b/dropshot/tests/test_openapi.rs
@@ -136,6 +136,36 @@ async fn handler7(
     unimplemented!();
 }
 
+#[derive(JsonSchema, Serialize)]
+struct NeverDuplicatedTopLevel {
+    b: NeverDuplicatedNextLevel,
+}
+
+#[derive(JsonSchema, Serialize)]
+struct NeverDuplicatedNextLevel {
+    v: bool,
+}
+
+#[endpoint {
+    method = PUT,
+    path = "/dup1",
+}]
+async fn handler8(
+    _rqctx: Arc<RequestContext<()>>,
+) -> Result<HttpResponseOk<NeverDuplicatedTopLevel>, HttpError> {
+    unimplemented!();
+}
+
+#[endpoint {
+    method = PUT,
+    path = "/dup2",
+}]
+async fn handler9(
+    _rqctx: Arc<RequestContext<()>>,
+) -> Result<HttpResponseOk<NeverDuplicatedTopLevel>, HttpError> {
+    unimplemented!();
+}
+
 fn make_api() -> Result<ApiDescription<()>, String> {
     let mut api = ApiDescription::new();
     api.register(handler1)?;
@@ -145,6 +175,8 @@ fn make_api() -> Result<ApiDescription<()>, String> {
     api.register(handler5)?;
     api.register(handler6)?;
     api.register(handler7)?;
+    api.register(handler8)?;
+    api.register(handler9)?;
     Ok(api)
 }

```

Here's how that changes the generated schema for that test:

```diff
dap@zathras dropshot $ git diff dropshot/tests/test_openapi.json
diff --git a/dropshot/tests/test_openapi.json b/dropshot/tests/test_openapi.json
index 1233f79..688000d 100644
--- a/dropshot/tests/test_openapi.json
+++ b/dropshot/tests/test_openapi.json
@@ -26,6 +26,58 @@
         }
       }
     },
+    "/dup1": {
+      "put": {
+        "operationId": "handler8",
+        "responses": {
+          "200": {
+            "description": "successful operation",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "title": "NeverDuplicatedTopLevel",
+                  "type": "object",
+                  "properties": {
+                    "b": {
+                      "$ref": "#/components/schemas/NeverDuplicatedNextLevel"
+                    }
+                  },
+                  "required": [
+                    "b"
+                  ]
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/dup2": {
+      "put": {
+        "operationId": "handler9",
+        "responses": {
+          "200": {
+            "description": "successful operation",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "title": "NeverDuplicatedTopLevel",
+                  "type": "object",
+                  "properties": {
+                    "b": {
+                      "$ref": "#/components/schemas/NeverDuplicatedNextLevel"
+                    }
+                  },
+                  "required": [
+                    "b"
+                  ]
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "/impairment": {
       "get": {
         "operationId": "handler6",
@@ -288,6 +340,17 @@
   },
   "components": {
     "schemas": {
+      "NeverDuplicatedNextLevel": {
+        "type": "object",
+        "properties": {
+          "v": {
+            "type": "boolean"
+          }
+        },
+        "required": [
+          "v"
+        ]
+      },
       "ResponseItem": {
         "type": "object",
         "properties": {
```

We see that the top-level type shows up twice, inline in both places, while the next level type gets a reference instead.

## Analysis

When you use the `endpoint` macro, we generate a `From<...> for ApiEndpoint` for this specific handler

```rust
#[allow(non_camel_case_types, missing_docs)]
#[doc = "API Endpoint: handler9"]
struct handler9 {}
#[allow(non_upper_case_globals, missing_docs)]
#[doc = "API Endpoint: handler9"]
const handler9: handler9 = handler9 {};
impl From<handler9>
    for dropshot::ApiEndpoint<
        <Arc<RequestContext<()>> as dropshot::RequestContextArgument>::Context,
    >
{
    fn from(_: handler9) -> Self {
        async fn handler9(
            _rqctx: Arc<RequestContext<()>>,
        ) -> Result<HttpResponseOk<NeverDuplicatedTopLevel>, HttpError> {
            ::core::panicking::panic("not implemented");
        }
        dropshot::ApiEndpoint::new(
            "handler9".to_string(),
            handler9,
            dropshot::Method::PUT,
            "/dup2",
        )
    }
}
```

(That's from `cargo expand -p dropshot --test=test_openapi` with my changes.)

When you register the handler function, we use that `From` impl (well, the corresponding `Into`) here:
https://github.com/oxidecomputer/dropshot/blob/e04241c9eb6a19bd9b0b42718adad8b6d7bab29c/dropshot/src/api_description.rs#L222-L226

We see above that that invokes `ApiEndpoint::new()`, which winds up invoking `metadata()` on the response type:
https://github.com/oxidecomputer/dropshot/blob/e04241c9eb6a19bd9b0b42718adad8b6d7bab29c/dropshot/src/api_description.rs#L60

`ResponseType` is `HttpResponseOk` here, and the `metadata()` impl comes from `HttpTypedResponse`:
https://github.com/oxidecomputer/dropshot/blob/e04241c9eb6a19bd9b0b42718adad8b6d7bab29c/dropshot/src/handler.rs#L1140-L1149

What's important here is that we supply `Body::json_schema` as the `schema` function here:
https://github.com/oxidecomputer/dropshot/blob/e04241c9eb6a19bd9b0b42718adad8b6d7bab29c/dropshot/src/handler.rs#L1144

Again, that's all at registration time.  Later when we generate the schema, we use that `schema` function to generate a schema for the response body:
https://github.com/oxidecomputer/dropshot/blob/e04241c9eb6a19bd9b0b42718adad8b6d7bab29c/dropshot/src/api_description.rs#L524-L527

Let's take a closer look at the derived `JsonSchema` impls.  For the top-level type, it looks like this (back to `cargo expand`):

```rust
impl schemars::JsonSchema for NeverDuplicatedTopLevel {
    fn schema_name() -> std::string::String {
        "NeverDuplicatedTopLevel".to_owned()
    }
    fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
        {
            let schema = {
                let mut schema_object = schemars::schema::SchemaObject {
                    instance_type: Some(schemars::schema::InstanceType::Object.into()),
                    ..Default::default()
                };
                <NeverDuplicatedNextLevel as schemars::JsonSchema>::add_schema_as_property(
                    gen,
                    &mut schema_object,
                    "b".to_owned(),
                    None,
                    true,
                );
                schemars::schema::Schema::Object(schema_object)
            };
            gen.apply_metadata(schema, None)
        }
    }
}
```

The schema for `NeverDuplicatedNextLevel` must be added via `add_schema_as_property`, which winds up invoking not `json_schema()` on `NeverDuplicatedNextLevel`, but [gen.subschema_for()](https://docs.rs/schemars/0.8.3/schemars/gen/struct.SchemaGenerator.html#method.subschema_for).  That's documented as:

> Generates a JSON Schema for the type T, and returns either the schema itself or a $ref schema referencing T’s schema.
> 
> If T is referenceable, this will add T’s schema to this generator’s definitions, and return a $ref schema referencing that schema. Otherwise, this method behaves identically to JsonSchema::json_schema.

This explains why the non-top-level schemas get references: JsonSchema internally uses a function that tries to add references if possible.  The top-level ones don't get references because we're using `json_schema()` directly, which always generates a new, non-reference schema.

davepacheco:
Fixes #104.


dropshot/src/handler.rs
```diff
@@ -1141,13 +1141,20 @@ where
...
             success: Some(T::STATUS_CODE),
             description: Some(T::DESCRIPTION.to_string()),
         }
     }
 }
+
+fn make_subschema_for<T: JsonSchema>(
+    gen: &mut schemars::gen::SchemaGenerator,
+) -> schemars::schema::Schema {
+    gen.subschema_for::<T>()
```
ahl, line 1155: nice find; I hadn't seen that method previously

ahl (approved):
ship it

ahl (approved):
ship it

dropshot/src/handler.rs
davepacheco: Heh, I only found it when digging into why non-top-level types _did_ create references instead of inline schemas.

davepacheco:
There were a bunch of changes after @ahl's review, but they were mostly churn.  The net change from 9ffb17d to 5873dbb5e382c301c4e7858dd640602da6b96474 (the current tip of this branch) is (1) I applied the change to `TypedBody` and (2) I added test cases for duplicated types in query parameters, request bodies, and a mix of places.

PR #107, opened 2021-05-03
update openapiv3 -> 0.5.0

ahl:


davepacheco:
Cool.  I'm guessing there's no changelog update needed for this one?

ahl:
Correct; just an internal change. Is there some way I should denote that?

davepacheco:
I don't think there's any need.  The changelog already has a link to the raw list of commits for each release for people who want to know about more than what we've highlighted.

PR #108, opened 2021-05-06
Allow for serde_json::Value in body/return types

ahl:
Currently we fail on an `unreachable!()`: not quite true--obviously--since I forgot about permissive types such as `serde_json::Value`. This fix adds handling for that type. It also improves the panic message for the empty set type.

Note that I ran the resulting schema through openapi-generator and the model looks exactly as we would hope:

```rust
pub struct BodyParam {
    #[serde(rename = "_any")]
    pub _any: Option<serde_json::Value>,
    #[serde(rename = "_x")]
    pub _x: String,
}
```

(albeit with some extraneous renames... generated code, what can you do)

### DIFF
--- a/dropshot/src/api_description.rs
+++ b/dropshot/src/api_description.rs
@@ -640,7 +640,21 @@ fn j2oas_schema(
     schema: &schemars::schema::Schema,
 ) -> openapiv3::ReferenceOr<openapiv3::Schema> {
     match schema {
-        schemars::schema::Schema::Bool(_) => unreachable!(),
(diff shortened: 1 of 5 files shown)


davepacheco:
LGTM -- fixes the bug I was seeing.

davepacheco:
Seems like we can probably skip a change log entry for this, too.

ahl:
We could add: the use of permissive schemas (e.g. `serde_json::Value`) in API types is allowed

PR #112, opened 2021-05-29
Add README.md for crates.io

### LINKED ISSUE #111: Documentation on crates.io
I just stumbled on dropshot and when I went to [crates.io](https://crates.io/crates/dropshot) to check it out, it said `dropshot v0.5.1 appears to have no README.md file`.

I see that docs.rs has the dropshot docs.

Will it be possible to investigate what is wrong with crates.io?

ahl:
Fixes #111 

Adds the README and some annotations for crates.io

### DIFF
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -6,6 +6,9 @@ authors = ["David Pacheco <dap@oxidecomputer.com>"]
 edition = "2018"
 license = "Apache-2.0"
 repository = "https://github.com/oxidecomputer/dropshot/"
+readme = "README-crates.io.md"
+keywords = ["rest", "openapi", "async"]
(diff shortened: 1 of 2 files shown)


davepacheco (approved):
For historical reference, here are some points we discussed offline:

- We want to keep asciidoc for the top-level README.  Crates.io doesn't support asciidoc.  This was probably not the only problem causing #111.  The README is at the top level of the repo, but not contained in the "dropshot" package, so crates.io wouldn't have found it anyway.
- We could have added a README.md containing the content in this PR, but people (and tools) might miss the more complete README.adoc.  So we picked another filename and added the "readme" property to dropshot/Cargo.toml. 
- The Rust docs should be a pretty good introduction to the crate for users.  This is different content than what's in the README.

PR #113, opened 2021-06-01
add maintainer docs

davepacheco:


MAINTAINERS.adoc
```diff
@@ -0,0 +1,129 @@
...
+* For applications (policy "increase"), it updates Cargo.lock to specify the newer version.  It does not always update Cargo.toml.
+* For libraries (policy https://github.com/dependabot/dependabot-core/blob/f2167589e8210d53ac64e784a77c6aa4a14e31d6/cargo/lib/dependabot/cargo/update_checker/requirements_updater.rb#L45-L50[seems to be "increase-if-necessary" for Rust]), it updates Cargo.toml only when the semver constraint in Cargo.toml does not already allow the newer package version.
+
+For applications (such as Oxide's internal Omicron project), we point Dependabot at the root of the https://doc.rust-lang.org/cargo/reference/workspaces.html[Cargo workspace] (the root of the repository).  This causes Dependabot to find Cargo.lock, treat the package as an application, and update Cargo.lock and the Cargo.toml files for all packages in the workspace.
+
+We do not do this in Dropshot.  We _do_ <<_checking_in_cargo_lock_file,check in a Cargo.lock file>>, but we point Dependabot at the individual packages in the Dropshot workspace.  Dependabot does not find the Cargo.lock file and so treats Dropshot as a library, only updating Cargo.toml when _necessary_ to accommodate a new dependency.
```
smklein, line 33: Earlier we made the assertion the following regarding Lockfiles:

> Since the CI tests always run against a particular commit, this allows us to know exactly which dependencies were used for a particular CI run. It also communicates to consumers what we've tested.

If dependabot can update Cargo.toml files without updating Cargo.lock files, hasn't this assertion been broken? If dependabot updates a rev, but Cargo.lock is left out of date, then won't CI runs be inconsistent?

If we need to add the caveat that "Cargo.lock shows what dependencies are used, unless dependabot changed something, in which case it might be out-of-date", IMO the value of keeping a lockfile around drops significantly.

MAINTAINERS.adoc
```diff
@@ -0,0 +1,129 @@
...
+... Use `git log -1` to see what commit you're on.
+... Look up that commit on GitHub.  It will have a URL like `https://github.com/oxidecomputer/dropshot/commit/COMMIT_SHA` (with an actual `COMMIT_SHA`).  CI succeeded if there's a green check at the left of the commit message.  You can click the green check to be sure you're looking at the right thing.  It should say "All checks have passed".
+. Figure out the tag used for the most recent release.  The easiest is probably to view https://github.com/oxidecomputer/dropshot/releases[the recent releases].   We'll call this `PREV_RELEASE_TAG` in the next step.
+. Figure out what kind of update this is relative to the last one released.  https://github.com/sunng87/cargo-release/blob/master/docs/reference.md#bump-level[See `cargo release` docs for the available levels.]  We're usually doing a "patch", "minor", or "major" release.  We'll call this "RELEASE_KIND".  (According to the docs, you can also use a specific version number that you want to release here.)
+. Do the release.  If you're daring, just run:
++
+[source,text]
+----
+$ cargo release --prev-tag-name=PREV_RELEASE_TAG -vv RELEASE_KIND
```
smklein, line 74: Might be worth mentioning to:

```bash
$ cargo install cargo-release
```

first?

MAINTAINERS.adoc
```diff
@@ -0,0 +1,129 @@
...
++
+Inspect the resulting commits and the new tags from above.  There should be four commits: one to each of "dropshot" and "dropshot_endpoint" that sets the version to the one you specified, and one that sets the version to the _next_ version.  For example, if the current version is 0.5.1 and you're publishing 0.5.2, the repo will start at 0.5.2-pre, you'll see one commit that sets it to 0.5.2, and then you'll see one that sets it to 0.5.3-pre.
++
+Then undo the above step:
++
+[source,text]
+----
+$ git reset --hard THE_COMMIT_YOU_STARTED_WITH
+$ git tag -d THE_NEW_TAG
+----
```
smklein, line 99: Since these two steps cancel each other out, why bother doing them? Doesn't the dry run provide enough information about what will happen?

MAINTAINERS.adoc
```diff
@@ -0,0 +1,129 @@
...
++
+and do the release again with the publish step:
++
+[source,text]
+----
+$ cargo release --skip-push --prev-tag-name=PREV_RELEASE_TAG -vv minor
+----
++
+At this point, the new crates should be published to crates.io.  Check and
+push the commits and the tags.
```
smklein, line 109: To make sure I'm understanding correctly: The crate is published *before* the commits are pushed? Is this because we're creating tags and publishing, but explicitly passing `--skip-push`?

Is there a reason we're avoiding letting `cargo release` do this for us?

MAINTAINERS.adoc
davepacheco: Yeah, that's a good point.  Better options here are probably:

- don't check in Cargo.lock
- point dependabot at the whole workspace and treat it like an application
- point dependabot at the whole workspace and try to configure it to treat it like a library.  This seems best to me, but I'm not super optimistic based on the quality of the docs and the time I spent reverse engineering this last time.

MAINTAINERS.adoc
davepacheco: Good idea.

MAINTAINERS.adoc
davepacheco: If I were going to drop either the `--dry-run` step or the `--skip-push --skip-publish` step, I would skip the `--dry-run` because it doesn't provide any information over the other one.  I believe I saw cases (while working out the config) where the dry run succeeded and the `--skip-push --skip-publish` failed.

I think what you're getting at (and I kind of alluded to later without saying it) is that there's a bunch of fear in this procedure.  I think that fear stems from my initial experiments with `cargo-release` resulting in it doing things I found very surprising.  I don't remember more than that, but the lesson I internalized was that I wanted to see _exactly_ what it was doing before committing to anything.  I expect others might not be so paranoid about it, which is why I started with the "if you're daring" one-liner, but kept in enough detail for somebody to do it very carefully if they wanted.

MAINTAINERS.adoc
davepacheco: That is correct.  The reason is what I mentioned above: I wanted to be able to check the tool's output as much as possible before committing to it.  Separating this allows one to check them before pushing.  (Yes, we may have already published, but that's why I also do a the non-publishing version first, too.)

The one-liner at the top of this section has `cargo release` do it for you.

MAINTAINERS.adoc
davepacheco: It turns out that the docs are actually correct:

> The property '#/updates/0/versioning-strategy' value "increase-if-necessary" did not match one of the following values: lockfile-only, auto

From looking at the source a few weeks ago, the "auto" option decides whether it's an app or a library based on the presence of Cargo.lock (Cargo.lock present => application).  So if we use "auto", dropshot will be treated like an application, which means dependabot will be somewhat aggressive about updating dependencies.  If we chose "lockfile-only", I believe it would bump Cargo.lock as _compatible_ updates to dependencies were published, but never update Cargo.toml (which largely defeats the point).

I think I'd like to try letting dependabot treat dropshot like an application.  The downside is that if we have a dep version like 1.2.3, and 1.3 comes out, it will update Cargo.toml to specify 1.3, even though we might be fine with 1.2.3 _and_ 1.3, so we're imposing additional constraints on our consumers.  I'm hopeful that we can avoid this problem by avoiding such specific dependencies in Cargo.toml unless we really need them.  That is: if we leave our dep at 1.2, and 1.3 is compatible with that, dependabot won't update Cargo.toml.  (I'd expect it to update Cargo.lock, and that's fine and probably useful.)

MAINTAINERS.adoc
smklein: Ah, yeah, at first read I think I somehow missed that "if you want to go all-out, use this one liner" option. Makes sense to have a more intensive breakdown to see all the incremental changes, and then folks can rely on the shorthand once they (or the collective we) trust the process.

smklein (approved):
Thanks for the docs / updates!

davepacheco:
I just sync'd this up, updated Cargo.lock, and fixed the docs for the Dependabot change I made (discussed above).  I plan to land this shortly.

PR #116, opened 2021-06-04
Unused endpoints should produce a lint warning

### LINKED ISSUE #114: Should warn or error on unregistered handlers
If you write an `endpoint` function and forget to `api.register` it, the server will 404 on those routes, but it's not easy to figure out the problem. It would be helpful if there was a compiler error or warning for unregistered route handlers.

ahl:
Fixes #114

### DIFF
--- a/dropshot/tests/fail/unused_endpoint.stderr
+++ b/dropshot/tests/fail/unused_endpoint.stderr
@@ -1,0 +1,11 @@
+error: constant is never used: `unused_endpoint`
+  --> $DIR/unused_endpoint.rs:23:10
+   |
+23 | async fn unused_endpoint(
+   |          ^^^^^^^^^^^^^^^
+   |
+note: the lint level is defined here
+  --> $DIR/unused_endpoint.rs:5:9
+   |
+5  | #![deny(dead_code)]
+   |         ^^^^^^^^^
--- a/dropshot_endpoint/src/lib.rs
+++ b/dropshot_endpoint/src/lib.rs
@@ -212,8 +212,18 @@ fn do_endpoint(
         })
         .collect::<Vec<_>>();
 
+    // For reasons that are not well understood unused constants that use the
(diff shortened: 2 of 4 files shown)


ahl:
@smklein can I ask you to take a quick look? I think you've been in this vicinity before

dropshot_endpoint/src/lib.rs
```diff
@@ -212,8 +212,18 @@ fn do_endpoint(
         })
         .collect::<Vec<_>>();
 
+    // For reasons that are not well understood unused constants that use the
```
smklein, line 215: Spooky, but at least we have the negative test.

dropshot/tests/fail/unused_endpoint.stderr
```diff
@@ -0,0 +1,11 @@
+error: constant is never used: `unused_endpoint`
```
smklein, line 1: Since the behavior is "constant is never used", instead of "endpoint not registered", this presumably wouldn't fire with a `pub` symbol, right?

(I think that's probably okay, we can't catch everything, but just checking my understanding)

smklein (approved):
Change seems reasonable, test seems useful - hopefully this helps more dropshot users.

Although I doubt it'll hit anyone hard, this is technically a breaking change - we should probably add a note in the changelog.

dropshot/tests/fail/unused_endpoint.stderr
ahl: Yeah. Good catch. If the function is `pub` then I'm not sure there's much we can do. This is unfortunate as one is more likely to lose track of an endpoint in a larger program and one is also more likely to have endpoints in several files in larger programs.

Perhaps in the future we could consider adding endpoints to a distributed slice (https://github.com/dtolnay/linkme) to avoid manual registration entirely.

At least this change helps in simple cases as folks ramp up on Dropshot...

PR #117, opened 2021-06-09
Use `tokio::sync::RwLock` instead of `futures::lock::Mutex` to package the `request` in `RequestContext`.

denis2glez:
Unlike `Mutex`, `RwLock` allows any number of readers to acquire the lock as long as a writer is not holding the lock.

### DIFF
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -49,7 +49,7 @@ use crate::pagination::PAGINATION_PARAM_SENTINEL;
 
 use async_trait::async_trait;
 use bytes::Bytes;
-use futures::lock::Mutex;
+use tokio::sync::RwLock;
 use http::StatusCode;
 use hyper::Body;
 use hyper::Request;
@@ -91,7 +91,7 @@ pub struct RequestContext<Context: ServerContext> {
     /** shared server state */
     pub server: Arc<DropshotState<Context>>,
     /** HTTP request details */
-    pub request: Arc<Mutex<Request<Body>>>,
+    pub request: Arc<RwLock<Request<Body>>>,
     /** HTTP request routing variables */
     pub path_variables: BTreeMap<String, String>,
     /** unique id assigned to this request */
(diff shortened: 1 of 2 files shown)


smklein:
Hey, thanks for the contribution.

First of all - as you pointed out - RWLock *does* allow more readers to acquire the lock, as long as writers aren't present. However, it's worthwhile asking ourselves: if RWLock exists, and has this extra feature, why would anyone bother using a Rust Mutex?

The performance differences between the two have subtleties that vary across a handful of axes:

- *Workload*: This is probably the most notable factor when making a choice between locking primitives. In very general, handwavey terms, RWLocks **usually** have a higher constant overhead than traditional mutexes, but they provide more concurrency for workloads that are dominated by concurrent reads.

It's not clear to me that this use case in Dropshot is dominated by concurrent reads.

- *Read/Write Priority*: The Tokio [RwLock documents](https://docs.rs/tokio/1.6.1/tokio/sync/struct.RwLock.html) that it is write-preferring, such that writes have a preference over readers to avoid starvation.

> The priority policy of Tokio’s read-write lock is fair (or write-preferring), in order to ensure that readers cannot starve writers. Fairness is ensured using a first-in, first-out queue for the tasks awaiting the lock; if a task that wishes to acquire the write lock is at the head of the queue, read locks will not be given out until the write lock has been released. This is in contrast to the Rust standard library’s std::sync::RwLock, where the priority policy is dependent on the operating system’s implementation.

- *Per-platform differences*: The implementation of RWLocks typically varies by OS primitives provided, meaning that performance may differ between OS targets.

With all this said...

... identifying if this change is a "win" or "loss" from a performance point-of-view is difficult to analyze without profiling.

If you're interested in pushing this change, could I ask for the following:

- Evidence that you believe this workload is read-dominated? (I was under the impression that this path was fairly serial anyway, meaning that the constant overhead of locking would be a more prominent factor than hypothetical concurrency. https://github.com/oxidecomputer/dropshot/pull/106 elaborates on this further, making some changes that avoid the need for locking altogether).
- Profiling data that demonstrates an improvement?

denis2glez:
@smklein, thanks for your thorough reply.
Actually, my first impression was that this workload was dominated by concurrent reads, but then I started to conclude that this path was more linear indeed. I certainly overlooked #106 :facepalm:.
I'll keep looking around... closing this PR.

PR #118, opened 2021-06-22
Fix handling of escape codes in paths

ahl:
Previously we haven't had any special handling for HTTP path escape codes. This implicitly required consumers to do all their own decoding... which they surely were not. This PR adds proper handling for input paths i.e. paths as the results of a query, and it enforces stricter requirements for paths specified by consumers for endpoints. For example, we strip out "//" for a query, but `panic!()` if a consumer were to specify a path that contained "//" (note we also panic if a consumer tries to register two handlers for the same path).


dropshot/src/api_description.rs
```diff
@@ -1241,6 +1241,29 @@ mod test {
...
+            path = "I don't start with a slash"
+        }]
+        async fn test_twobodies_handler(
+            _: Arc<RequestContext<()>>,
+        ) -> Result<Response<Body>, HttpError> {
+            unimplemented!();
+        }
+
+        let mut api = ApiDescription::new();
+        let error = api.register(test_twobodies_handler).unwrap_err();
```
smklein, line 1259: "test_twobodies_handler" does not seem like the right name for this... how about:

```suggestion
        async fn bad_path_handler(
            _: Arc<RequestContext<()>>,
        ) -> Result<Response<Body>, HttpError> {
            unimplemented!();
        }

        let mut api = ApiDescription::new();
        let error = api.register(bad_path_handler).unwrap_err();
```

dropshot/src/api_description.rs
```diff
@@ -1241,6 +1241,29 @@ mod test {
...
+        assert_eq!(
+            error,
+            "only one body extractor can be used in a handler (this function \
+             has 2)"
```
smklein, line 1263: Is this assertion dead code? The function doesn't seem to have any body extractors, right?

dropshot/src/router.rs
```diff
@@ -464,9 +481,34 @@ impl<'a, Context: ServerContext> Iterator for HttpRouterIter<'a, Context> {
...
+ *    5.2).
+ *
+ * As dropshot requests are necessarily made with absolute (rather than
+ * relative) targets (any resolution presumed to have been done by the client)
+ * we make no special interpretation of dot-segments. In short, the strings "."
+ * and ".." may appear as segments (and thus as the values for path-derived
+ * variables). Consumers should take special care concerning these values when
+ * using Dropshot to serve files. Absent proper input validation, it could be
+ * possible for requests to retrieve files not-intended to be exposed (e.g.
+ * "GET /../../../etc/passwd").
```
smklein, line 507: Given the position that "clients are responsible for correctly canonicalizing . and ..", I'm actually kinda surprised with the conclusion that "Dropshot will ignore all that stuff, and hope the client got it right".

Couldn't we explicitly ban "." and ".." in path segments? If we throw an error on their discovery, we'd at least halt "known-bad-client behavior", with minimal effort (since we can still hold the stance "Dropshot won't resolve it for you").

dropshot/src/router.rs
```diff
@@ -501,7 +539,41 @@ pub fn path_to_segments(path: &str) -> Vec<&str> {
...
      * ourselves.
      */
-    path.split('/').filter(|segment| !segment.is_empty()).collect::<Vec<_>>()
+    let mut ret = Vec::new();
+    for segment in path.0.split('/') {
+        if !segment.is_empty() {
+            ret.push(percent_decode_str(segment).decode_utf8()?.to_string());
+        }
+    }
+    Ok(ret)
```
smklein, line 548: Completely optional nit, if we want to have one less `mut`:

```suggestion
   Ok(
     path.0.split('/')
       .filter(|segment| !segment.is_empty())
       .map(|segment| {
         Ok(percent_decode_str(segment).decode_utf8()?.to_string())
       })
       .collect()
   );
```

dropshot/src/router.rs
```diff
@@ -139,6 +141,19 @@ impl PathSegment {
     }
 }
 
+/**
+ * Wrapper for a path that's the result of user input i.e. an HTTP query.
+ * We use this type to avoid confusion with paths used to define routes.
+ */
+#[derive(Debug)]
+pub struct InputPath<'a>(&'a str);
```
smklein, line 149: I have no issue with this current use of typing, but I feel like I've usually seen the strong types used as an "output" of parsing functions, rather than input. IMO that seems more useful, since it propagates the notion that "this `&str` has been validated", whereas I think `InputPath` is currently truly indistinguishable from an arbitrary `&str`.

(I'm kinda eyeing `route_path_to_segments` and `input_path_to_segments` below - it does seem like both could take an arbitrary `&str` as input, but the results kinda have different implications)

smklein (approved):
One of those patches where I really appreciate having all the tests - thank you!

Looks good, just a handful of comments.

dropshot/src/api_description.rs
ahl: Well that's some obvious copypasta. Thanks for the careful look.

dropshot/src/api_description.rs
ahl: more copy pasta--sorry about that.

dropshot/src/router.rs
ahl: I spoke at length with @davepacheco about this. I'm not sure why we didn't arrive at your suggestion--perhaps we had been too micro-focused on various RFCs.

If a client were to implement file access they would need to consider symlinks as well as dot segments--that would obviously be well beyond what Dropshot itself could do.

All that said: I agree with your position and I don't think it's particularly valuable to allow an API to specify ".." as a path segment for an endpoint. @davepacheco if you disagree with this it seems easy to undo later.

dropshot/src/router.rs
ahl: I had something a bit like that, but I struggled to elegantly deal with the embedded `Result`. I really like this trick of returning `Ok()` to force the closure to have a `Result` type -- that's cool. Also cool that `Result` implements `FromIterator`. I think this work if I remove the surrounding `Ok(...)`. Thanks!!

dropshot/src/router.rs
ahl: I don't have a strong feeling about what would be most idiomatic here. I wanted to "quarantine" the input type to avoid accidental use of it. I agree that it's all a little meh. I'd love any input regarding how to improve it.

ahl:
@smklein thanks for the thorough review. I've pushed a new commit that I think addresses your feedback.

dropshot/src/router.rs
smklein: Cool, sounds good to me! I figured it would be better to start with a "safe-but-conservative" approach, since we can always walk it back later (but the other direction would be more difficult).

dropshot/src/router.rs
smklein: I wanna give the caveat - I think this patch is decidedly a net improvement, so this comment shouldn't be a blocker!

The reason I'm a little skeptical of the current usage is that I don't think the quarantining is actually happening very clearly (aside from requiring a bunch of `.into()` calls). Any function that operates on an `InputPath` *may as well* operate on a `&str`, since there is (seemingly?) no check or consideration to distinguish them.

IMO, the path that's the result of a "user input" actually *is* a str. If we want to distinguish that from a route-defining path, which has a higher burden of validation, perhaps *that* is the type we should abstract more cautiously.

As an example, `route_path_to_segments` returns a `Vec<&str>` - but it implicitly carries the implication that "this is actually a non-empty `&str` type, without `/` characters.

We could make that more explicit by returning a `Vec<Segment>`, where *Segment* is a `&str`, but it can be only constructed during validation.

Similarly, `input_path_to_segments` could return a `Vec<Segment>`, to indicate that it represents a percent-decoded, non "./.." str.

(This actually raises the question - should these functions actually be returning distinct `Segment` types? They appear to be performing slightly different validations)

smklein:
> @smklein thanks for the thorough review. I've pushed a new commit that I think addresses your feedback.

One last comment on the `InputPath` type, but LGTM!

dropshot/src/router.rs
smklein: FWIW: I think this pattern is generally the one Cliff calls "the typestate pattern".

- http://cliffle.com/blog/rust-typestate/
- https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

ahl:
I'm going to attempt to revisit the discussion of `InputPath` in #110

PR #119, opened 2021-06-28
file server example

### LINKED ISSUE #115: Serve static files
Web frameworks tend to make it easy to serve a local directory of static assets from an arbitrary route prefix. If we don't want to build something one-liner-ish into Dropshot directly, it would still be helpful to have an example somewhere of how to do it by hand.

This could be done dynamically with a wildcard route (#110) or use the directory listing of the asset directory to generate a static list of routes at compile time, which makes the wildcard route unnecessary. It's hard to tell, but I think Rocket is [doing the former](https://github.com/SergioBenitez/Rocket/blob/a26f93c57298612fd83e2c37edd0a6bbd7d18fdf/core/lib/src/fs/server.rs#L184) and Actix is doing the latter.

### Rocket ([example](https://github.com/SergioBenitez/Rocket/blob/f1ecb79a7ea536bacc7b7e92d33bf5406c0c635f/examples/static-files/src/main.rs))

```rs
use rocket::fs::{FileServer, relative};

#[rocket::launch]
        fn rocket() -> _ {
            rocket::build()
                .mount("/", FileServer::from(relative!("static")))
        }
```

### Actix ([docs](https://actix.rs/docs/static-files/))

```rs
use actix_files as fs;
use actix_web::{App, HttpServer};

#[actix_web::main]
async fn main() -> std::io::Result<()> {
    HttpServer::new(|| {
        App::new().service(fs::Files::new("/static", ".").show_files_listing())
    })
    .bind("127.0.0.1:8080")?
    .run()
    .await
}
```

### Express ([docs](https://expressjs.com/en/starter/static-files.html))

```js
app.use('/static', express.static('public'))
```

ahl:
Fixes #115 

This adds a (relatively) simple file server example with appropriate checking for dot-segments and symlinks. It displays a directory listing with links and picks a MIME type based on the filename.

### DIFF
--- a/dropshot/examples/file_server.rs
+++ b/dropshot/examples/file_server.rs
@@ -1,0 +1,197 @@
+// Copyright 2021 Oxide Computer Company
+
+use dropshot::ApiDescription;
+use dropshot::ConfigLogging;
+use dropshot::ConfigLoggingLevel;
+use dropshot::HttpError;
+use dropshot::HttpServerStarter;
+use dropshot::RequestContext;
+use dropshot::{endpoint, Path};
+use http::{Response, StatusCode};
+use hyper::Body;
+use schemars::JsonSchema;
+use serde::Deserialize;
+use std::path::PathBuf;
+use std::sync::Arc;
+
+/**
+ * Our context is simply the root of the directory we want to serve.
+ */
+struct FileServerContext {
+    base: PathBuf,
+}
+
+#[tokio::main]
+async fn main() -> Result<(), String> {
+    /*
+     * We must specify a configuration with a bind address.  We'll use 127.0.0.1
+     * since it's available and won't expose this server outside the host.  We
+     * request port 0, which allows the operating system to pick any available
+     * port.
+     */
+    let config_dropshot = Default::default();
+
+    /*
+     * For simplicity, we'll configure an "info"-level logger that writes to
+     * stderr assuming that it's a terminal.
+     */
+    let config_logging = ConfigLogging::StderrTerminal {
+        level: ConfigLoggingLevel::Info,
+    };
+    let log = config_logging
+        .to_logger("example-basic")
+        .map_err(|error| format!("failed to create logger: {}", error))?;
+
+    /*
+     * Build a description of the API -- in this case it's not much of an API!.
+     */
+    let mut api = ApiDescription::new();
+    api.register(static_content).unwrap();
+
+    /*
+     * Specify the directory we want to serve.
+     */
+    let context = FileServerContext {
+        base: PathBuf::from("."),
+    };
+
+    /*
+     * Set up the server.
+     */
+    let server = HttpServerStarter::new(&config_dropshot, api, context, &log)
+        .map_err(|error| format!("failed to create server: {}", error))?
+        .start();
+
+    /*
+     * Wait for the server to stop.  Note that there's not any code to shut down
+     * this server, so we should never get past this point.
+     */
+    server.await
+}
+
+/**
+ * Dropshot deserializes the input path into this Vec.
+ */
+#[derive(Deserialize, JsonSchema)]
+struct AllPath {
+    path: Vec<String>,
+}
+
+/**
+ * Serve files from the specified root path.
+ */
+#[endpoint {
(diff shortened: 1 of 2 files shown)


dropshot/examples/file_server.rs
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * Serve files from the specified root path.
+ */
+#[endpoint {
+    method = GET,
+
+    /*
+     * Match literally every path including the empty path.
+     */
+    path = "/{path:.*}",
```
kulakowski, line 89: I'm double checking: I think I saw a PR about these sorts of paths that included questions like can `..` be a path component? Or am I thinking of something else?

dropshot/examples/file_server.rs
```diff
@@ -0,0 +1,213 @@
...
+     * request port 0, which allows the operating system to pick any available
+     * port.
+     */
+    let config_dropshot = Default::default();
+
+    /*
+     * For simplicity, we'll configure an "info"-level logger that writes to
+     * stderr assuming that it's a terminal.
+     */
+    let config_logging = ConfigLogging::StderrTerminal {
```
kulakowski, line 38: Mostly because this is a convenient place for me to ask: suppose you wanted the logging level to be configurable via a command line argument. Have we picked a crate and style for that sort of thing?

dropshot/examples/file_server.rs
```diff
@@ -0,0 +1,213 @@
...
+            dir_link, dir_link
+        )
+        .as_str(),
+    );
+    body.push_str("<ul>\n");
+    while let Some(entry) = dir.next_entry().await? {
+        let name = entry.file_name();
+        let name = name.to_string_lossy();
+        /*
+         * Note that Dropshot handles paths with and without trailing slashes
```
kulakowski, line 184: (Meta: I haven't used github for code review in a serious way in my life, so let me know if I'm holding any of the comment/review features strangely please!)

Yeah, per my comment above, I think the details of path handling are maybe the most interesting thing to see in this PR, besides the fact of just being an example. Things like /./ as a middle component too, though I see dropshot handles that.

kulakowski (approved):
Minor comments, mostly for my edification :)

dropshot/examples/file_server.rs
ahl: You may well have and the answer is that we don't allow so-called "dot-segments" to be present. See https://github.com/oxidecomputer/dropshot/blob/main/dropshot/src/router.rs#L769

dropshot/examples/file_server.rs
ahl: yeesh. Good question and not one I have an answer to. Seems like something @davepacheco or @smklein might know off the top of their heads.

dropshot/examples/file_server.rs
ahl: Do you mean in terms of a new test for the `input_path_to_segments` function?

dropshot/examples/file_server.rs
bnaecker: @kulakowski I've been accepting a flag usually named `--log-level` or similar, which is a `slog::Level`, and using that type's `FromStr` implementation. This is basically how the [`nexus` binary works](https://github.com/oxidecomputer/omicron/blob/3a6531e46d24f8efa3eeb57fc866d621acd9eb9a/omicron-nexus/src/config.rs#L18) as well, although that's reading from a TOML file instead of the command-line. Hope that helps!

dropshot/examples/file_server.rs
kulakowski: Yup, that's the one!

dropshot/examples/file_server.rs
kulakowski: I think based on my understanding now, that's just well understood to be the behavior of dropshot, so I don't think you need to add any particular testing of it in a thing using dropshot.

kulakowski: I think based on my understanding now, that's just well understood to be the behavior of dropshot, so I don't think you need to add any particular testing of it in a thing using dropshot.

PR #122, opened 2021-07-15
Enable usage of syn from 'panic = abort' code

smklein:
Refer to the inline code comment for all the details.


dropshot/src/router.rs
```diff
@@ -180,10 +181,38 @@ impl<'a> From<&'a str> for InputPath<'a> {
     }
 }
 
+static INFORM_PROC_MACRO2_WE_ARE_NOT_A_PROC_MACRO: Once = Once::new();
+
 /*
  * Validate that the string is a valid Rust identifier.
  */
 fn valid_identifier(var: &str) -> bool {
+    // TODO: Remove this "Once" callback when the following issue
```
ahl, line 190: FWIW @davepacheco has tried (and failed) to hold the line on /* */ comments in dropshot so... do what you will with that .

ahl (approved):
Nice work chasing this down? Is there anything related to the initial bug beyond what you document in the code that's worth capturing in this PR?

dropshot/src/router.rs
smklein: Interesting, I always considered that comment style "valid, but non-idiomatic" - the Rust book only shows the "//" style (see: [here](https://doc.rust-lang.org/stable/book/ch03-04-comments.html))  and there was even an (admittedly unsuccessful) attempt to remove block comments: https://github.com/rust-lang/rfcs/issues/1371

(Plus I [heard a rustlang maintainer also had the spicy take that they're not idiomatic](https://github.com/rust-lang/book/issues/693#issuecomment-301200629))

Do you want me to switch before landing? I noticed we're pretty inconsistent across many of Oxide's repos (dunno if we should clarify which repos do / don't have this styling more explicitly).

smklein: [The rustdoc book](https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html) also exclusively uses the non-block comment style ("//" or "///")

smklein:
> Nice work chasing this down? Is there anything related to the initial bug beyond what you document in the code that's worth capturing in this PR?

I don't believe so - it's a bit hairy, and might change if we made other calls to `syn`/`quote` from this codebase, but we don't really have a better "initialization point" that wouldn't push this onto dropshot users.

I'm kinda hoping that this gets resolved in upstream rust, fixed in proc_macro2, we can remove the hack, and we'll be done with it.

dropshot/src/router.rs
ahl: I know.

PR #125, opened 2021-08-06
WIP implementation of TLS

smklein:


dropshot/Cargo.toml
```diff
@@ -11,24 +11,28 @@ keywords = ["rest", "openapi", "async"]
...
 futures = "0.3.1"
+futures-core = "0.3.16"
 hostname = "0.3.0"
 http = "0.2.0"
 indexmap = "1.0.0"
 openapiv3 = "0.5.0"
 paste = "1.0.0"
 percent-encoding = "2.1.0"
 proc-macro2 = "1.0.27"
+rustls = "0.19.1"
```
smklein, line 27: Dunno how much this is on your radar, but it looks like a 0.20 beta exists, and changes some parts of the TLS api.

I saw an example here: https://github.com/ctz/rustls/blob/main/rustls-mio/examples/tlsserver.rs ; 

The [rustls::ServerConnection object](https://docs.rs/rustls/0.20.0-beta2/rustls/struct.ServerConnection.html) looked new.

dropshot/src/server.rs
```diff
@@ -467,6 +712,38 @@ impl<C: ServerContext> Service<Request<Body>> for ServerRequestHandler<C> {
     }
 }
 
+fn error(err: String) -> std::io::Error {
+    std::io::Error::new(std::io::ErrorKind::Other, err)
+}
+
+// Load public certificate from file.
+fn load_certs(filename: &str) -> std::io::Result<Vec<rustls::Certificate>> {
```
smklein, line 720: Nitpick: I recommend using a `&Path` object for the filenames - there's an example here: https://github.com/tokio-rs/tls/blob/master/tokio-rustls/examples/server/src/main.rs#L33-L41

(Parsing it in the config as a `PathBuf`, as they did, also seems reasonable).

dropshot/src/server.rs
```diff
@@ -149,15 +201,208 @@ impl<C: ServerContext> HttpServerStarter<C> {
...
+        let pinned = Pin::new(&mut self.stream);
+        pinned.poll_shutdown(ctx)
+    }
+}
+
+struct HttpsAcceptor {
+    acceptor: Box<dyn Stream<Item = std::io::Result<TlsConn>> + Send + Unpin>,
+}
+
+impl hyper::server::accept::Accept for HttpsAcceptor {
```
smklein, line 272: I noticed that `Accept` is also implemented for [AddrIncoming](https://docs.rs/hyper/0.14.11/hyper/server/conn/struct.AddrIncoming.html) - is there something which prevents us from using that structure directly?

dropshot/src/server.rs
```diff
@@ -149,15 +201,208 @@ impl<C: ServerContext> HttpServerStarter<C> {
...
+                    loop {
+                        let (socket, addr) = tcp.accept().await?;
+                        let stream = acceptor.accept(socket).await?;
+                        let conn = TlsConn::new(stream, addr);
+                        yield Ok(conn)
+                    }
+                },
+                local_addr,
+            )
+        };
```
smklein, line 373: I'm not sure I grok the need for all the `TlsConn`  + `HttpAcceptor` structures.

If we have a `TcpListener` object (`tcp`) that is configured to use HTTPS, couldn't we just use:

```rust
let stream = hyper::server::conn::AddrIncoming::from_listener(tcp)?;
let server = Server::builder(stream);
```

dropshot/src/server.rs
smklein: (more on this thread down below)

smklein (commented):
Nice, cool to see this coming together!

dropshot/src/server.rs
```diff
@@ -149,15 +201,208 @@ impl<C: ServerContext> HttpServerStarter<C> {
...
+            let mut cfg =
+                rustls::ServerConfig::new(rustls::NoClientAuth::new());
+            cfg.set_single_cert(dont_use_this_cert, dont_use_this_key)?;
+            cfg.set_protocols(&[b"h2".to_vec(), b"http/1.1".to_vec()]);
+            let cfg = Arc::new(cfg);
+            let acceptor = TlsAcceptor::from(cfg);
+
+            let tcp = TcpListener::from_std(std::net::TcpListener::bind(
+                &config.bind_address,
+            )?)?;
```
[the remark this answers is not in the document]
kulakowski: It does so asynchronously though! This is one of the points of tension in making http(s) have the same interface.

dropshot/src/server.rs
kulakowski: And not `P: AsRef<Path>` (possibly just forwarding to an internal &Path version)? Just curious where y'all have fallen on that choice.

dropshot/src/server.rs
```diff
@@ -149,15 +201,208 @@ impl<C: ServerContext> HttpServerStarter<C> {
...
+            debug!(app_state.log, "registered endpoint";
+                "method" => &method,
+                "path" => &path
+            );
+        }
+
+        let (conn, local_addr) = {
+            let dont_use_this_cert = load_certs(
+                "/Users/gk/oxide/src/dropshot/dropshot/data/\
+                 dont_use_this_cert.pem",
```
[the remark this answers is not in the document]
kulakowski: Yeah, I wanted to make sure that just pushing that out ~directly to clients was a fine start. Thanks.

dropshot/src/server.rs
```diff
@@ -75,21 +81,67 @@ pub struct ServerConfig {
  * A thin wrapper around a Hyper Server object that exposes some interfaces that
  * we find useful.
  */
-pub struct HttpServerStarter<C: ServerContext> {
+
```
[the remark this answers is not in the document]
kulakowski: Line break done.

dropshot/src/server.rs
smklein: Oh, that works great too! That basically just adds flexibility, but yeah, making it explicit that this is a path instead of an arbitrary string is a good sanitization choice.

(technically the "as ref" version is the most flexible, but it matters less for a non-pub function)

dropshot/src/server.rs
smklein: Nvm - I see I got my configs mixed up

dropshot/data/dont_use_this_cert.pem
```diff
@@ -0,0 +1,113 @@
+-----BEGIN CERTIFICATE-----
```
davepacheco, line 1: I believe that lines outside the BEGIN and END lines will be ignored in PEM files.  Can we add some comments explaining where this file came from and how it's used?

Even better, if this is used for testing, would it make sense to generate it in the test suite rather than checking it in?  That way someone wouldn't have to figure out how to regenerate it when we need to change it -- the process will be automated and run all the time.

dropshot/data/dont_use_this_key.rsa
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
davepacheco, line 1: Same comment here -- can we generate this as needed?  And if not, can we put a comment here explaining how it's used and how to generate it?

dropshot/examples/https.rs
```diff
@@ -0,0 +1,151 @@
+// Copyright 2020 Oxide Computer Company
+/*!
+ * Example use of Dropshot.
```
davepacheco, line 3: I think this is just copied/pasted.  Can we give a better description of this example?

davepacheco (commented):
Thanks for taking this on!  I only took a quick skim so far.  Something I think might help a lot (even for the review phase) is adding the user-facing documentation for this.  I don't have a sense yet for how users would enable this, either for dev or production.  I'm assuming it's not just setting "tls = true" in the config, right?

dropshot/data/dont_use_this_cert.pem
teisenbe: Removed these and switched to generating self-signed certificates in the tests

dropshot/Cargo.toml
teisenbe: Upgraded

teisenbe (commented):
I've taken this work up over in #224.

PR #133, opened 2021-09-27
more stringent checking of ItemFn signature fields

### LINKED ISSUE #132: Support for generic endpoints
Generic functions are not currently supported as Dropshot endpoints. This capability would be useful for writing generic Dropshot interfaces over traits. A minimum reproducible example is provided below.

In the mean time, the error message that comes up when this is attempted should be something to the effect of "generic endpoints are not currently supported". The error that comes up now (below) does not make it clear what's going on.

Example:

```rust
use std::sync::Arc;
use dropshot::{        
    endpoint,
    ConfigDropshot,
    ConfigLogging,
    ConfigLoggingLevel,
    ApiDescription,
    HttpServerStarter,
    RequestContext,
    HttpResponseOk,
    HttpError,
};

trait Greeter {
    fn say_hello(&self) -> String;
}

struct Context<G: Greeter + std::marker::Send + std::marker::Sync + 'static> {
    greeter: G,
}

#[endpoint { method = GET, path = "/greet" }]                                                      
async fn greet<G: Greeter + std::marker::Send + std::marker::Sync + 'static>(
    ctx: Arc<RequestContext<Context<G>>>
) -> Result<HttpResponseOk<String>, HttpError> {

    let api_context = ctx.context();
    Ok(HttpResponseOk(api_context.greeter.say_hello()))

}                                                                                                  

#[tokio::main]
async fn main() -> Result<(), String> {

    let log =
        ConfigLogging::StderrTerminal { level: ConfigLoggingLevel::Info }
        .to_logger("minimal-example")
        .map_err(|e| e.to_string())?;

    // I realize this bit is going to be tricky, as the concrete type to make this an
    // instantiated function does not come until the HttpServerStarter::new call
    // when we pass in the Context with the specific Greeter implementation
    api.register(greet).unwrap();

    let api_context = Context{ greeter: Mandarin{} };

    let server = HttpServerStarter::new(
        &ConfigDropshot { ..Default::default() },
        api,
        api_context,
        &log,
    ).expect("server new").start();

    server.await

}

struct Mandarin {}
impl Greeter for Mandarin {
    fn say_hello(&self) -> String { "你好".to_string() }
}

struct English{}
impl Greeter for English {
    fn say_hello(&self) -> String { "hello".to_string() }
}
```

Which produces the compiler error
```
error[E0412]: cannot find type `G` in this scope
  --> src/main.rs:24:37
   |
24 |     ctx: Arc<RequestContext<Context<G>>>
   |                                     ^   - help: you might be missing a type parameter: `<G>`
   |                                     |
   |                                     not found in this scope

error[E0412]: cannot find type `G` in this scope
  --> src/main.rs:24:37
   |
24 |     ctx: Arc<RequestContext<Context<G>>>
   |                                     ^ not found in this scope

error[E0277]: the trait bound `ApiEndpoint<_>: From<greet>` is not satisfied
  --> src/main.rs:41:9
   |
41 |     api.register(greet).unwrap();
   |         ^^^^^^^^ the trait `From<greet>` is not implemented for `ApiEndpoint<_>`
   |
   = note: required because of the requirements on the impl of `Into<ApiEndpoint<_>>` for `greet`

Some errors have detailed explanations: E0277, E0412.
For more information about an error, try `rustc --explain E0277`.
```

ahl:
closes #132 

This adds a test for generics and also examines other fields we weren't checking for. I explicitly did not add tests for those fields because I don't think we really care about the stability of that output, but I'd be happy to if folks think it would be beneficial. The `fail` tests already take kind of a long time so I'm loth to add more.

### DIFF
--- a/dropshot_endpoint/src/lib.rs
+++ b/dropshot_endpoint/src/lib.rs
@@ -114,11 +114,43 @@ fn do_endpoint(
 
     let ast: ItemFn = syn::parse2(item.clone())?;
 
+    if ast.sig.constness.is_some() {
+        return Err(Error::new_spanned(
+            ast.sig.constness,
+            "endpoint handlers may not be const functions",
+        ));
+    }
+
     if ast.sig.asyncness.is_none() {
         return Err(Error::new_spanned(
             ast.sig.fn_token,
             "endpoint handler functions must be async",
         ));
+    }
+
+    if ast.sig.unsafety.is_some() {
+        return Err(Error::new_spanned(
+            ast.sig.unsafety,
+            "endpoint handlers may not be unsafe",
+        ));
+    }
+
+    if ast.sig.abi.is_some() {
+        return Err(Error::new_spanned(
+            ast.sig.abi,
+            "endpoint handler may not use an alternate ABI",
+        ));
+    }
+
+    if !ast.sig.generics.params.is_empty() {
+        return Err(Error::new_spanned(
+            ast.sig.generics,
+            "generics are not permitted for endpoint handlers",
+        ));
+    }
+
+    if ast.sig.variadic.is_some() {
+        return Err(Error::new_spanned(ast.sig.variadic, "no language C here"));
     }
 
     let name = &ast.sig.ident;
--- a/dropshot/tests/fail/bad_endpoint13.rs
+++ b/dropshot/tests/fail/bad_endpoint13.rs
@@ -1,0 +1,26 @@
+// Copyright 2021 Oxide Computer Company
+
+#![allow(unused_imports)]
+
+use dropshot::endpoint;
+use dropshot::HttpError;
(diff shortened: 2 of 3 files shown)


dropshot_endpoint/src/lib.rs
```diff
@@ -114,13 +114,45 @@ fn do_endpoint(
...
+
+    if !ast.sig.generics.params.is_empty() {
+        return Err(Error::new_spanned(
+            ast.sig.generics,
+            "generics are not permitted for endpoint handlers",
+        ));
+    }
+
+    if ast.sig.variadic.is_some() {
+        return Err(Error::new_spanned(ast.sig.variadic, "no language C here"));
```
davepacheco, line 153: 😆

davepacheco (approved):

PR #135, opened 2021-10-12
do our best to emit macro output even in the face of errors

### LINKED ISSUE #134: rust-analyzer newly confused by endpoint macro
Recently, the Rust Analyzer folks enabled (deeper?) processing of proc attribute macros, in: rust-analyzer/rust-analyzer#10366.

It seems that as of that commit, code completion inside endpoint functions (those wrapped by `#[endpoint]`) is a bit flaky.  It will sometimes work once or twice, but then can eventually get confused and stops working altogether.  This state seems to be sticky until you restart RA.

While I am not yet convinced RA is doing entirely the right thing here, it has been suggested in rust-analyzer/rust-analyzer#10498 that:

> The problem is likely that the proc macro emits an error when the function is syntactically invalid. (What error message does rustc emit when faced with this incomplete code?) In that case, probably your only options are to either report this to the proc macro authors and wait for them to fix it, or to disable attribute macro expansion again.

Indeed, `endpoint` _does_ emit an error when the function body is syntactically invalid -- but I'm not really sure what we should do instead?  Should we be doing our best to construct what will be an invalid token stream and passing that back without error, unless the stream is so invalid that it is unrecognisable?

ahl:
closes #134 

This restructures the `#[endpoint .. ]` macro so that it only fails in the case of some really busted input. Our goal is to do our level best to produce reasonable output in order to avoid confusing rust analyzer.

Note that this changed some of the error messages; I think these are largely improvements, but @davepacheco I'd love your opinion.

@jclulow if you're able to see if this resolves your issue that would be tremendously helpful. I played around on the latest RA, and while I was able to conjure up what you described, I'm not 100% convinced that this fixes it (i.e. and that perhaps I'm now just getting lucky).

### DIFF
--- a/dropshot/tests/fail/bad_endpoint1.stderr
+++ b/dropshot/tests/fail/bad_endpoint1.stderr
@@ -1,6 +1,5 @@
-error: Endpoint requires arguments
-Endpoint handlers must have the following signature:
-    async fn bad_endpoint(
+error: Endpoint handlers must have the following signature:
+    async fn(
         rqctx: std::sync::Arc<dropshot::RequestContext<MyContext>>,
         [query_params: Query<Q>,]
         [path_params: Path<P>,]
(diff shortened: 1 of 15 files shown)


ahl:
> Thanks, Adam! I will test this tomorrow (Tuesday).

No rush. I'm not 100% about this approach; it may well be the wrong direction.

PR #136, opened 2021-10-13
Add testing helper for DELETE requests

teisenbe:
This introduces test_util.object_delete(), to go along with the existing test_util helpers. It's helpful for testing CRUD APIs.

### DIFF
--- a/dropshot/src/test_util.rs
+++ b/dropshot/src/test_util.rs
@@ -614,6 +614,25 @@ pub async fn objects_post<S: Serialize + Debug, T: DeserializeOwned>(
 }
 
 /**
+ * Issues an HTTP DELETE to the specified object URL to delete an object.
+ */
+pub async fn object_delete<T: DeserializeOwned>(
+    client: &ClientTestContext,
+    object_url: &str,
+) -> T {
+    let mut response = client
+        .make_request_with_body(
+            Method::DELETE,
+            &object_url,
+            "".into(),
+            StatusCode::OK,
+        )
+        .await
+        .unwrap();
+    read_json::<T>(&mut response).await
+}
+
+/**
  * Iterate a paginated collection.
  */
 pub async fn iter_collection<T: Clone + DeserializeOwned>(
--- a/dropshot/tests/test_demo.rs
+++ b/dropshot/tests/test_demo.rs
@@ -18,8 +18,10 @@
 use dropshot::endpoint;
 use dropshot::test_util::read_json;
 use dropshot::test_util::read_string;
+use dropshot::test_util::object_delete;
 use dropshot::ApiDescription;
 use dropshot::HttpError;
+use dropshot::HttpResponseDeleted;
 use dropshot::HttpResponseOk;
 use dropshot::Path;
 use dropshot::Query;
@@ -52,6 +54,7 @@ fn demo_api() -> ApiDescription<usize> {
     api.register(demo_handler_path_param_uuid).unwrap();
     api.register(demo_handler_path_param_u32).unwrap();
     api.register(demo_handler_untyped_body).unwrap();
+    api.register(demo_handler_delete).unwrap();
 
     /*
      * We don't need to exhaustively test these cases, as they're tested by unit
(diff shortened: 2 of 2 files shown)


teisenbe:
Hmm, those tests seem to be failing for me on main. Not sure what's going on there

smklein:
> Hmm, those tests seem to be failing for me on main. Not sure what's going on there

What's the error message you're seeing? I just did `git checkout main && cargo test`, not seeing anything

dropshot/src/test_util.rs
```diff
@@ -613,6 +613,25 @@ pub async fn objects_post<S: Serialize + Debug, T: DeserializeOwned>(
...
+ */
+pub async fn object_delete<T: DeserializeOwned>(
+    client: &ClientTestContext,
+    object_url: &str,
+) -> T {
+    let mut response = client
+        .make_request_with_body(
+            Method::DELETE,
+            &object_url,
+            "".into(),
```
smklein, line 627: Since we're passing `"".into()` as the body, should we just invoke `make_request_no_body`?

dropshot/src/test_util.rs
```diff
@@ -613,6 +613,25 @@ pub async fn objects_post<S: Serialize + Debug, T: DeserializeOwned>(
...
+            StatusCode::OK,
```
smklein, line 628: Running this branch locally, I'm seeing a test failure here because the response is `StatusCode::NO_CONTENT`, not `StatusCode::OK`.

Here's where `HttpResponseDeleted` is defined: 
https://github.com/oxidecomputer/dropshot/blob/f99a003271f41f933172e33c3cc88a2f67bbdc7c/dropshot/src/handler.rs#L1234-L1251

This makes sense to me - it looks like our `HttpResponseDeleted` sends a `NO_CONTENT` response. Should that be the response we expect here?

dropshot/src/test_util.rs
```diff
@@ -613,6 +613,25 @@ pub async fn objects_post<S: Serialize + Debug, T: DeserializeOwned>(
...
+        )
+        .await
+        .unwrap();
+    read_json::<T>(&mut response).await
```
smklein, line 632: ... and as a follow-up, if we are sending a `NO_CONTENT` message back, it probably doesn't make sense to try to read anything from the response - it'll likely be empty.

dropshot/src/test_util.rs
davepacheco: I think "204 No Content" is more customary for DELETE operations, especially if you're not returning anything useful in the body.  [Here's a source.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE).

teisenbe:
Hmm, the windows build failed `logging::test::test_config_file` with this output:
```
---- logging::test::test_config_file stdout ----
config "file": Ok(File { level: Warn, path: "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\dropshot-af77ac2c25c9d09a.exe.2092.file_dir\\log.out", if_exists: Fail })
note: configured to log to "C:\Users\RUNNER~1\AppData\Local\Temp\dropshot-af77ac2c25c9d09a.exe.2092.file_dir\log.out"
config "file": Ok(File { level: Warn, path: "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\dropshot-af77ac2c25c9d09a.exe.2092.file_dir\\log.out", if_exists: Append })
note: configured to log to "C:\Users\RUNNER~1\AppData\Local\Temp\dropshot-af77ac2c25c9d09a.exe.2092.file_dir\log.out"
config "file": Ok(File { level: Trace, path: "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\dropshot-af77ac2c25c9d09a.exe.2092.file_dir\\log.out", if_exists: Truncate })
note: configured to log to "C:\Users\RUNNER~1\AppData\Local\Temp\dropshot-af77ac2c25c9d09a.exe.2092.file_dir\log.out"
thread 'logging::test::test_config_file' panicked at 'assertion failed: should_be_before.timestamp() <= time_after.timestamp()', dropshot\src\test_util.rs:786:13
```

This seems unrelated to my change; I can open a ticket for this

PR #140, opened 2021-10-13
Update dependencies and expected output

smklein:
Fixes https://github.com/oxidecomputer/dropshot/issues/138

This is the result of checking out main, and running:

```
$ cargo update
$ TRYBUILD=overwrite cargo test
```

I believe this implies the version of our dependencies may cause different compiler output.

### DIFF
--- a/dropshot/tests/fail/bad_endpoint10.stderr
+++ b/dropshot/tests/fail/bad_endpoint10.stderr
@@ -1,5 +1,5 @@
 error[E0271]: type mismatch resolving `<String as TypeEq>::This == HttpError`
-  --> $DIR/bad_endpoint10.rs:16:6
+  --> tests/fail/bad_endpoint10.rs:16:6
    |
 16 | ) -> Result<HttpResponseOk<()>, String> {
    |      ^^^^^^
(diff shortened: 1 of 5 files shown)


smklein:
> What dependency led to this change or was it a difference in the version of the toolchain?

Ultimately, it was a dependency, *not* the toolchain as I originally expected. I didn't bisect the deps, I just did `cargo update`, so I'm actually unsure which one it was (... I can dig, but frankly, do we care?).

ahl:
> (... I can dig, but frankly, do we care?).

No need for you to dig. I (though perhaps not we) care because it's surprising that a dependency could cause this change. The only one that makes sense is trybuild and indeed that seems to be what happened:

https://github.com/dtolnay/trybuild/pull/132

smklein:
> > (... I can dig, but frankly, do we care?).
> 
> No need for you to dig. I (though perhaps not we) care because it's surprising that a dependency could cause this change. The only one that makes sense is trybuild and indeed that seems to be what happened:
> 
> [dtolnay/trybuild#132](https://github.com/dtolnay/trybuild/pull/132)

Ah, good find. I didn't mean to sound dismissive - I only figured the solution would remain the same regardless of the culprit!

Thanks for the look.

Issue #46, opened 2020-08-15
cannot presently correlate request source IP and other request logs


jclulow:
When using the `Logger` provided on the `RequestContext`, some properties are already included by default; most usefully for correlation of multiple log entries is the request ID.  Unfortunately the only place in the logs that the source IP and port of the connection through which the request was sent appears not to contain this request ID:

```
Aug 15 04:03:11.580 INFO accepted connection, remote_addr: 10.1.1.90:47312
Aug 15 04:03:11.580 WARN lunar waneshaft not ambifacient, uri: /prefabulate/amulite, method: POST, req_id: acd330b9-9e9c-494d-ac7a-610ba0fc8933
Aug 15 04:03:11.581 INFO request completed, error_message_external: invalid, error_message_internal: invalid, response_code: 400, uri: /prefabulate/amulite, method: POST, req_id: acd330b9-9e9c-494d-ac7a-610ba0fc8933
```

One imagines the request ID is not actually available yet at the time we receive the incoming connection, as it will presumably be taken from a request header (if provided) to allow tracing.  It seems like we could, then, do at least one of:

* another ID that tracks the _connection_; e.g., `$PID/$CONNID` where `CONNID` would start at 0 and increment for each connection served by this process.  This could also be a UUID separate from the one the client passes in, which may subsequently be useful for OpenTracing-style spans and sub-spans anyway.
* the remote_addr field appear in each log line as well as the request ID
* defer emitting the `accepted connection` message until we have enough context to know what the request ID is (seems not great)

davepacheco:
My suggestion for now would be to include the remote address and port in subsequent log entries.  I've no objection to a connection id either but this seems useful enough that avoiding the extra lookup when debugging seems worth it.

PR #148, opened 2021-10-14
Adds remote-address connection info to server

bnaecker:
- Reworks how a `hyper::Server` is created, to get access to the local
  address before starting the server itself. This allows it to be put
  into the key-value pairs for the server's logger, so they're printed
  on each subsequent message.
- Updates the internal connection/request handler types to get the
  remote address and store it in the logger passed to the request
  handling functions. The remote address is also printed on each
  subsequent log message.

### DIFF
--- a/dropshot/src/server.rs
+++ b/dropshot/src/server.rs
@@ -56,6 +56,8 @@ pub struct DropshotState<C: ServerContext> {
     pub router: HttpRouter<C>,
     /** server-wide log handle */
     pub log: Logger,
+    /** bound local address for the server. */
+    pub local_addr: SocketAddr,
 }
 
 /**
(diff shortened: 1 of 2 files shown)


bnaecker:
This should address #46 when merged.

dropshot/src/server.rs
```diff
@@ -123,6 +125,9 @@ impl<C: ServerContext> HttpServerStarter<C> {
         private: C,
         log: &Logger,
     ) -> Result<HttpServerStarter<C>, hyper::Error> {
+        let incoming = AddrIncoming::bind(&config.bind_address)?;
```
davepacheco, line 128: Nice find.

PR #160, opened 2021-10-19
Adds USDT probes to Dropshot

bnaecker:
- Adds the `RequestInfo` and `ResponseInfo` structs to package
  information available to DTrace
- Adds USDT probes `request_start` and `request_finish`
- Exposes the probes in the `http_request_handle_wrap` function
- Adds feature flag for optionally enabling probes
- Adds notes to README about the probes, how to enable and see via
  DTrace

### DIFF
--- a/dropshot/src/lib.rs
+++ b/dropshot/src/lib.rs
@@ -497,6 +497,39 @@
  * structure that includes a `PaginationParams` using `#[serde(flatten)]`, and
  * this ought to work, but it currently doesn't due to serde_urlencoded#33,
  * which is really serde#1183.
+ *
+ * ### DTrace probes
+ *
+ * Dropshot optionally exposes two DTrace probes, `request_start` and
+ * `request_finish`. These provide detailed information about each request,
+ * such as their ID, the local and remote IPs, and the response information.
+ * See the [`RequestInfo`] and [`ResponseInfo`] types for a complete listing
+ * of what's available.
+ *
+ * These probes are implemented via the [`usdt`] crate, and require a nightly
+ * toolchain. As such, they're behind the feature flag `"usdt-probes"`. You
+ * can build Dropshot with these probes via `cargo +nightly build --features usdt-probes`.
+ *
+ * Once in place, the probes can be seen via DTrace. For example, running:
+ *
+ * ```bash
+ * $ cargo +nightly run --example basic --features usdt-probes
+ * ```
+ *
+ * And making several requests to it with `curl`, we can see the DTrace
+ * probes with an invocation like:
+ *
+ * ```bash
+ * # dtrace -n 'dropshot*:::request_* { printf("%s", copyinstr(arg0)); }'
+dtrace: description 'dropshot*:::request_* ' matched 2 probes
+CPU     ID                    FUNCTION:NAME
+ 18  81674 _ZN8dropshot6server24http_request_handle_wrap28_$u7b$$u7b$closure$u7d$$u7d$17h00739672f19564edE:request_start {"ok": {"id":"c15659af-adb2-4785-b053-768ded70a7d8","local_addr":"127.0.0.1:37458","remote_addr":"127.0.0.1:42850","method":"GET","path":"/counter","query":null}}
+ 18  81673 _ZN8dropshot6server24http_request_handle_wrap28_$u7b$$u7b$closure$u7d$$u7d$17h00739672f19564edE:request_finish {"ok": {"id":"c15659af-adb2-4785-b053-768ded70a7d8","local_addr":"127.0.0.1:37458","remote_addr":"127.0.0.1:42850","status_code":200,"message":""}}
+  3  81674 _ZN8dropshot6server24http_request_handle_wrap28_$u7b$$u7b$closure$u7d$$u7d$17h00739672f19564edE:request_start {"ok": {"id":"678ddbe6-8433-4ddf-a8b6-732fea22f368","local_addr":"127.0.0.1:37458","remote_addr":"127.0.0.1:45707","method":"PUT","path":"/counter","query":null}}
+  3  81673 _ZN8dropshot6server24http_request_handle_wrap28_$u7b$$u7b$closure$u7d$$u7d$17h00739672f19564edE:request_finish {"ok": {"id":"678ddbe6-8433-4ddf-a8b6-732fea22f368","local_addr":"127.0.0.1:37458","remote_addr":"127.0.0.1:45707","status_code":204,"message":""}}
+  0  81674 _ZN8dropshot6server24http_request_handle_wrap28_$u7b$$u7b$closure$u7d$$u7d$17h00739672f19564edE:request_start {"ok": {"id":"0fea32fb-12de-463e-a0a9-17c6542dd51a","local_addr":"127.0.0.1:37458","remote_addr":"127.0.0.1:58478","method":"POST","path":"/counter","query":null}}
+  0  81673 _ZN8dropshot6server24http_request_handle_wrap28_$u7b$$u7b$closure$u7d$$u7d$17h00739672f19564edE:request_finish {"ok": {"id":"0fea32fb-12de-463e-a0a9-17c6542dd51a","local_addr":"127.0.0.1:37458","remote_addr":"127.0.0.1:58478","status_code":405,"message":"Method Not Allowed"}}
+ * ```
  */
 
 /*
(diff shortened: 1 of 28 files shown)


dropshot/src/server.rs
```diff
@@ -154,6 +154,16 @@ impl<C: ServerContext> HttpServerStarter<C> {
         let server = builder.serve(make_service);
         info!(app_state.log, "listening");
 
+        if let Err(e) = crate::register_probes() {
+            warn!(
```
davepacheco, line 158: I think this should be an error.  (I almost want to say it should be fatal.  I've been burned by things silently failing to enable DTrace support.  And if it was built for it, there's no reason this should fail at runtime, right?  But I imagine this sounds extreme.)

dropshot/src/server.rs
```diff
@@ -154,6 +154,16 @@ impl<C: ServerContext> HttpServerStarter<C> {
         let server = builder.serve(make_service);
         info!(app_state.log, "listening");
 
+        if let Err(e) = crate::register_probes() {
```
davepacheco, line 157: What happens if this is called multiple times in the same process?  What if it's concurrent?

dropshot/src/server.rs
```diff
@@ -290,6 +300,22 @@ async fn http_request_handle_wrap<C: ServerContext>(
         "uri" => format!("{}", request.uri()),
     ));
     trace!(request_log, "incoming request");
+    dropshot_request_start!(|| {
+        let uri = request.uri();
+        crate::RequestInfo {
+            id: request_id.parse().unwrap(),
```
davepacheco, line 306: Maybe we should have `generate_request_id()` return a Uuid rather than a String so that we don't have to parse and unwrap here.  (It's a little scary to unwrap inside an IS_ENABLED block.)

dropshot/src/server.rs
bnaecker: Ahh, wait you mean return an error from the start method. Yes, that seems totally fine to me.

davepacheco: I did actually mean an "error" level log message.  I could also see letting the caller decide.  I'm curious what @ahl thinks.

ahl: To @davepacheco 's earlier point: it would suck to have this fail and not notice. While it's true, @bnaecker, that there's an `ioctl(2)`, I wouldn't describe this as I/O per se.

I think this raises the question: "how could this fail transiently?" I think only if the system was low on memory and it could not allocate the necessary kernel data structures needed to store the provider state.  I can't think of another scenario where registration would fail transiently.

In that situation what would we want? Perhaps that answer would vary from service to service. If the system is already in a rocky state, perhaps we'd rather have, say, a service up that can tell us that we're low on memory. Unfortunately I could see either option (panic!() or ignore) biting us in various situations.

dropshot/src/server.rs
ahl: We deduplicate providers based on the **user-land address of the DOF structure**:

https://github.com/illumos/illumos-gate/blob/master/usr/src/uts/common/dtrace/dtrace.c#L15139-L15146

This made sense when we thought the DOF would be in some static, ro section, and that registration was just ioct-ing it downstairs. Because we're generating the DOF this isn't the case.

So... yeah. Unless I'm mistaken, this kind of sucks.

Perhaps if the `set_dtrace_probes` section were writable we could `cas` the first record to write a sentinel value and make sure no one sees those same records.

dropshot/src/server.rs
bnaecker: It sounds like there isn't an option that doesn't trade some things off. We can emit an error message to the log, no problem. Returning an `Err` seems OK, but there's no way for the caller to indicate that they don't care about such a situation and want to start the server regardless. (Not that we'd want to do that, but someone else might.) If we panic, that's also not great for obvious reasons.

I think the right thing here is to not have the server call `register_probes`, but to let the caller do it via that public function. It's a bit risky, since they have to read the documentation saying that they need to call this, but I don't see a better way. It seems overall worse to, for example, include an argument in the `start()` method saying how to handle this situation -- just let the caller do it, right?

dropshot/src/server.rs
bnaecker: Let me check my understanding, @ahl. Since we're reading some data from our special probes section into a buffer in memory, and the address of _that_ buffer is the value of the `dofhp_addr` field that DTrace uses to de-duplicate probes, it's possible that the same probe can be re-registered. Is that right?

That does sound pretty problematic. Some logic in USDT to ensure we only register probes once does sound important. Either a flag in each section, or a static variable in the library with the names of the registered probes sound like OK options. I'd have to think more about tradeoffs here.

dropshot/src/server.rs
bnaecker: @davepacheco Coming back to this. There's a note in the docstring for `generate_request_id` discussing options for including other information in the ID, such as the service name. That'd only make sense if this is a String, not a Uuid. Is that still a goal? If not, I can update the function to return a Uuid and avoid any further parsing/unwrapping.

bnaecker: A simpler option here is to change the `RequestInfo::id` field to also be a string. I'll go with that for now.

dropshot/src/server.rs
bnaecker: Just a note on the deduplication. This will be fixed in the next version of `usdt`. I'll wait for that to be merged, then update the PR accordingly.

dropshot/src/server.rs
bnaecker: Done in b52dc29

dropshot/src/server.rs
bnaecker: Done in b52dc29

dropshot/src/server.rs
bnaecker: I've updated this in b52dc29, which requires that the caller register probes and gives them the option of how to handle failure. The examples now all do `dropshot::register_probes().expect("Failed to register DTrace probes")`. Let me know if y'all have further thoughts on this.

dropshot/src/lib.rs
```diff
@@ -510,6 +510,11 @@
  * toolchain. As such, they're behind the feature flag `"usdt-probes"`. You
  * can build Dropshot with these probes via `cargo +nightly build --features usdt-probes`.
  *
+ * > *Important:* The probes _must_ be registered with the DTrace kernel
+ * mdoule for them to be visible via `dtrace(1M)`. This should be done with
```
davepacheco, line 514:
```suggestion
 * module for them to be visible via `dtrace(1M)`. This should be done with
```

dropshot/src/lib.rs
```diff
@@ -565,16 +568,16 @@ pub(crate) struct ResponseInfo {
...
+///
+/// This method _must_ be called for any of the DTrace probes in Dropshot
+/// to be visible from the `dtrace(1M)` command-line tool.
 pub fn register_probes() -> Result<(), usdt::Error> {
-    if cfg!(feature = "usdt-probes") {
-        usdt::register_probes()
-    } else {
-        Ok(())
-    }
+    usdt::register_probes()
```
davepacheco, line 580: Is this no longer gated behind a feature?  I'm probably misremembering but I thought this was to keep things working on "stable".

davepacheco:
Looking good!  Would you also mind adding a note to the changelog?

dropshot/src/lib.rs
bnaecker: There's still a feature flag for Dropshot. This function is always available in `usdt`, regardless of whether a real or no-op implementation is selected, to keep the public API the same in any case.

dropshot/src/lib.rs
bnaecker: Thanks!

bnaecker:
@davepacheco Yes, I'll add notes to the changelog, and thanks for the review!

I'm actually planning to update this a bit, in a way we've discussed previously. I think it's a bit better to perform registration in the server construction, so that no one needs to remember to do this or change existing code to get probes, and store the result of that in a field of the server. Instead of a separate call, one can check the field to assert probes have been enabled (or behave however one chooses).

dropshot/src/server.rs
bnaecker: Change of plan -- I've now stored the result of registration in a field of the `HttpServer`. Probes are always registered if possible (no need to call `register_probes()` manually), and callers can still decide how to handle failures.

PR #162, opened 2021-10-21
use explicit carets in Cargo files to work around Dependabot

davepacheco:
[When Cargo looks at dependency semver strings, it treats strings with no operator as having an implicit leading caret.](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)  However, Dependabot seems not to treat it this way, leading to the problems I mentioned in #156.

This change adds an explicit caret to all of our implicit caret dependencies.  I also went back and undid the recent Cargo.toml changes that Dependabot made.  This change should be safe because we tested these versions a week ago (and this PR will test them again).  The benefit is that it eases the constraint on consumers, since we accept a wider range of versions.

Going forward, the hope is that:

- For Dropshot users, it behaves like other libraries: Cargo.toml is only updated when we really depend on a newer version of something or when a semver-incompatible dependency update comes out.  This allows consumers to use a range of versions.
- By using Dependabot (and working around this issue), we also get the benefit of finding out pretty quickly when dependency changes have broken Dropshot
- By checking in Cargo.lock, when we find out about such breakage, it's in a Dependabot PR, not because the "main" branch build has broken.

I'm not totally sure whether this will work.  I'm going by a few data points from #156 and dependabot/dependabot-core#4009.  But this is pretty low risk and easy to undo.

### DIFF
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -11,28 +11,32 @@ keywords = ["rest", "openapi", "async"]
 categories = ["network-programming", "web-programming::http-server"]
 
 [dependencies]
-async-trait = "0.1.51"
-base64 = "0.13.0"
-bytes = "1"
-futures = "0.3.1"
-hostname = "0.3.0"
-http = "0.2.0"
-indexmap = "1.0.0"
-openapiv3 = "0.5.0"
-paste = "1.0.0"
-percent-encoding = "2.1.0"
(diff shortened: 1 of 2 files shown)


davepacheco:
I'm less optimistic this will work after [this comment](https://github.com/dependabot/dependabot-core/issues/4009#issuecomment-948796583).  It may be time to look at Renovate instead.

PR #182, opened 2021-11-08
Update usdt to v0.1.19 and update probe calls

pfmooney:


pfmooney:
It should be noted that the dropshot tests don't complete successfully when run under the nightly cargo/compiler, so testing the probes via that method (since `asm` and thus nightly is required) is a no-go.

dropshot/src/lib.rs
```diff
@@ -568,7 +568,7 @@ pub(crate) struct ResponseInfo {
 }
 
 #[usdt::provider]
-mod dropshot {
+mod probes {
```
davepacheco, line 571: Did this change the provider name?

davepacheco:
> It should be noted that the dropshot tests don't complete successfully when run under the nightly cargo/compiler, so testing the probes via that method (since `asm` and thus nightly is required) is a no-go.

Is that a recent change, do you know?

davepacheco:
@bnaecker can you take a look at this one too?

dropshot/src/lib.rs
bnaecker: That would change the provider name, yes, to `probes`. If we want to specify the path at which the probe macros are available, the `probe_path` and `probe_name` arguments to the `usdt::provider` macro are the way to do that. See [here](https://docs.rs/usdt/0.1.19/usdt/#configurable-names) for a description of how to do that. cc @pfmooney

bnaecker: To be explicit, if you want the probes available at a module called `crate::probes`, you'd do:

```rust
#[usdt::provider(probe_path = "probes")]
mod dropshot {
    ...
}
```

dropshot/src/server.rs
```diff
@@ -327,7 +328,7 @@ async fn http_request_handle_wrap<C: ServerContext>(
         "uri" => format!("{}", request.uri()),
     ));
     trace!(request_log, "incoming request");
-    dropshot_request_start!(|| {
+    probes::request_start!(|| {
```
bnaecker, line 331: This should work fine, once the provider module has been renamed as I mentioned above.

dropshot/src/lib.rs
pfmooney: Sorry, I misunderstood that documentation, since both dtrace probes and rust modules are `:`-delimited.  I know this would constitute another breaking change for usdt, but what about overriding the provider name via an argument to `#[usdt::provider]`, rather than one to override how the macros are exposed?  I know the module and the functions are essentially a fiction, consumed by the proc-macro to emit the real probe bits, but I think the advantage to your new module-scoped-probes implementation is that it more closely follows the ergonomics of normal rust structures:
You define some module (`probes`) and call the member functions (or in this case, macros) from it.  Having to name the module `dropshot` and then rename it from the proc-macro for how it exposes those probes to the source subverts those expectations, IMO.

bnaecker (commented):
Looks OK, though I'm with Dave that we can avoid renaming the provider.

dropshot/src/lib.rs
pfmooney: > To be explicit, if you want the probes available at a module called `crate::probes`, you'd do:
> 
> ```rust
> #[usdt::provider(probe_path = "probes")]
> mod dropshot {
>     ...
> }
> ```

Yeah, I understand it's possible.  I'm just suggesting that the naming be switched around in usdt so it be more like this:

```rust
#[usdt::provider(probe_provider = "dropshot")]
mod probes {
    ...
}
```

bnaecker: @pfmooney To make sure I understand, you're suggesting that the name of the module is always what's written there, and that's how consumers refer to the generated probe macros. What can be changed, though, is the name of the provider -- it defaults to the name of the module, but you could override it. That right?

It seems reasonable to me, but I'd love @ahl's thoughts too. Definitely another breaking API change, so we'll have to hold this PR behind that, if it's desirable.

pfmooney: Yeah, that the proc-macro not rename the module, but rather could be instructed to (re) name the provider as it is exposed to dtrace.

Also, I agree that this PR need to wait or be reworked to not rename the provider in the mean time.

bnaecker:
> > It should be noted that the dropshot tests don't complete successfully when run under the nightly cargo/compiler, so testing the probes via that method (since `asm` and thus nightly is required) is a no-go.
> 
> Is that a recent change, do you know?

I just tested `main` on the latest nightly, 1.58, and it does indeed fail at the compiler error messages tests. The same test suite, although different actual tests, fails on nightly-2021-03-25, which is used to format the code. I'm not sure what the right way forward here is. The compiler error messages are only different in superficial ways, so we could just overwrite them. But since the code isn't tested in CI with nightly, I'm not sure when this changed.

pfmooney:
I'm just going to correct this wad to get the provider name right (using the existing 0.1.19 syntax) so that downstream dropshot (and usdt) consumers can move forward, rather than gating it on future updates.

bnaecker:
@pfmooney Sounds good. I'm making updates as you suggested today, so I can put up a follow-up PR later with the changes. It might take a minute, since it's all backed up behind getting this feature-flag madness tamed.

pfmooney:
> @pfmooney Sounds good. I'm making updates as you suggested today, so I can put up a follow-up PR later with the changes. It might take a minute, since it's all backed up behind getting this feature-flag madness tamed.

Yeah, I don't want to impose any pressure in that regard, hence solving it this way for now.  I'm happy to go about the countryside switching the syntax later if/when you update `usdt` for the syntax I suggested.

pfmooney:
By building one of the shipped examples and attaching dtrace, I can confirm that the provider is appropriately named with the mentioned notation:
```
root@carbide:~# dtrace -p $(pgrep basic) -ln ':::request_start'
   ID   PROVIDER            MODULE                          FUNCTION NAME
    9 dropshot8511             basic _ZN8dropshot6server24http_request_handle_wrap28_$u7b$$u7b$closure$u7d$$u7d$17h1d5afea1a7cf5f71E request_start
```

davepacheco (approved):

PR #183, opened 2021-11-10
update for openapiv3 1.0.0-beta.5 and pin to that version

ahl:
@pfmooney sorry about that! I think this should address your issue.


pfmooney:
Thanks!

PR #185, opened 2021-11-11
Add `object_update` fn to `test_utils`

zephraph:
Adds an `update` option to the suite of `objects` helpers. The only real different here is I've passed the `status` as a param because PUTs could result in 200 or 204 on success.

### DIFF
--- a/dropshot/src/test_util.rs
+++ b/dropshot/src/test_util.rs
@@ -620,6 +620,21 @@ pub async fn objects_post<S: Serialize + Debug, T: DeserializeOwned>(
 }
 
 /**
+ * Issues an HTTP PUT to the specified collection URL to update an object.
+ */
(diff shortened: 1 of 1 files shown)


davepacheco (commented):
This seems fine but for what it's worth I think part of the reason for `object_delete` and `objects_post` was to ensure a consistent status code.  If we're not doing that here, is this still a useful shorthand?

Might call it `object_put` for consistency with `objects_post`, `object_delete`, and `object_get`.  (`objects_list` is admittedly different.)

zephraph:
Yeah, the response code being required was ultimately because we're going to want to test the different PUT cases. You could create a resource, update it, and then do an idempotent check with similar calling code.

zephraph:
@davepacheco I updated the name to `object_put`, that does seem a bit more consistent.

davepacheco (approved):

PR #189, opened 2021-11-16
Export ExtractorMetadata so consumers can impl Extractor

david-crespo:
You need `ExtractorMetadata` in order to be able to implement `Extractor`, e.g., if you want to write your own Cookies extractor.

https://github.com/oxidecomputer/dropshot/blob/61439b8b3ba84ff365b61d59f45254b8326a5012/dropshot/src/handler.rs#L179-L188


davepacheco:
I already merged it but we might also want to add an integration test that defines its own extractor to make sure we can do it.

PR #190, opened 2021-11-18
update CHANGELOG in prep for next release

davepacheco:
I went through the commit log looking for notable changes we might want to put into the changelog.


ahl (approved):
thanks for doing this; did a cursory scan of the log but didn't see anything else obviously needing documentation

PR #204, opened 2021-11-29
update for nightly-2021-11-24

ahl:
We're moving omicron and all dependents past the asm_sym flag day. See https://github.com/oxidecomputer/omicron/pull/448


davepacheco (approved):
I assume this is a breaking change to be added to the changelog, right?

PR #208, opened 2021-12-08
Adds conditional compilation directives around usdt code

bnaecker:
This allows consumers to use dropshot in a project where another package
may have `usdt` as a dependency, and enable the underlying `asm`
implementation. Because of feature unification, that causes `dropshot`
to use the `asm` implementation, but not include the `#![feature(asm)]`
attribute, leading to a compile failure.

### DIFF
--- a/dropshot/src/server.rs
+++ b/dropshot/src/server.rs
@@ -8,6 +8,7 @@ use super::config::ConfigDropshot;
 use super::error::HttpError;
 use super::handler::RequestContext;
 use super::http_util::HEADER_REQUEST_ID;
+#[cfg(feature = "usdt-probes")]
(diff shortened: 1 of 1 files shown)


davepacheco (approved):
This looks good.  Before merging it'd be nice to confirm this does what we expect in consumers of dropshot (i.e., maybe try modifying oxidecomputer/omicron#491 to just pull in this dropshot instead of changing any of the dropshot deps and see if this alone fixes that problem?)

bnaecker:
@davepacheco I should have mentioned, that I did that locally. I removed the `usdt-probes` flag from a consumer's `Cargo.toml`, and everything built as expected.

PR #215, opened 2021-12-15
futures 0.3.18 yanked, back to 0.3.17

rcgoodfellow:
See: https://docs.rs/futures/0.3.18/futures/index.html


davepacheco:
Thanks.  Looks like it was yanked due to rust-lang/futures-rs#2529.

PR #222, opened 2021-12-20
Updated file serving example for streaming

smklein:
Additionally, update test utils to accept `transfer-encoding` headers, which are transmitted alongside streams.

### DIFF
--- a/dropshot/src/test_util.rs
+++ b/dropshot/src/test_util.rs
@@ -35,16 +35,36 @@ use crate::logging::ConfigLogging;
 use crate::pagination::ResultsPage;
 use crate::server::{HttpServer, HttpServerStarter, ServerContext};
 
-/**
- * List of allowed HTTP headers in responses.  This is used to make sure we
- * don't leak headers unexpectedly.
- */
-const ALLOWED_HEADER_NAMES: [&str; 5] = [
-    "content-length",
-    "content-type",
-    "date",
-    "x-request-id",
-    "transfer-encoding",
+enum AllowedValue<'a> {
+    Any,
+    OneOf(&'a [&'a str]),
+}
+
+struct AllowedHeader<'a> {
+    name: &'a str,
+    value: AllowedValue<'a>,
+}
+
+impl<'a> AllowedHeader<'a> {
+    const fn new(name: &'a str) -> Self {
+        Self {
+            name,
+            value: AllowedValue::Any,
+        }
+    }
+}
+
+// List of allowed HTTP headers in responsees.
+// Used to make sure we don't leak headers unexpectedly.
+const ALLOWED_HEADERS: [AllowedHeader<'static>; 5] = [
+    AllowedHeader::new("content-length"),
+    AllowedHeader::new("content-type"),
+    AllowedHeader::new("date"),
+    AllowedHeader::new("x-request-id"),
+    AllowedHeader {
+        name: "transfer-encoding",
+        value: AllowedValue::OneOf(&["chunked"]),
+    },
 ];
 
 /**
@@ -204,11 +224,21 @@ impl ClientTestContext {
          * statically-defined above.
          */
         let headers = response.headers();
-        for header_name in headers.keys() {
+        for (header_name, header_value) in headers {
             let mut okay = false;
-            for allowed_name in ALLOWED_HEADER_NAMES.iter() {
-                if header_name == allowed_name {
-                    okay = true;
+            for allowed_header in ALLOWED_HEADERS.iter() {
+                if header_name == allowed_header.name {
+                    match allowed_header.value {
+                        AllowedValue::Any => {
+                            okay = true;
+                        }
+                        AllowedValue::OneOf(allowed_values) => {
+                            let header = header_value
+                                .to_str()
+                                .expect("Cannot turn header value to string");
+                            okay = allowed_values.contains(&header);
+                        }
+                    }
                     break;
                 }
             }
--- a/dropshot/src/test_util.rs
+++ b/dropshot/src/test_util.rs
@@ -39,8 +39,8 @@ use crate::server::{HttpServer, HttpServerStarter, ServerContext};
  * List of allowed HTTP headers in responses.  This is used to make sure we
  * don't leak headers unexpectedly.
  */
-const ALLOWED_HEADER_NAMES: [&str; 4] =
-    ["content-length", "content-type", "date", "x-request-id"];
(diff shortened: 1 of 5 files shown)


smklein:
This was inspired by @jclulow 's work in buildomat: https://github.com/oxidecomputer/buildomat/blob/main/server/src/main.rs#L270-L281

It seems like streaming file transfer should be our default example, as it integrates pretty well with Hyper.

I'm attempting to make use of this pattern in https://github.com/oxidecomputer/omicron/pull/457 , but tests are failing without the accepted `transfer-encoding` header support.

dropshot/src/test_util.rs
```diff
@@ -39,8 +39,13 @@ use crate::server::{HttpServer, HttpServerStarter, ServerContext};
...
  * don't leak headers unexpectedly.
  */
-const ALLOWED_HEADER_NAMES: [&str; 4] =
-    ["content-length", "content-type", "date", "x-request-id"];
+const ALLOWED_HEADER_NAMES: [&str; 5] = [
+    "content-length",
+    "content-type",
+    "date",
+    "x-request-id",
+    "transfer-encoding",
```
davepacheco, line 47: Do we want to check for a particular value of the "transfer-encoding" header?

Context: These tests ensure that Dropshot returns correct responses, including the headers, not just the body.  The allowlist here tries to make sure we're not leaking unexpected headers.  It makes sense to add this header to the list.  If a header is in this list, presumably we also want to test that Dropshot is providing the correct value for it.

smklein: Sure, I've added a test which checks for a specific value.

I'm not sure the test utility should care, admittedly. As an example, I believe "transfer-encoding = chunked" is an HTTP 1.1 feature, which HTTP 2.0 doesn't have - but most code acting on a response should work either way, since Hyper should be responsible for multiplexing the different implementations into the same stream-based API.

smklein:
Added a couple streaming tests, showing a streaming server, and a client which can either "read the whole response into memory" or opt to receive it piece-by-piece.

dropshot/tests/test_streaming.rs
```diff
@@ -0,0 +1,116 @@
+// Copyright 2021 Oxide Computer Company
+
+//! Test cases for streaming reqeusts.
```
davepacheco, line 3:
```suggestion
//! Test cases for streaming requests.
```

dropshot/tests/test_streaming.rs
```diff
@@ -0,0 +1,116 @@
...
+
+#[endpoint {
+    method = GET,
+    path = "/streaming",
+}]
+async fn api_streaming(
+    _rqctx: Arc<RequestContext<usize>>,
+) -> Result<Response<Body>, HttpError> {
+    let mut file = tempfile::tempfile()
+        .map_err(|_| HttpError::for_bad_request(None, "EBADF".to_string()))
```
davepacheco, line 34: nit: can we use some other string here?  If I see EBADF anywhere I assume it came from errno, and those are really awful to debug.

dropshot/tests/test_streaming.rs
```diff
@@ -0,0 +1,116 @@
...
+const BUF_COUNT: usize = 128;
+
+#[endpoint {
+    method = GET,
+    path = "/streaming",
+}]
+async fn api_streaming(
+    _rqctx: Arc<RequestContext<usize>>,
+) -> Result<Response<Body>, HttpError> {
+    let mut file = tempfile::tempfile()
```
davepacheco, line 33: Is there any particular reason to use an actual file instead of, say, a [Cursor](https://doc.rust-lang.org/std/io/struct.Cursor.html)?  Or maybe better: a custom Stream that just emits a fixed number of arbitrary bytes?

It'd be neat to have this stream emit some number of bytes that's likely to be larger than available memory.  If the test works, that would validate that we're not at any point buffering the whole contents of the response anywhere.

dropshot/tests/test_streaming.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    let file_stream = FileBytesStream::new(file);
+    Ok(Response::builder()
+        .status(StatusCode::OK)
+        .body(file_stream.into_body())?)
+}
+
+fn check_has_chunked_headers(response: &Response<Body>) {
+    let transfer_encoding_header =
+        response.headers().get("transfer-encoding").expect(
```
davepacheco, line 53: We maybe want to check that this header is _not_ set when we're not streaming.  I guess that's implied if it's not one of the allowed headers already?

dropshot/src/test_util.rs
davepacheco: I'm not sure I follow.  What I mean is: in general, "transfer-encoding" [appears to have only a handful of individual values](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding), plus comma-separated lists of those values.  If Dropshot sent "transfer-encoding: oxide", that would be broken.  It would also be broken if it sent "transfer-encoding: chunked" and then sent a response that wasn't chunked-encoded, although this is presumably done by hyper and such breakage would almost certainly cause the request to fail on the client anyway.

I could see asserting:

- that the value is either absent or "chunked" (knowing that we don't support the other values)
- that the value is absent unless we're hitting an API that's intentionally streaming.  Maybe this is what you're getting at -- it wouldn't be wrong if Dropshot chose to send non-streaming responses with "chunked", but it would be surprising and we might want to know if that happens without someone intentionally making that change.

davepacheco (approved):
These comments are all nitty.

dropshot/src/test_util.rs
smklein: Sure, I'll add a toggle here for expecting headers with "any" value, or with a "known set of okay values". This will only permit "chunked" for now, but could let us permit a set of values in the future.

dropshot/tests/test_streaming.rs
smklein: I used a file to show integration with the [hyper_staticfile::FileBytesStream](https://docs.rs/hyper-staticfile/0.8.0/hyper_staticfile/struct.FileBytesStream.html), since that was a use case that motivated adding this support - we have a file server example in here, but IMO streaming files is a *really* common use-case that I wanted to show.

dropshot/tests/test_streaming.rs
smklein: Sure, I'll add another test with a non-streaming endpoint.

PR #224, opened 2021-12-23
Add initial TLS support to dropshot

teisenbe:
This adds new configuration parameters `https=<bool>`, `cert_file=<string>`, and `key_file=<string>` for making Dropshot listen for HTTP-over-TLS connections instead of plain HTTP. When https=true, only TLS connections will be accepted, and when https=false, only plain HTTP connections will be accepted.

This is built off of George's WIP in #125


teisenbe:
After I'm back from the holidays, I plan to add new suite of tests for the https support, covering some of the behavior around the certificate loading and supported protocol modes, and making sure that non-TLS connections fail. Also fixing up the broken doc test and lints

teisenbe:
Inspecting advertised options from the client side isn't supported by any libraries directly, though the libcurl bindings let you get access to the raw TLS stream bytes. It might not be feasible with the state of libraries right now to easily extract the data I would want for the test cases, so I'm inclined to skip that last test case. Any thoughts?

dropshot/tests/test_config.rs
```diff
@@ -74,15 +171,15 @@ async fn test_config_bind_address() {
...
 
     let client = hyper::Client::new();
     let bind_ip_str = "127.0.0.1";
-    let bind_port: u16 = 12215;
+    let bind_port = 12215;
 
     /*
      * This helper constructs a GET HTTP request to
-     * http://$bind_ip_str:$port/, where $port is the argument to the
+     * https://$bind_ip_str:$port/, where $port is the argument to the
```
smklein, line 182: Should this be `http`?

dropshot/tests/test_config.rs
```diff
@@ -140,13 +233,124 @@ async fn test_config_bind_address() {
...
+
+    let client = make_client();
+    let hostname = "localhost";
+    let bind_ip_str = "127.0.0.1";
+    /* This must be different than the bind_port used in the http test. */
+    let bind_port = 12217;
+
+    /*
+     * This helper constructs a GET HTTP request to
+     * http://$bind_ip_str:$port/, where $port is the argument to the
```
smklein, line 296: Should this be `https`?

smklein:
> Inspecting advertised options from the client side isn't supported by any libraries directly, though the libcurl bindings let you get access to the raw TLS stream bytes. It might not be feasible with the state of libraries right now to easily extract the data I would want for the test cases, so I'm inclined to skip that last test case. Any thoughts?

For Clients verifying Servers:

I was eyeballing `rustls::ClientConnection`  (accessible from [tokio_rustls::client::TlsStream](https://docs.rs/tokio-rustls/latest/tokio_rustls/client/struct.TlsStream.html#method.get_ref)). It looks like there are some methods that allow examining the server side:
- https://docs.rs/rustls/0.20.2/rustls/client/struct.ClientConnection.html#method.peer_certificates
- https://docs.rs/rustls/0.20.2/rustls/client/struct.ClientConnection.html#method.protocol_version

Also, the client seems like it can verify the Server Certificates using a custom configuration option: https://docs.rs/rustls/0.20.2/rustls/struct.ConfigBuilder.html#method.with_custom_certificate_verifier

For Servers verifying Clients:
- Calling https://docs.rs/rustls/0.20.2/rustls/struct.ConfigBuilder.html#method.with_client_cert_verifier with https://docs.rs/rustls/latest/rustls/server/struct.AllowAnyAuthenticatedClient.html  seems like  the easiest way to do this, but a more specific `ClientCertVerifier` seems like it could work.

It does kinda seem like rustls tries to make all the verification occur at the cert-level, and all subsequent access plaintext-only. I can see how this might make certain aspects of testing harder, but it *does* also seem kinda nice from a correctness point-of-view.

Dunno though - was that in scope of what you were hoping to test?

smklein (commented):
This looks great - I appreciate the concurrency you pushed for here (i.e., in `HttpsAcceptor`) while also including detailed docs!

My only comments are basically nits, I'm really excited to start pushing further on integration with the rest of our codebase.

dropshot/examples/https.rs
```diff
@@ -0,0 +1,186 @@
...
+ */
+fn generate_keys() -> Result<(NamedTempFile, NamedTempFile), String> {
+    let keypair =
+        rcgen::generate_simple_self_signed(vec!["localhost".to_string()])
+            .map_err(|e| e.to_string())?;
+    let cert = keypair.serialize_pem().map_err(|e| e.to_string())?;
+    let priv_key = keypair.serialize_private_key_pem();
+
+    let mut cert_file =
+        NamedTempFile::new().map_err(|_| "failed to create cert_file")?;
```
davepacheco, line 39: I really wish there were a way to include a name with this file.  That way if this crashes and someone finds these files they'll have some idea what they are and that they can probably delete them.  But I don't see an easy way to do this with the `tempfile` crate.  Oh well.

dropshot/examples/https.rs
```diff
@@ -0,0 +1,186 @@
...
+    Ok((cert_file, key_file))
+}
+
+#[tokio::main]
+async fn main() -> Result<(), String> {
+    /*
+     * Begin by generating TLS certificates and keys. A normal application would
+     * just pass the paths to these via ConfigDropshot.
+     */
+    let (cert_file, key_file) = generate_keys()?;
```
davepacheco, line 57: I love this example.  It concisely shows what you need to make a regular Dropshot server a TLS server.

dropshot/src/config.rs
```diff
@@ -49,13 +50,35 @@ pub struct ConfigDropshot {
...
+     * should think about making it default to true in the future. */
+    pub https: bool,
+    /** Path to a PEM file containing a certificate chain for the
+     *  server to identify itself with. The first certificate is the
+     *  end-entity certificate, and the remaining are intermediate
+     *  certificates on the way to a trusted CA.
+     *
+     *  Only valid if https=true
+     */
+    pub cert_file: PathBuf,
```
davepacheco, line 65: It seems a little confusing to have values for these properties when TLS is not configured, even if the value is just the empty PathBuf.  Would it make sense for `https` to be an enum with variants like `None` and `Tls(TlsConfig)`, where `TlsConfig` is a struct with `cert_file` and `key_file`?  I wonder if you could do this here with one of the [serde enum tagging modes](https://serde.rs/enum-representations.html).

dropshot/tests/test_demo.rs
```diff
@@ -201,10 +200,7 @@ async fn test_demo2json() {
     let testctx = common::test_setup("demo2json", api);
 
     /* Test case: optional field */
-    let input = DemoJsonBody {
-        test1: "bar".to_string(),
-        test2: None,
-    };
+    let input = DemoJsonBody { test1: "bar".to_string(), test2: None };
```
davepacheco, line 203: I wonder if this file got reformatted.  It seems like a lot of this diff is spurious.

dropshot/tests/test_pagination.rs
```diff
@@ -54,7 +54,7 @@ extern crate slog;
 #[macro_use]
```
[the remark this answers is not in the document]
smklein: errrr I think this was resolved. Must have been looking at an old commit.

dropshot/tests/test_config.rs
```diff
@@ -140,13 +233,124 @@ async fn test_config_bind_address() {
...
+
+    /*
+     * Make another request to make sure it fails now that we've shut down
+     * the server.  We need a new client to make sure our client-side connection
+     * starts from a clean slate.  (Otherwise, a race during shutdown could
+     * cause us to successfully send a request packet, only to have the TCP
+     * stack return with ECONNRESET, which gets in the way of what we're trying
+     * to test here.)
+     */
+    let client = make_client();
```
davepacheco, line 345: It seems like a bunch of this function is duplicated with the earlier one.  Is it worth trying to share code here?

dropshot/tests/test_tls.rs
```diff
@@ -0,0 +1,204 @@
...
+use std::convert::TryFrom;
+use std::path::Path;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+use std::time::SystemTime;
+
+pub mod common;
+
+fn create_log(test_name: &str) -> slog::Logger {
+    let log_path = dropshot::test_util::log_file_for_test(test_name)
```
davepacheco, line 19: I think you want to use `LogContext::new` here.  I'm not sure `log_file_for_test` should even be exposed from test_util.rs.  (Basically, I'm worried these tests leak the log file.)

dropshot/src/server.rs
```diff
@@ -149,49 +225,294 @@ impl<C: ServerContext> HttpServerStarter<C> {
...
+                            Ok(conn) => yield Ok(conn),
+                            Err(e) => {
+                                // If TLS negotiation fails, log the cause but
+                                // don't forward it along. Yielding an error
+                                // from here will terminate the server.
+                                // TODO: We may want to export a counter for
+                                // different error types, since this may contain
+                                // useful things like "your certificate is
+                                // invalid"
+                                info!(log, "tls accept err: {}", e);
```
davepacheco, line 363: I would think this would be `error!` or at least `warn!`.

dropshot/src/server.rs
```diff
@@ -149,49 +225,294 @@ impl<C: ServerContext> HttpServerStarter<C> {
...
+                    },
+                    accept_result = tcp_listener.accept() => {
+                        let (socket, addr) = match accept_result {
+                            Ok(v) => v,
+                            Err(e) => {
+                                match e.kind() {
+                                    std::io::ErrorKind::ConnectionAborted => {
+                                        continue;
+                                    },
+                                    // The other errors that can be returned
```
davepacheco, line 375: Can we add a TODO-resilience or TODO-robustness or something here?  I want to make sure we look at this when it comes to hardening.

dropshot/src/server.rs
```diff
@@ -149,49 +225,294 @@ impl<C: ServerContext> HttpServerStarter<C> {
...
+                tokio::select! {
+                    Some(negotiation) = tls_negotiations.next(), if
+                            !tls_negotiations.is_empty() => {
+
+                        match negotiation {
+                            Ok(conn) => yield Ok(conn),
+                            Err(e) => {
+                                // If TLS negotiation fails, log the cause but
+                                // don't forward it along. Yielding an error
+                                // from here will terminate the server.
```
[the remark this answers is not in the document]
teisenbe: Yeah, I had the same original interpretation, and the code was originally yielding.  When I was testing, a thing that would trigger this was Chrome sending a cert mismatch alert message to fail the negotiation (due to self-signed certs), which would then cause the server to exit. Though this makes me realize I should make sure we have a test case to hit this :)

dropshot/tests/test_config.rs
teisenbe: Yes, thank you. I apparently updated the wrong comment

dropshot/tests/test_pagination.rs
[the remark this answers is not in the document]
teisenbe: Yeah, in the older commits I hadn't used the cargo fmt thing since I'd forgotten, but the pile of lint errors in the build was a good reminder

davepacheco (commented):
This looks great!  Thanks for taking it on.

dropshot/Cargo.toml
```diff
@@ -84,6 +84,10 @@ pem = "1.0"
 rcgen = "0.8.14"
 tempfile = "3.2.0"
 
+[dev-dependencies.rustls]
+version = "0.20.2"
+features = [ "dangerous_configuration" ]
```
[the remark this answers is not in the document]
teisenbe: Yeah, the rustls docs are a little inaccurate there. It's also needed (though not visibly in the docs) to set a custom certificate verifier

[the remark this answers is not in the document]
teisenbe: https://docs.rs/rustls/latest/src/rustls/client/builder.rs.html#34-49

dropshot/tests/test_pagination.rs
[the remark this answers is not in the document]
teisenbe: Hmm, it wasn't after all. My editor was dueling with the manual `cargo +nightly-2021-03-25 fmt`, fixed

dropshot/Cargo.toml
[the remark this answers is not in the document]
smklein: Ack, thanks. This usage makes sense then!

dropshot/src/server.rs
teisenbe: It looks like at least for fatal alerts, the returned error does not even specify which fatal alert occurred

dropshot/examples/https.rs
teisenbe: I found one, using tempfile::Builder. I'll make use of it here and in the tests

dropshot/src/config.rs
teisenbe: Yeah, agreed. I was having trouble thinking of how to represent this into YAML. I'll dig into the enum tagging stuff

dropshot/tests/test_tls.rs
teisenbe: Whoops, copied this from test_config. Thanks! I'll see about fixing things over there, too

dropshot/tests/test_config.rs
teisenbe: I think so, in case we want to expand any test cases. Changed it

dropshot/tests/test_demo.rs
teisenbe: Looks like it, gone in latest patch

teisenbe:
> > Inspecting advertised options from the client side isn't supported by any libraries directly, though the libcurl bindings let you get access to the raw TLS stream bytes. It might not be feasible with the state of libraries right now to easily extract the data I would want for the test cases, so I'm inclined to skip that last test case. Any thoughts?
> 
> For Clients verifying Servers:
> 
> I was eyeballing `rustls::ClientConnection` (accessible from [tokio_rustls::client::TlsStream](https://docs.rs/tokio-rustls/latest/tokio_rustls/client/struct.TlsStream.html#method.get_ref)). It looks like there are some methods that allow examining the server side:
> 
>     * https://docs.rs/rustls/0.20.2/rustls/client/struct.ClientConnection.html#method.peer_certificates
> 
>     * https://docs.rs/rustls/0.20.2/rustls/client/struct.ClientConnection.html#method.protocol_version
> 
> 
> Also, the client seems like it can verify the Server Certificates using a custom configuration option: https://docs.rs/rustls/0.20.2/rustls/struct.ConfigBuilder.html#method.with_custom_certificate_verifier
> 
> For Servers verifying Clients:
> 
>     * Calling https://docs.rs/rustls/0.20.2/rustls/struct.ConfigBuilder.html#method.with_client_cert_verifier with https://docs.rs/rustls/latest/rustls/server/struct.AllowAnyAuthenticatedClient.html  seems like  the easiest way to do this, but a more specific `ClientCertVerifier` seems like it could work.
> 
> 
> It does kinda seem like rustls tries to make all the verification occur at the cert-level, and all subsequent access plaintext-only. I can see how this might make certain aspects of testing harder, but it _does_ also seem kinda nice from a correctness point-of-view.
> 
> Dunno though - was that in scope of what you were hoping to test?

I was hoping more for being able to enumerate configurations offered by the server during negotiation so we can have tests that mark our expectations for suites being supported or not supported. Some of these are precluded by rustls just not implementing known bad ciphersuites and TLS versions, but it would be nice to write something that would catch a regression from us swapping out TLS libraries.

I think I can write these tests with existing client libraries (might need to grab one besides rustls though) by constraining the client to only accept a single negotiated mode and inspecting the result.

Let me know if you think these tests don't seem worth the effort given we don't have plans to change from rustls any time soon

dropshot/src/config.rs
teisenbe: Err, TOML, not YAML :) YAML would be rather straightforward

smklein:
> I was hoping more for being able to enumerate configurations offered by the server during negotiation so we can have tests that mark our expectations for suites being supported or not supported. Some of these are precluded by rustls just not implementing known bad ciphersuites and TLS versions, but it would be nice to write something that would catch a regression from us swapping out TLS libraries.
> 
> I think I can write these tests with existing client libraries (might need to grab one besides rustls though) by constraining the client to only accept a single negotiated mode and inspecting the result.
> 
> Let me know if you think these tests don't seem worth the effort given we don't have plans to change from rustls any time soon

The test does seem potentially useful... the only odd part might be ensuring we have a *disliked* version of TLS to test against.

This seems doable with the `reqwest` library, if you wanna spin up an arbitrary client:

- https://docs.rs/reqwest/0.11.8/reqwest/struct.ClientBuilder.html#method.max_tls_version

Honestly, I'll defer to your judgement. It feels like a fine line between "testing Dropshot" vs "testing rustls" - if what you had in mind is doable easily with that API, I say "go for it", otherwise, might not be worth the cost right now.

dropshot/src/config.rs
teisenbe: I had completely forgotten that TOML has dictionary types, which will likely be the way I'll handle this. I'll probably just do an Option<TlsConfig> for now, since I don't really know of a case for us doing some form of https that isn't TLS

teisenbe: Please take a look, I've reworked it

teisenbe:
> 

Yeah. I don't think those APIs are enough for what I wanted. It's probably fine to skip it. As new TLS vulnerabilities are found, we can add tests to ensure we've disabled various unsafe modes

smklein (approved):

dropshot/examples/https.rs
```diff
@@ -35,13 +36,19 @@ fn generate_keys() -> Result<(NamedTempFile, NamedTempFile), String> {
     let cert = keypair.serialize_pem().map_err(|e| e.to_string())?;
     let priv_key = keypair.serialize_private_key_pem();
 
+    let make_temp = || {
+        tempfile::Builder::new()
+            .prefix("dropshot-https-example-")
```
davepacheco, line 41: Thanks for doing this!

dropshot/src/config.rs
```diff
@@ -51,22 +56,20 @@ pub struct ConfigDropshot {
     /** maximum allowed size of a request body, defaults to 1024 */
     pub request_body_max_bytes: usize,
 
-    // TLS settings
-    /** Whether to use HTTPS. For now, this defaults to false, but we
-     * should think about making it default to true in the future. */
-    pub https: bool,
+    /** If present, enables TLS with the given configuration */
+    pub tls: Option<ConfigTls>,
```
davepacheco, line 60: Love this.

dropshot/tests/common/mod.rs
```diff
@@ -29,15 +29,18 @@ pub fn test_setup(
...
+    let log = logctx.log.new(o!());
+    TestContext::new(api, 0 as usize, &config_dropshot, Some(logctx), log)
+}
+
+pub fn create_log_context(test_name: &str) -> LogContext {
+    let log_config = ConfigLogging::File {
         level: ConfigLoggingLevel::Debug,
         path: "UNUSED".to_string(),
-        if_exists: ConfigLoggingIfExists::Fail,
+        if_exists: ConfigLoggingIfExists::Append,
```
davepacheco, line 41: Why this change?  Isn't the name supposed to be unique?

teisenbe: Accidental, will change. I had copied that block from a different test, which was using Append, and didn't notice the difference

davepacheco (approved):

PR #234, opened 2022-01-07
create a minimal dropshot server example

jmpesp:
this commit adds a minimal dropshot server example that users can refer
to when starting their own projects.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "anyhow"
+version = "1.0.52"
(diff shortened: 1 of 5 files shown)


jmpesp:
Nevermind! Ignore this, I found the `examples` directory...

smklein:
Have you seen https://github.com/oxidecomputer/dropshot/tree/main/dropshot/examples ? Just wondering where that fell short for you

smklein:
ah I see you found it! maybe we should add a more clear link in the README...

Issue #77, opened 2021-01-05
better understand rustfmt stability


davepacheco:
See #76 for background.  I wanted to see if there was a better approach here.  I found [that rustfmt intends that your code should _not_ be reformatted if you stick to stable options](https://github.com/rust-lang/rustfmt/issues/4286).  From my testing, this appears to be true both with stable options _and_ with the unstable options that we're using in Dropshot.  (We previously thought that we were seeing a bunch of churn.  I now suspect that churn was a result of accidentally switching between nightly and stable versions of otherwise equivalent `rustfmt` releases.  Details below.)

Given that, I see three options here:

1. Keep things the way they are.  This means we document in the README that users must use a nightly release for `rustfmt` and we tell them which specific release to use.  Developers are expected to configure this correctly for themselves, whether that's vim, VScode (settings for which are in this repo), or muscle memory (`cargo +nightly-2020-07-26 fmt` rather than `cargo fmt`).
2. Keep things mostly the same, but remove `require_version` from `rustfmt.toml`.  We document that developers must use `nightly`, and that we expect any nightly is likely to work.  (If we run into cases where a specific nightly starts formatting the code differently, we'll see that in the pre-integration check.)  In this case, we're using nightly only for its _policy_ that unstable options are supported, not because we need a bleeding-edge version of the software.  Like option (1), developers still have to configure things locally.  On the plus side, we don't have to keep bumping the specific `require_version`, CI config, or README instructions.  Right now, there's basically no reason to _ever_ bump these unless that build stops becoming available or we adopt a newer rustfmt that actually changes the way code is formatted.  But people might be confused about why our instructions reference an ancient 2020-07 nightly build or suggest that we update it.  Instead, the CI can use the known-good version until we need to update it, and everybody else can just see and remember "nightly" rather than "nightly-2020-07-26".  (This might be a pretty big advantage for people that use "nightly" for other purposes -- like other projects that require "nightly" -- since they can use that toolchain for Dropshot, too.)
3. Drop all the unstable options altogether, including `require_version`.  Keep us on stable rustfmt.  This is smoothest for everybody -- you only need one toolchain, you don't need any custom local config, and there's nothing to forget about.  

Option 3 sounds great, but some of the defaults really do seem unfortunate.  Here's the result of reformatting today's Dropshot with the default options:
https://github.com/oxidecomputer/dropshot/commit/7d12647f9ae93c1b7d31f14dfd1a7aede4e6621b

[...]

PR #242, opened 2022-01-10
Simplify rustfmt config and switch to stable rustfmt

davepacheco:
See #77 for lots of background.  About a year ago, we switched to a simpler stable-only rustfmt config in omicron to try it out.  That's been going well.  It's time to apply that here, too.

This change:

- updates the rustfmt config file to match Omicron's
- updates the GitHub actions config to stop specifying a specific nightly for rustfmt
- updates the documentation
- reformats all code to match the new config (just by running `cargo fmt`)

This has a (trivial) conflict with #139 so I've based this on #139 because I expect that to land shortly.  I've set the branch to "pin-toolchain" to make the diff easier to read.


.vscode/settings.json
```diff
@@ -1,8 +0,0 @@
-{
```
davepacheco, line 1: @ahl I believe it's correct to remove this file now.  I think we did the same for omicron when we switch to stable rustfmt.

.github/workflows/rust.yml
```diff
@@ -14,17 +14,12 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@230611dbd0eb52da1e1f4f7bc8bb0c3a339fc8b7
-      - uses: actions-rs/toolchain@88dc2356392166efad76775c878094f4e83ff746
```
davepacheco: @zephraph I think this change is correct -- we don't have to specify this action any more in this job, right?

david-crespo (approved):

.github/workflows/rust.yml
smklein: I thought this action was just to get access to *any* rust toolchain, and we needed:

- "nightly-2021-03-25" here, and
- "Whatever is in the rust-toolchain.toml" file below

But it seems like, based on the "checks" output, we actually happen to be downloading the version we need (specified in the rust-toolchain.toml) on the first call to `cargo --version`.

Checking my understanding: With this new change, are we just relying on whatever happens to be installed on Github's default ubuntu-18.04 image to bootstrap?

davepacheco: I don't believe we're relying on whatever happens to be installed because we have a rust-toolchain file in the repo.

davepacheco: To expans on that: [the Actions environment, at least for Ubuntu 18.04, includes Rust 1.57](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-Readme.md#rust-tools).  I believe this is updated pretty regularly when a new Rust release comes out.  So this will likely already be the right thing when this runs.  But if it's not (e.g., if they've updated to 1.58 and we have not yet updated our rust-toolchain file), then the Rust that _is_ installed by default will honor the rust-toolchain file like it always does, installing the specified toolchain and running the rest of the build with it.

smklein: Right, that's what I meant by "bootstrap" in my comment above. Sounds reasonable, if we expect that'll always exist in the Actions environment across our supported test systems.

smklein (approved):

.github/workflows/rust.yml
davepacheco: I think that's right.  It's documented to be present in the Actions environments for Ubuntu, Windows, and Mac.  I would imagine it would be a pretty big breaking change if it were removed.  If it were, we could add this action back.

We might also add the action back if we wanted to override rust-toolchain to do builds with nightly or beta or some other stable (e.g., a MSRV).

PR #247, opened 2022-01-14
next_token and page_token -> camelCase

ahl:
This is a breaking change and I will document it once I have a PR number

### DIFF
--- a/dropshot/examples/pagination-multiple-sorts.rs
+++ b/dropshot/examples/pagination-multiple-sorts.rs
@@ -5,8 +5,8 @@
  *
  * When you run this program, it will start an HTTP server on an available local
  * port.  See the log for example URLs to use.  Try passing different values of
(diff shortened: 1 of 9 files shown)


davepacheco:
These are query parameters right?  Do we want those to be camel case?  I thought snake case was more idiomatic for them.

ahl:
1. this covers both the page_token / pageToken parameter *and* the next_page / nextPage object property
2. I think (based on https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#components-object-example) that camelCase is more typical for both parameters and properties, but ...
3. we should choose what we want and the categories I see are: operation IDs (snake_case), type names (PascalCase), object properties (camelCase?), and parameters (?)

ahl:
after looking into this a bit more, I think that we should use snake case for properties and parameters. A question for another day would be if dropshot consumers wanted to make a different choice, but we can address that when it arises.

PR #252, opened 2022-01-21
Populate OpenAPI summary field with first line of endpoint doc comments

### LINKED ISSUE #244: generate openapi summary
this will be the first line of the doc comment with the longer description being the rest if it is longer

ahl:
Fixes #244 

This change causes Dropshot to infer that the first line of the doc commend on an endpoint is the summary; subsequent lines (if there are any) are the description.

Note that there's going to be a bunch of follow-on work required in omicron to properly document endpoints.

### DIFF
--- a/dropshot_endpoint/Cargo.toml
+++ b/dropshot_endpoint/Cargo.toml
@@ -13,12 +13,15 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1"
 quote = "1"
-serde_tokenstream = "0.1.3"
+serde_tokenstream = "0.1"
 
 [dependencies.serde]
 version = "1.0"
 features = [ "derive" ]
 
 [dependencies.syn]
-version = "1.0.85"
+version = "1.0"
 features = [ "parsing", "printing" ]
+
+[dev-dependencies]
+schema = "0.0.1"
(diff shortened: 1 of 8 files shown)


dropshot_endpoint/Cargo.toml
```diff
@@ -13,12 +13,15 @@ proc-macro = true
...
-serde_tokenstream = "0.1.3"
+serde_tokenstream = "0.1"
 
 [dependencies.serde]
 version = "1.0"
 features = [ "derive" ]
 
 [dependencies.syn]
-version = "1.0.85"
+version = "1.0"
```
davepacheco, line 23: This is basically unrelated: I'm all for this but FYI I think it's dependabot that's replacing these with more specific versions when it updates deps.  It doesn't always do this and I'm not sure exactly when it does.  If the Renovate experiment goes well I hope it will give us more control over this.

dropshot_endpoint/src/lib.rs
```diff
@@ -1104,4 +1152,103 @@ mod tests {
             Some("Endpoint requires arguments".to_string())
         );
     }
+
+    #[test]
```
davepacheco, line 1156: Nice tests.

davepacheco (approved):

dropshot_endpoint/src/lib.rs
ahl: Note the use of my dope schema crate! ;-)

PR #271, opened 2022-02-11
Remove the identifier restriction on path components

ahl:
Dropshot deserializes pairs of path component variable names and actual path component values into a type parameter to a Path Extractor. Typically these will be valid Rust identifiers, but they don't need to be e.g. if the consumer were to use `#[serde(rename = "while")]` to rename the field during deserialization. The only restriction we need to enforce is that the path components have *a* name (i.e. it is non-empty). A mismatch would be detected at runtime... as happens already independent of whether renamed fields are valid Rust identifiers.

### DIFF
--- a/dropshot/src/router.rs
+++ b/dropshot/src/router.rs
@@ -15,7 +15,6 @@ use http::StatusCode;
 use percent_encoding::percent_decode_str;
 use std::collections::BTreeMap;
 use std::collections::BTreeSet;
-use std::sync::Once;
 
 /**
  * `HttpRouter` is a simple data structure for routing incoming HTTP requests to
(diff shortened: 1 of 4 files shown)


dropshot/src/router.rs
```diff
@@ -146,11 +145,15 @@ impl PathSegment {
                 (var, None)
             };
 
+            /*
+             * Note that the only constraint on the variable name is that it is
+             * not empty. Consumers may choose odd names like '_' or 'type'
+             * that are not valid Rust identifiers and use rename them with
```
davepacheco, line 151:
```suggestion
             * that are not valid Rust identifiers and rename them with
```

davepacheco (approved):

dropshot/src/router.rs
ahl: Thanks. I had something like "use serde(rename = ...)" and clearly only half-unwound that idea...

PR #276, opened 2022-02-15
allow consumers to add headers to responses

ahl:
This it to unblock @bnaecker 

@davepacheco I can see other ways of doing this (as we discussed); this seemed the least invasive, but I'm happy to weave this in more seamlessly if it seems objectionable.

### DIFF
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -96,8 +96,19 @@ pub struct RequestContext<Context: ServerContext> {
     pub path_variables: VariableSet,
     /** unique id assigned to this request */
     pub request_id: String,
+    /** mutable state for the request; the details of this are internal */
+    pub mutable_state: std::sync::Mutex<RequestContextMutableState>,
     /** logger for this specific request */
     pub log: Logger,
+}
+
+/**
+ * Container for state that consumers can change while handling a request.
+ */
+#[derive(Default)]
+pub struct RequestContextMutableState {
+    response_headers:
+        Vec<(http::header::HeaderName, http::header::HeaderValue)>,
 }
 
 impl<Context: ServerContext> RequestContext<Context> {
@@ -138,6 +149,21 @@ impl<Context: ServerContext> RequestContext<Context> {
              * default.
              */
             .unwrap_or(server_config.page_default_nitems))
+    }
+
+    /**
+     * Add a header that will be included with the HTTP response.
+     */
+    pub fn add_response_header(
+        self: Arc<Self>,
+        header: http::header::HeaderName,
+        value: http::header::HeaderValue,
+    ) {
+        self.mutable_state
+            .lock()
+            .unwrap()
+            .response_headers
+            .push((header, value));
     }
 }
 
@@ -486,9 +512,18 @@ where
          * resolved statically.
          */
         let rqctx = Arc::new(rqctx_raw);
+        let response_rqctx = rqctx.clone();
         let funcparams = Extractor::from_request(Arc::clone(&rqctx)).await?;
         let future = self.handler.handle_request(rqctx, funcparams);
-        future.await
+
+        future.await.map(|mut response| {
+            let mutable_state = response_rqctx.mutable_state.lock().unwrap();
+
+            for (header, value) in &mutable_state.response_headers {
+                response.headers_mut().insert(header, value.clone());
+            }
+            response
+        })
     }
 }
 
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -107,8 +107,7 @@ pub struct RequestContext<Context: ServerContext> {
  */
 #[derive(Default)]
 pub struct RequestContextMutableState {
-    response_headers:
-        Vec<(http::header::HeaderName, http::header::HeaderValue)>,
+    response_headers: http::header::HeaderMap,
 }
 
 impl<Context: ServerContext> RequestContext<Context> {
@@ -152,18 +151,50 @@ impl<Context: ServerContext> RequestContext<Context> {
     }
 
     /**
-     * Add a header that will be included with the HTTP response.
+     * Set a header that will be included with the HTTP response.
+     *
+     * If the header was previously unspecified, then None is returned.
+     *
+     * If the header was previously specified, the new value is associated with
+     * the key and all previous values are removed.
      */
-    pub fn add_response_header(
-        self: Arc<Self>,
-        header: http::header::HeaderName,
+    pub fn set_response_header<K>(
+        self: &Arc<Self>,
+        header: K,
         value: http::header::HeaderValue,
-    ) {
+    ) -> Option<http::header::HeaderValue>
+    where
+        K: http::header::IntoHeaderName,
+    {
         self.mutable_state
             .lock()
             .unwrap()
             .response_headers
-            .push((header, value));
+            .insert(header, value)
+    }
+
+    /**
+     * Set a header that will be included with the HTTP response.
+     *
+     * If the header was previously unspecified, then `false` is returned.
+     *
+     * If the header was previously specified, the new value is pushed to the
+     * end of the list of values associated with the key, and `true` is
+     * returned.
+     */
+    pub fn append_response_header<K>(
+        self: &Arc<Self>,
+        header: K,
+        value: http::header::HeaderValue,
+    ) -> bool
+    where
+        K: http::header::IntoHeaderName,
+    {
+        self.mutable_state
+            .lock()
+            .unwrap()
+            .response_headers
+            .append(header, value)
     }
 }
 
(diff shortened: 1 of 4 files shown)


dropshot/src/test_util.rs
```diff
@@ -51,9 +51,9 @@ impl<'a> AllowedHeader<'a> {
     }
 }
 
-// List of allowed HTTP headers in responsees.
+// List of allowed HTTP headers in responses.
 // Used to make sure we don't leak headers unexpectedly.
-const ALLOWED_HEADERS: [AllowedHeader<'static>; 5] = [
+const ALLOWED_HEADERS: [AllowedHeader<'static>; 6] = [
```
bnaecker, line 56: To make sure I understand, we're explicitly whitelisting headers that can be inserted, right? Does this get used anywhere? For example, it might be good to have `add_response_header` return a `Result`, if one tries to add an unsupported header.

bnaecker (commented):
Thanks for starting this work!

dropshot/src/test_util.rs
ahl: This is only used in our tests to validate that the resulting headers are what we expect. I don't think we need to do any validation of headers beyond what `HeaderName` already does for us. That might change in the future; I'm not sure if it's worth making `add_response_header` return a `Result<(), Error>` now -- thoughts?

dropshot/src/handler.rs
```diff
@@ -96,10 +96,21 @@ pub struct RequestContext<Context: ServerContext> {
...
     pub log: Logger,
 }
 
+/**
+ * Container for state that consumers can change while handling a request.
+ */
+#[derive(Default)]
+pub struct RequestContextMutableState {
+    response_headers:
+        Vec<(http::header::HeaderName, http::header::HeaderValue)>,
```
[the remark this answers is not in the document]
ahl: I guess I could see it either way. If it was a mapping would your proposal be that we return an error if there's a collision?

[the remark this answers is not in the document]
davepacheco: I believe multiple headers with the same name are allowed in HTTP, and it looks like [hyper's representation](https://docs.rs/hyper/latest/hyper/header/struct.HeaderMap.html) is a multimap.  You could even use that here.

I'd suggest we provide some combination of:

- "set" header -- replaces any previous value (corresponds to `HeaderMap::insert`)
- "append" header -- adds a new header (if you use this multiple times, you'll get multiple values for this header in the response -- corresponds to `HeaderMap::append`)

The former is less error-prone for most cases, but the latter is more general.  I don't have much preference about which or both.

dropshot/src/handler.rs
```diff
@@ -139,6 +150,21 @@ impl<Context: ServerContext> RequestContext<Context> {
              */
             .unwrap_or(server_config.page_default_nitems))
     }
+
+    /**
+     * Add a header that will be included with the HTTP response.
+     */
+    pub fn add_response_header(
+        self: Arc<Self>,
+        header: http::header::HeaderName,
```
davepacheco, line 159: nit, and *please* feel free to ignore: I see that `HeaderMap::insert()` takes `IntoHeaderName`, and I wonder if that's not more general in a useful way.  We could even expose a `&mut HeaderMap` directly to people and avoid dealing with a bunch of these issues ourselves.

dropshot/src/handler.rs
```diff
@@ -139,6 +150,21 @@ impl<Context: ServerContext> RequestContext<Context> {
...
+        value: http::header::HeaderValue,
```
davepacheco, line 160: What do we want to do if somebody tries to set a header that Dropshot really should own (like maybe "date" or "transfer-encoding")?  I can imagine:

- silently ignore it
- let them clobber it
- panic
- make this function return a `Result`, even though it feels like this is something that could be detectable at compile time, except that I don't see how to do that

I hate all of these.  What do you think?

dropshot/src/handler.rs
```diff
@@ -486,9 +512,18 @@ where
...
+        let response_rqctx = rqctx.clone();
         let funcparams = Extractor::from_request(Arc::clone(&rqctx)).await?;
         let future = self.handler.handle_request(rqctx, funcparams);
-        future.await
+
+        future.await.map(|mut response| {
+            let mutable_state = response_rqctx.mutable_state.lock().unwrap();
+
+            for (header, value) in &mutable_state.response_headers {
+                response.headers_mut().insert(header, value.clone());
```
davepacheco, line 523: As written, it looks like if somebody sets the value multiple times with `add_response_header`, we'll store them all in the Vec but the end result will be to only take the last one.

dropshot/src/test_util.rs
```diff
@@ -62,6 +62,7 @@ const ALLOWED_HEADERS: [AllowedHeader<'static>; 5] = [
         name: "transfer-encoding",
         value: AllowedValue::OneOf(&["chunked"]),
     },
+    AllowedHeader::new("x-dropshot-test-header"),
```
davepacheco, line 65: nit: since this is one we know we use internally, it'd be sad if we had a bug that leaked this header and we didn't detect it in this test.

I guess the header itself is only ever used in the tests so maybe that's not much of a risk.  The alternative seems like it would be to allow each to test to configure the allowable headers, which might be kind of nice, but seems like a lot of work to take on for this case.

dropshot/tests/test_demo.rs
```diff
@@ -606,6 +609,35 @@ async fn test_delete_request() {
...
+            "/testing/header",
+            None as Option<()>,
+            StatusCode::NO_CONTENT,
+        )
+        .await
+        .expect("expected success");
+
+    println!("{:?}", response.headers());
+
+    let header = response
```
davepacheco, line 632: What do you think of adding another header to this test case so that we test what happens if you add the header multiple times?  (That's regardless of what we decide to do in that case.)

dropshot/src/handler.rs
```diff
@@ -486,9 +512,18 @@ where
...
         let rqctx = Arc::new(rqctx_raw);
+        let response_rqctx = rqctx.clone();
         let funcparams = Extractor::from_request(Arc::clone(&rqctx)).await?;
         let future = self.handler.handle_request(rqctx, funcparams);
-        future.await
+
+        future.await.map(|mut response| {
+            let mutable_state = response_rqctx.mutable_state.lock().unwrap();
+
+            for (header, value) in &mutable_state.response_headers {
```
davepacheco, line 522: I just realized: is this too late?  I'm trying to work this out myself and I'm not sure.

Basically, I'm wondering: is it possible that we've already started streaming the response at this point?  What happens if so?

dropshot/src/handler.rs
ahl: That would make it much easier to use. Good idea.

dropshot/src/handler.rs
ahl: I think we let them clobber it. It's either going to let consumers do something ... tricky/sketchy, or be a foot-gun, but it's a foot-gun that would--I think--require careful aiming. In other words, I think it might be a useful escape hatch and is unlikely to cause mysterious errors.

dropshot/src/handler.rs
ahl: That's as I intended. What do you think would be a better interface?

dropshot/src/test_util.rs
ahl: It seemed like more work than was warranted to let each test tack on additional allowed headers, but I let me know what you think. It's probably not **that** much more work.

dropshot/tests/test_demo.rs
ahl: Good idea.

dropshot/src/handler.rs
ahl: How could we be streaming it out? This function returns it here:
https://github.com/oxidecomputer/dropshot/blob/main/dropshot/src/server.rs#L766
I get that it's called  from an `impl Future`
https://github.com/oxidecomputer/dropshot/blob/main/dropshot/src/server.rs#L766
but we're still going to `await` until we have a an actual `Response<Body>` in hand.

dropshot/src/test_util.rs
bnaecker: Ah, thanks I missed that. I'm fine with not returning a Result yet, too. Dave suggested exposing the `HeaderMap` directly, which I think skirts a lot of these questions I had about multiple values for headers, invalid names, etc. That sounds like a good approach at this point.

dropshot/src/handler.rs
bnaecker: I agree with Dave here. Let `HeaderMap` handle issues of conversion, duplicate values, etc.

dropshot/src/handler.rs
davepacheco: I think you resolved this separately by using the HeaderMap.

dropshot/src/test_util.rs
davepacheco: Having the header name be a static in the tests is already a help.

dropshot/src/handler.rs
davepacheco: I think this is fine.

I'd gotten some wires crossed.  I was thinking that if you've got a whole `Response`, then you've already serialized the body, and in a server that supports streaming, once you've serialized the body, it's too late to add headers.  But having a `Response` doesn't mean you've serialized the response body -- the body can be a Stream whose bytes are read later.

ahl:
closing in favor of #283

PR #283, opened 2022-02-22
Introduce structured and unstructured response headers

ahl:
This adds `HttpResponseHeaders<T: HttpTypedResponse, H: JsonSchema + Serialize + Send + Sync + 'static = NoHeaders>` that allows for consumers to specify structured headers (that also appear in the OpenAPI output) and unstructured, *ad-hoc* headers that may be specified at runtime.

Previously, we had been--in places--using `schemars` a little bit wrong in that we were using an intermediate form rather than its fully processed form. In particular this meant that the JSON Schema we looked at didn't always conform to the variant of JSON Schema that conforms most closely with OpenAPI v3.0.x. This delta updates our use of `schemars` which has the upshot of both requiring some additional handling... but also recovering some `description` fields that we had previously discarded unintentionally. This change also pulls descriptions that had been embedded within schemas up to the parameters they describe.


dropshot/tests/test_demo.rs
```diff
@@ -774,6 +814,25 @@ async fn demo_handler_delete(
     Ok(HttpResponseDeleted())
 }
 
+#[endpoint {
+    method = GET,
+    path = "/testing/headers",
+}]
+async fn demo_handler_headers(
+    _rqctx: RequestCtx,
+) -> Result<HttpResponseHeaders<HttpResponseUpdatedNoContent>, HttpError> {
```
davepacheco, line 823: This is cool.  And if you have a body to send back, I imagine this is like `HttpResponseHeaders<HttpResponseOk<SomeType>>`?

dropshot/src/handler.rs
```diff
@@ -1285,6 +1316,170 @@ impl From<HttpResponseUpdatedNoContent> for HttpHandlerResult {
     }
 }
 
+#[derive(Serialize, JsonSchema)]
+pub struct NoHeaders {}
+pub struct HttpResponseHeaders<
```
davepacheco, line 1321: This was probably on your list already but it'd be great to have some user doc comments for this.  (Sorry if I missed them elsewhere!)

dropshot/tests/test_demo.rs
ahl: Exactly right

dropshot/src/handler.rs
ahl: pushed now; sorry that was a local change I made during a self-review that I neglected to push

PR #284, opened 2022-02-22
enable buildomat check suites on push and pull request

jclulow:


davepacheco (approved):
I'm not sure what this does absent any other files but it looks innocuous enough!

jclulow:
It just allows buildomat to listen to this repository.  An additional file in a subsequent PR will be required to describe the job or jobs that will be executed.

PR #286, opened 2022-02-25
Error responses in OpenAPI output

ahl:
This takes the approach of adding an error response for all 4xx and 5xx responses. One could imagine extending this in the future e.g. to include

- additional typed information on a per-consumer basis
- additional typed information per endpoint
- additional untyped information e.g. included just as a permissive schema

### DIFF
--- a/dropshot/src/error.rs
+++ b/dropshot/src/error.rs
@@ -114,6 +114,7 @@ pub struct HttpError {
  */
 #[derive(Debug, Deserialize, Serialize, JsonSchema)]
 #[schemars(rename = "Error")]
+#[schemars(description = "Error information from a response.")]
 pub struct HttpErrorResponseBody {
     pub request_id: String,
     // The combination of default and required removes "nullable" from the
--- a/dropshot/src/error.rs
+++ b/dropshot/src/error.rs
@@ -45,6 +45,7 @@
  */
 
 use hyper::Error as HyperError;
+use schemars::JsonSchema;
 use serde::Deserialize;
 use serde::Serialize;
 use std::error::Error;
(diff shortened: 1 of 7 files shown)


david-crespo:
I'll test this out with the generator.

dropshot/tests/test_openapi_fuller.json
```diff
@@ -414,6 +510,25 @@
           "x"
         ]
       },
+      "Error": {
+        "description": "Error information from a response.",
+        "type": "object",
+        "properties": {
+          "error_code": {
+            "type": "string"
+          },
```
david-crespo, line 519: Is this nullable? I believe there are cases in Omicron where the error code comes back `null`.

https://github.com/oxidecomputer/omicron/blob/53c6dc75dc49365983c032a73761320fbfe7e4f7/nexus/tests/integration_tests/unauthorized.rs#L716
https://github.com/oxidecomputer/omicron/blob/d08038c470370b08b5351d3ea9f35364e15461cb/nexus/tests/integration_tests/authn_http.rs#L79

dropshot/tests/test_openapi_fuller.json
```diff
@@ -414,6 +510,25 @@
...
+          "request_id": {
+            "type": "string"
+          }
+        },
+        "required": [
+          "message",
+          "request_id"
+        ]
```
david-crespo, line 530: 🤦 yes `error_code` can be left out. Never mind. Though actually I have a question — does not required mean it can be given a value of `null`, or only that it can be left out?

david-crespo:
Looks good to me as-is, but I would like an explicit statement how we want `error_code` to work. In Nexus we have the field wrapped in `Option`, and when it's `None`, it gets serialized as `null` (which I don't believe is compatible with the schema here).

```sh
~/oxide/omicron $ curl -sSq http://127.0.0.1:12220/organizations -X DELETE
{
  "request_id": "3684648c-f824-42fe-8596-e46e10435a81",
  "error_code": null,
  "message": "Method Not Allowed"
}
```

dropshot/src/error.rs
```diff
@@ -107,9 +108,18 @@ pub struct HttpError {
...
+ * output?
+ */
+#[derive(Debug, Deserialize, Serialize, JsonSchema)]
+#[schemars(rename = "Error")]
+#[schemars(description = "Error information from a response.")]
 pub struct HttpErrorResponseBody {
     pub request_id: String,
+    // The combination of default and required removes "nullable" from the
+    // OpenAPI-flavored JSON Schema output.
+    #[schemars(default, required)]
```
ahl, line 122: @david-crespo per your comments; while many types by default interpret `Option<T>` as both optional and `nullable` I chose to just have this be optional since it will never be `null` and I find the schema to be more comprehensible without it.

david-crespo: So integrating this change will cause Nexus to stop making it `null` sometimes? Or is there a change also required in Nexus?

ahl: I'll add a test to confirm. In general, I prefer to omit a field rather than having a null field, but I realize that isn't necessarily right or even a broadly held opinion.

PR #293, opened 2022-03-02
Improve error message for endpoints returning non-`Send` futures

jgallagher:
I ran into this last week and had to ask for help diagnosing the error message I was getting from dropshot, so took a stab at improving the error message for this case. There's one wart in the implementation I don't love; I'll call that out in a comment below. For the before/after on the error message itself, see the changes to `bad_endpoint15.stderr` in 082758d530c59babe99f2b9ec0af650f11f7973b.

There may be other bounds we want to check (see https://github.com/oxidecomputer/dropshot/pull/18#pullrequestreview-460974354); adding them should be very straightforward. If this change looks good to folks, I can add those in this or a followup PR.


dropshot_endpoint/src/lib.rs
```diff
@@ -286,6 +290,33 @@ fn do_endpoint(
...
+            let argname = format_ident!("arg{}", i);
+            quote! { #argname }
+        })
+        .collect::<Vec<_>>();
+    let impl_checks = if !arg_is_receiver {
+        quote! {
+            const _: fn() = || {
+                fn future_endpoint_must_be_send<T: ::std::marker::Send>(_t: T) {}
+                fn check_future_bounds(#( #arg_names: #arg_types ),*) {
+                    #item
```
jgallagher, line 309: This is the implementation wart I referred to in the PR comment; it's a full duplicate of the endpoint definition. At first I tried moving this into an `impl` block on the struct named `#name`, but that has adverse effects on the error message case for endpoints taking `self`, so went with this instead. Open to other ideas!

jgallagher: Fixed in [8568fba](https://github.com/oxidecomputer/dropshot/pull/293/commits/8568fba136488bf74d08a8f782b50ba91a022448); thanks @ahl for the suggestion

dropshot/tests/fail/bad_endpoint15.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright 2021 Oxide Computer Company
```
ahl, line 1: update the year here and elsewhere?

dropshot_endpoint/src/lib.rs
```diff
@@ -286,6 +290,32 @@ fn do_endpoint(
         })
         .collect::<Vec<_>>();
 
+    // We want to construct a function that will call the user's endpoint, so
+    // we can check the future it returns for bounds that otherwise produce
+    // inscrutable error messages (like returning a non-`Send` future). We
+    // produce a wrapper function that takes all the same argument types,
+    // which requires building up a list of argument names.
```
ahl, line 297: why can't we use the names from the original function?

dropshot_endpoint/src/lib.rs
```diff
@@ -286,6 +290,32 @@ fn do_endpoint(
...
+        .map(|i| {
+            let argname = format_ident!("arg{}", i);
+            quote! { #argname }
+        })
+        .collect::<Vec<_>>();
+    let impl_checks = if !arg_is_receiver {
+        quote! {
+            const _: fn() = || {
+                fn future_endpoint_must_be_send<T: ::std::marker::Send>(_t: T) {}
+                fn check_future_bounds(#( #arg_names: #arg_types ),*) {
```
ahl, line 308: that's cool; I didn't know `quote!` could do this...

ahl (commented):
looks great!

dropshot_endpoint/src/lib.rs
jgallagher: It will break if they used `_` as one or more argument names. Not likely, but certainly legal.

PR #294, opened 2022-03-02
add clone to bodies

jessfraz:
So I know this is a bit strange, but was wondering if we could add clone to these, thanks!

### DIFF
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -916,7 +916,7 @@ pub(crate) fn schema2struct(
  * that implements `serde::Deserialize`.  See this module's documentation for
  * more information.
  */
-#[derive(Debug)]
+#[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


dropshot/src/handler.rs
```diff
@@ -1001,7 +1001,7 @@ where
  * `UntypedBody` is an extractor for reading in the contents of the HTTP request
  * body and making the raw bytes directly available to the consumer.
  */
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
davepacheco, line 1004: It seems like no problem to put this onto `UntypedBody`.

I'm less sure about `TypedBody<BodyType>` only because I think it means that the caller has to impl `Clone` on their `BodyType`.  I'd be interested in what others think about that.

jessfraz:
Yeah that's fair, I had already done that so I didn't even think about it

jessfraz:
fwiw I can do this a different way in the other repo so I wouldn't worry about it too much!

PR #296, opened 2022-03-08
support for tag descriptions

### LINKED ISSUE #245: tags high level object generation
add the high level tags object where the tag name and description exist, this is not being done currently, related to https://github.com/oxidecomputer/omicron/issues/535

ahl:
fixes #245 

This is largely how @davepacheco suggested in #245. I think there are some shortcomings such as requiring consumers to specify a `TagConfig` before specifying endpoints, but I can't think of a particularly more elegant solution right now.


davepacheco:
This looks great!

What do you think about an example that uses this so we could see it a little more end-to-end?

Also: changelog?

davepacheco:
This looks great!

What do you think about an example that uses this so we could see it a little more end-to-end?

Also: changelog?

davepacheco (approved):

PR #310, opened 2022-03-28
allow consumers to respond with freeform bodies and a particular status code

ahl:
For crucible and the update system, it's useful to be able to respond with a freeform body. Dropshot allows this today, but doing so removes both the expected status code and structural information about errors. This introduces a endpoint response type:

```rust
async fn stream_bytes_at_me_please(
    rqctx: Arc<RequestContext<()>>,
) -> Result<HttpResponseOk<FreeformBody>, HttpError> {
    let (sender, body) = Body::channel();
    // spawn a task to spew bytes into `sender`
    Ok(HttpResponseOk(body.into()))
}
```

Note this involves some reshuffling of types--I think--for the better. For example, `Empty` no longer `impl`s `Serialize` or `JsonSchema` and the numeric response code trait is separated from the content trait.


jessfraz:
oh cool!

PR #315, opened 2022-04-06
make HttpTypedResponse pub use

jessfraz:
I am trying to refactor some code and clean up some stuff and making this public would allow me to pass in a generic places that must implement this trait, such that then I can pass HttpResponseOk{Etc} around as a generic


ahl (commented):
looks good to me!

PR #317, opened 2022-04-07
Allow use of usdt probes with stable rust

pfmooney:
The `usdt` crate can be built with stable toolchains as of Rust v1.59,
except on MacOS, where the still-unstable `asm_sym` feature is needed.
Dropshot should do proper compiler version detection so it too can be
built with the stable toolchain when `usdt` is enabled.

### DIFF
--- a/dropshot/src/lib.rs
+++ b/dropshot/src/lib.rs
@@ -507,9 +507,11 @@
  * See the [`RequestInfo`] and [`ResponseInfo`] types for a complete listing
  * of what's available.
  *
- * These probes are implemented via the [`usdt`] crate, and require a nightly
- * toolchain. As such, they're behind the feature flag `"usdt-probes"`. You
- * can build Dropshot with these probes via `cargo +nightly build --features usdt-probes`.
+ * These probes are implemented via the [`usdt`] crate. They require a nightly
+ * toolchain if built on MacOS (which requires the unstable `asm_sym` feature).
+ * Otherwise a stable compiler >= v1.59 is required in order to present the
+ * necessary features.  Given these constraints, usdt functionality is behind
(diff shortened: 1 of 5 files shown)


dropshot/src/lib.rs
```diff
@@ -507,9 +507,11 @@
  * See the [`RequestInfo`] and [`ResponseInfo`] types for a complete listing
  * of what's available.
  *
- * These probes are implemented via the [`usdt`] crate, and require a nightly
- * toolchain. As such, they're behind the feature flag `"usdt-probes"`. You
- * can build Dropshot with these probes via `cargo +nightly build --features usdt-probes`.
+ * These probes are implemented via the [`usdt`] crate. They require a nightly
+ * toolchain if built on MacOS (which requires the unstable `asm_sym` feature).
+ * Otherwise a stable compiler <= v1.59 is required in order to present the
```
bnaecker, line 512: I believe you meant "a stable compiler >= 1.59 or a nightly compiler prior to that version". Something along those lines?

bnaecker (approved):

dropshot/src/lib.rs
pfmooney: Yes, silly me.  I'll fix this shortly.

ahl (approved):
looks great. thanks!

davepacheco:
Nice!  Is there any impact for the changelog?  What does this mean for dropshot consumers?

bnaecker:
Dropshot consumers should now be able to get USDT probes building on a stable compiler, as long as that's version >= 1.59, and they're not on macOS. The macOS implementation of `usdt` still relies on a nightly feature.

pfmooney:
Added an update to the changelog to mention this.

PR #327, opened 2022-04-15
file server example: more verbose errors, note TOCTOU

davepacheco:
I replaced the errno strings with more verbose errors.  Examples:

```
$ curl 127.0.0.1:51343/foo
{
  "request_id": "a088701b-b77e-44dc-937c-b77be9ec76c0",
  "message": "failed to access \"./foo\": No such file or directory (os error 2)"
}

$ curl 127.0.0.1:51343/bar
{
  "request_id": "1698473f-b389-4769-a3c1-900e8952a68f",
  "message": "attempted to follow symlink \"./bar\""
}

$ curl 127.0.0.1:51343/bonk
{
  "request_id": "1be60353-1459-4aca-ac65-4f9d3092d2f3",
  "message": "failed to access \"./bonk\": No such file or directory (os error 2)"
}

$ curl 127.0.0.1:51343/a/b/c
{
  "request_id": "e9d04c0c-b6e0-44d7-98bf-8058fae4f009",
  "message": "failed to access \"./a\": No such file or directory (os error 2)"
}
$ touch foo
$ curl 127.0.0.1:51343/foo/a/b
{
  "request_id": "3a63cfa6-69dd-4041-bfc7-62c221e5580b",
  "message": "expected directory: \"./foo\""
}
```

I also noticed there's kind of a big TOCTOU race in the way we check symlinks.  I think this is usually fixed with the `openat` family of functions.  I'm not sure if Rust exposes those.  Anyway, I didn't want to take that on right now so I've noted it for readers...but it does make me a little uneasy to have this in the example.

### DIFF
--- a/dropshot/examples/file_server.rs
+++ b/dropshot/examples/file_server.rs
@@ -1,5 +1,7 @@
 // Copyright 2021 Oxide Computer Company
 
+//! Example using Dropshot to serve files
+
 use dropshot::ApiDescription;
 use dropshot::ConfigLogging;
 use dropshot::ConfigLoggingLevel;
@@ -102,7 +104,7 @@ async fn static_content(
         if !entry.is_dir() {
             return Err(HttpError::for_bad_request(
                 None,
-                "ENOTDIR".to_string(),
+                format!("expected directory: {:?}", entry),
             ));
         }
         /* Dropshot won't ever give us dot-components. */
@@ -113,12 +115,21 @@ async fn static_content(
         /*
          * We explicitly prohibit consumers from following symlinks to prevent
          * showing data outside of the intended directory.
+         * TODO-security There's a time-of-check to time-of-use race here!
+         * Someone could replace "entry" with a symlink immediately after we
+         * check.
          */
-        let m = entry.symlink_metadata().map_err(|_| {
-            HttpError::for_bad_request(None, "ENOENT".to_string())
+        let m = entry.symlink_metadata().map_err(|e| {
+            HttpError::for_bad_request(
+                None,
+                format!("failed to access {:?}: {:#}", entry, e),
+            )
         })?;
         if m.file_type().is_symlink() {
-            return Err(HttpError::for_bad_request(None, "EMLINK".to_string()));
+            return Err(HttpError::for_bad_request(
+                None,
+                format!("attempted to follow symlink {:?}", entry),
+            ));
         }
     }
 
@@ -127,8 +138,11 @@ async fn static_content(
      * regular file we serve the file.
      */
     if entry.is_dir() {
-        let body = dir_body(entry).await.map_err(|_| {
-            HttpError::for_bad_request(None, "EBADF".to_string())
+        let body = dir_body(&entry).await.map_err(|e| {
+            HttpError::for_bad_request(
+                None,
+                format!("failed to read dir {:?}: {:#}", entry, e),
+            )
         })?;
 
         Ok(Response::builder()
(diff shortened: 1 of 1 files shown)


jclulow:
It would seem the requisite function is of course not yet stable: https://doc.rust-lang.org/std/os/wasi/fs/trait.OpenOptionsExt.html#tymethod.open_at

smklein:
We could use https://doc.rust-lang.org/std/os/unix/fs/trait.OpenOptionsExt.html#tymethod.custom_flags with O_NOFOLLOW, no?

davepacheco:
> We could use https://doc.rust-lang.org/std/os/unix/fs/trait.OpenOptionsExt.html#tymethod.custom_flags with O_NOFOLLOW, no?

There was an extensive discussion on various options in the oxide-rust chat channel.

No, `open(2)` with `O_NOFOLLOW` still follows symlinks in some cases. My read of the manual page is that it only _doesn't_ follow a link for the very last component of the path.  I did not verify this, but @rmustacc checked and verified that it's not suitable for this use.

Both the `libc` and `nix` crates appear to provide the `openat`/`fstat` functions that I believe are typically used to do the traversal safely.  The [openat](https://docs.rs/openat/latest/openat/) crate might too?  But the interface is surprisingly different and it's not clear to me how you'd actually do the traversal safely.

rmustacc:
> No, `open(2)` with `O_NOFOLLOW` still follows symlinks in some cases. My read of the manual page is that it only _doesn't_ follow a link for the very last component of the path. I did not verify this, but @rmustacc checked and verified that it's not suitable for this use.

Yeah, specifically if you look at the documentation on different POSIX systems, they specifically state that `O_NOFOLLOW` basically only applies to the final thing you are calling open on as opposed to the intermediate path. Many lookup paths will actually OR in and continue to actually follow symlinks in the intermediate path.

smklein:
> > We could use https://doc.rust-lang.org/std/os/unix/fs/trait.OpenOptionsExt.html#tymethod.custom_flags with O_NOFOLLOW, no?
> 
> There was an extensive discussion on various options in the oxide-rust chat channel.
> 
> No, `open(2)` with `O_NOFOLLOW` still follows symlinks in some cases. My read of the manual page is that it only _doesn't_ follow a link for the very last component of the path. I did not verify this, but @rmustacc checked and verified that it's not suitable for this use.
> 
> Both the `libc` and `nix` crates appear to provide the `openat`/`fstat` functions that I believe are typically used to do the traversal safely. The [openat](https://docs.rs/openat/latest/openat/) crate might too? But the interface is surprisingly different and it's not clear to me how you'd actually do the traversal safely.

I figured that if we opened the entries one at a time this would be workable - but I've also used BurntSushi's walkdir crate before too, which includes a toggle to [follow_links](https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html#method.follow_links) or not.

rmustacc:
> I figured that if we opened the entries one at a time this would be workable - but I've also used BurntSushi's walkdir crate before too, which includes a toggle to [follow_links](https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html#method.follow_links) or not.

It will still follow links at the root, but presumably that's ok. However, from the source it's not clear to me if it's actually doing an fstat on the underlying open file descriptor and using an openat relative to that or not. Though the crate's design, I thought, was for actually recursively walking an entire directory structure and all files recursively contained therin as opposed to simply traversing a single path.

ahl (commented):
Nice; thanks

davepacheco:
> I figured that if we opened the entries one at a time this would be workable - but I've also used BurntSushi's walkdir crate before too, which includes a toggle to [follow_links](https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html#method.follow_links) or not.

How do you use `walkdir` to safely look up a specific path?  I thought that was more for enumerating a directory tree's contents (like `ftw()`).

PR #328, opened 2022-04-15
#310 was a breaking change

davepacheco:


bnaecker (approved):

ahl:
An alternative approach would be to undo the breakage and rename the function back to `metadata` -- I was obviously careless when I decided to alleviate my confusion about all these functions I myopically named `metadata`.

PR #343, opened 2022-05-11
honor default values in JSON Schema and include it in the OpenAPI output

ahl:
Needed to address https://github.com/oxidecomputer/progenitor/issues/41


davepacheco:
When I released v0.7.0, I neglected to push the `cargo release` commits that update the changelog and version numbers.  This PR was the only change that landed before I realized this.  I force-pushed those commits and the tags so that the git history would accurately reflect what was released, and I cherry-picked this commit onto the tip of "main".  I hope this has minimal impact on people but let me know if you find issues.

PR #362, opened 2022-05-31
build with buildomat

ahl:


davepacheco:
Thanks for doing this!  This obviates #87.  There are a few changes there that seem useful here (like `--locked` and the `cargo version` output).

ahl:
closing in favor of #87

Issue #344, opened 2022-05-12
when you add custom headers to an endpoint it should propogate to OPTIONS


jessfraz:

ahl:
Makes sense. I’ll look into it. What service is this for?

jessfraz:
just playing around w some stuff for cio, I likely can find the spot its not urgent but this is to remind me

PR #370, opened 2022-06-08
allow for options method

jessfraz:
this is a method of fixing https://github.com/oxidecomputer/dropshot/issues/344 in that the user has to define options on endpoints

it would be nicer if somehow these were generated but since response headers can be defined programmatically on the fly I can't think of a way to generate them, so rather this turns on the ability to add OPTIONS as a method


ahl (commented):
looks good to me

PR #390, opened 2022-06-28
Improve errors when failing to load certificates

smklein:
The current implementation drops the underlying error. I'd like to see it!


augustuswm (approved):
Looks good to me.

PR #403, opened 2022-07-28
Implement websocket support in Dropshot as an Extractor and #[channel] macro

lifning:
This allows endpoints meant for handling continuous websocket streams to be easily constructed by providing an async handler to an extractor parameter in the endpoint function definition.

This change also refactors the previous `paginated` bool to instead be a dedicated type describing whether an endpoint uses pagination or websockets Dropshot-specific extensions, aspirationally leaving room for describing any more extension modes we may add atop the basic OpenAPI.

### DIFF
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -27,6 +27,7 @@ rustls = "0.20.6"
 rustls-pemfile = "1.0.1"
 serde_json = "1.0.83"
 serde_urlencoded = "0.7.1"
+sha1 = "0.10.1"
 slog-async = "2.4.0"
 slog-bunyan = "2.4.0"
 slog-json = "2.6.1"
@@ -84,6 +85,10 @@ trybuild = "1.0.64"
 # Used by the https examples and tests
 pem = "1.1"
 rcgen = "0.9.3"
+# Used in a doc-test demonstrating the WebsocketUpgrade extractor.
+tungstenite = "0.17.3"
+tokio-tungstenite = "0.17.2"
+futures-util = "0.3.21"
 
 [dev-dependencies.rustls]
 version = "0.20.6"
--- a/dropshot/examples/websocket.rs
+++ b/dropshot/examples/websocket.rs
@@ -1,0 +1,95 @@
+// Copyright 2022 Oxide Computer Company
+/*!
+ * Example use of Dropshot with a websocket endpoint.
+ */
+
+use dropshot::channel;
+use dropshot::ApiDescription;
+use dropshot::ConfigDropshot;
+use dropshot::ConfigLogging;
+use dropshot::ConfigLoggingLevel;
+use dropshot::HttpServerStarter;
+use dropshot::Query;
+use dropshot::RequestContext;
+use dropshot::WebsocketConnection;
+use futures_util::SinkExt;
+use schemars::JsonSchema;
+use serde::Deserialize;
+use std::sync::Arc;
+use tungstenite::protocol::Role;
+use tungstenite::Message;
+
+#[tokio::main]
+async fn main() -> Result<(), String> {
+    /*
+     * We must specify a configuration with a bind address.  We'll use 127.0.0.1
+     * since it's available and won't expose this server outside the host.  We
+     * request port 0, which allows the operating system to pick any available
+     * port.
+     */
+    let config_dropshot: ConfigDropshot = Default::default();
+
+    /*
+     * For simplicity, we'll configure an "info"-level logger that writes to
+     * stderr assuming that it's a terminal.
+     */
+    let config_logging =
+        ConfigLogging::StderrTerminal { level: ConfigLoggingLevel::Info };
+    let log = config_logging
+        .to_logger("example-basic")
+        .map_err(|error| format!("failed to create logger: {}", error))?;
+
+    /*
+     * Build a description of the API.
+     */
+    let mut api = ApiDescription::new();
+    api.register(example_api_websocket_counter).unwrap();
+
+    /*
+     * Set up the server.
+     */
+    let server = HttpServerStarter::new(&config_dropshot, api, (), &log)
+        .map_err(|error| format!("failed to create server: {}", error))?
+        .start();
+
+    /*
+     * Wait for the server to stop.  Note that there's not any code to shut down
+     * this server, so we should never get past this point.
+     */
+    server.await
+}
+
+/*
+ * HTTP API interface
+ */
+
+#[derive(Deserialize, JsonSchema)]
+struct QueryParams {
+    start: Option<u8>,
+}
+
+/**
+ * An eternally-increasing sequence of bytes, wrapping on overflow, starting
+ * from the value given for the query parameter "start."
+ */
+#[channel {
+    protocol = WEBSOCKETS,
+    path = "/counter",
+}]
+async fn example_api_websocket_counter(
+    _rqctx: Arc<RequestContext<()>>,
+    upgraded: WebsocketConnection,
+    qp: Query<QueryParams>,
+) -> dropshot::WebsocketChannelResult {
+    let mut ws = tokio_tungstenite::WebSocketStream::from_raw_socket(
+        upgraded.into_inner(),
+        Role::Server,
+        None,
+    )
+    .await;
+    let mut count = qp.into_inner().start.unwrap_or(0);
+    while ws.send(Message::Binary(vec![count])).await.is_ok() {
+        count = count.wrapping_add(1);
+    }
+    Ok(())
+}
--- a/dropshot/src/api_description.rs
+++ b/dropshot/src/api_description.rs
@@ -45,7 +45,7 @@ pub struct ApiEndpoint<Context: ServerContext> {
     pub summary: Option<String>,
     pub description: Option<String>,
     pub tags: Vec<String>,
-    pub paginated: bool,
+    pub extension_mode: ExtensionMode,
     pub visible: bool,
 }
 
@@ -78,7 +78,7 @@ impl<'a, Context: ServerContext> ApiEndpoint<Context> {
             summary: None,
             description: None,
             tags: vec![],
-            paginated: func_parameters.paginated,
+            extension_mode: func_parameters.extension_mode,
             visible: true,
         }
     }
@@ -688,11 +688,20 @@ impl<Context: ServerContext> ApiDescription<Context> {
                 })
                 .next();
 
-            if endpoint.paginated {
-                operation.extensions.insert(
-                    crate::pagination::PAGINATION_EXTENSION.to_string(),
-                    serde_json::json! {true},
-                );
+            match endpoint.extension_mode {
+                ExtensionMode::None => {}
+                ExtensionMode::Paginated => {
+                    operation.extensions.insert(
+                        crate::pagination::PAGINATION_EXTENSION.to_string(),
+                        serde_json::json! {true},
+                    );
+                }
+                ExtensionMode::Websocket => {
+                    operation.extensions.insert(
+                        crate::websocket::WEBSOCKET_EXTENSION.to_string(),
+                        serde_json::json!({}),
+                    );
+                }
             }
 
             let response = if let Some(schema) = &endpoint.response.schema {
@@ -1577,6 +1586,22 @@ pub struct TagDetails {
 pub struct TagExternalDocs {
     pub description: Option<String>,
     pub url: String,
+}
+
+/**
+ * Dropshot/Progenitor features used by endpoints which are not a part of the base OpenAPI spec.
+ */
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum ExtensionMode {
+    None,
+    Paginated,
+    Websocket,
+}
+
+impl Default for ExtensionMode {
+    fn default() -> Self {
+        ExtensionMode::None
+    }
 }
 
 #[cfg(test)]
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -40,16 +40,17 @@ use super::http_util::CONTENT_TYPE_JSON;
 use super::http_util::CONTENT_TYPE_OCTET_STREAM;
 use super::server::DropshotState;
 use super::server::ServerContext;
-use crate::api_description::ApiEndpointBodyContentType;
 use crate::api_description::ApiEndpointHeader;
 use crate::api_description::ApiEndpointParameter;
 use crate::api_description::ApiEndpointParameterLocation;
 use crate::api_description::ApiEndpointResponse;
 use crate::api_description::ApiSchemaGenerator;
+use crate::api_description::{ApiEndpointBodyContentType, ExtensionMode};
 use crate::pagination::PaginationParams;
 use crate::pagination::PAGINATION_PARAM_SENTINEL;
 use crate::router::VariableSet;
 use crate::to_map::to_map;
+use crate::websocket::WEBSOCKET_PARAM_SENTINEL;
 
 use async_trait::async_trait;
 use bytes::Bytes;
@@ -173,8 +174,8 @@ impl<T: 'static + ServerContext> RequestContextArgument
  * `RequestContext`.  Unlike most traits, `Extractor` essentially defines only a
  * constructor function, not instance functions.
  *
- * The extractors that we provide (`Query`, `Path`, `TypedBody`, and
- * `UntypedBody`) implement `Extractor` in order to construct themselves from
+ * The extractors that we provide (`Query`, `Path`, `TypedBody`, `UntypedBody`, and
+ * `WebsocketUpgrade`) implement `Extractor` in order to construct themselves from
  * the request. For example, `Extractor` is implemented for `Query<Q>` with a
  * function that reads the query string from the request, parses it, and
  * constructs a `Query<Q>` with it.
@@ -202,7 +203,7 @@ pub trait Extractor: Send + Sync + Sized {
  * the associated endpoint is paginated.
  */
 pub struct ExtractorMetadata {
-    pub paginated: bool,
+    pub extension_mode: ExtensionMode,
     pub parameters: Vec<ApiEndpointParameter>,
 }
 
@@ -223,15 +224,21 @@ macro_rules! impl_extractor_for_tuple {
 
         fn metadata(_body_content_type: ApiEndpointBodyContentType) -> ExtractorMetadata {
             #[allow(unused_mut)]
-            let mut paginated = false;
+            let mut extension_mode = ExtensionMode::None;
             #[allow(unused_mut)]
             let mut parameters = vec![];
             $(
                 let mut metadata = $T::metadata(_body_content_type.clone());
-                paginated = paginated | metadata.paginated;
+                extension_mode = match (extension_mode, metadata.extension_mode) {
+                    (ExtensionMode::None, x) | (x, ExtensionMode::None) => x,
+                    (x, y) if x != y => {
+                        panic!("incompatible extension modes in tuple: {:?} != {:?}", x, y);
+                    }
+                    (_, x) => x,
+                };
                 parameters.append(&mut metadata.parameters);
             )*
-            ExtractorMetadata { paginated, parameters }
+            ExtractorMetadata { extension_mode, parameters }
         }
     }
 }}
@@ -685,11 +692,23 @@ where
     );
     let schema = generator.root_schema_for::<ParamType>().schema.into();
 
-    let paginated = match schema_extensions(&schema) {
+    let extension_mode = match schema_extensions(&schema) {
         Some(extensions) => {
-            extensions.get(&PAGINATION_PARAM_SENTINEL.to_string()).is_some()
+            let paginated = extensions
+                .get(&PAGINATION_PARAM_SENTINEL.to_string())
+                .is_some();
+            let websocket =
+                extensions.get(&WEBSOCKET_PARAM_SENTINEL.to_string()).is_some();
+            match (paginated, websocket) {
+                (false, false) => ExtensionMode::None,
+                (false, true) => ExtensionMode::Websocket,
+                (true, false) => ExtensionMode::Paginated,
+                (true, true) => panic!(
+                    "Cannot use websocket and pagination in the same endpoint!"
+                ),
+            }
         }
-        None => false,
+        None => ExtensionMode::None,
     };
 
     /*
@@ -716,7 +735,7 @@ where
         })
         .collect::<Vec<_>>();
 
-    ExtractorMetadata { paginated, parameters }
+    ExtractorMetadata { extension_mode, parameters }
 }
 
 fn schema_extensions(
@@ -1039,7 +1058,10 @@ where
             },
             vec![],
         );
-        ExtractorMetadata { paginated: false, parameters: vec![body] }
+        ExtractorMetadata {
+            extension_mode: ExtensionMode::None,
+            parameters: vec![body],
+        }
     }
 }
 
@@ -1116,7 +1138,7 @@ impl Extractor for UntypedBody {
                 },
                 vec![],
             )],
-            paginated: false,
+            extension_mode: ExtensionMode::None,
         }
     }
 }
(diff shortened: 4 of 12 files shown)


dropshot/src/api_description.rs
```diff
@@ -78,7 +78,7 @@ impl<'a, Context: ServerContext> ApiEndpoint<Context> {
             summary: None,
             description: None,
             tags: vec![],
-            paginated: func_parameters.paginated,
+            extension_mode: func_parameters.extension_mode,
```
david-crespo, line 81: So this means a given endpoint can only be in one mode at a time, right? That makes sense for pagination and websockets, since they're mutually exclusive. I have no idea what other kinds of extensions we might dream up, but is it possible we could want to use more than one at a time?

dropshot/examples/websocket.rs
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * An eternally-increasing sequence of bytes, wrapping on overflow.
+ */
+#[endpoint {
+    method = GET,
+    path = "/counter",
+}]
+async fn example_api_websocket_counter(
+    _rqctx: Arc<RequestContext<()>>,
+    ws_upg: WebsocketUpgrade,
```
david-crespo, line 74: amazing API for this

dropshot/src/api_description.rs
lifning: yeah - i suppose if such a thing comes to pass, we could change to be something more sophisticated than a bare enum, but i didn't want to try to anticipate unknown unknowns yet, just represent that pagination and websockets are mutually exclusive extensions

dropshot/Cargo.toml
```diff
@@ -33,6 +33,7 @@ slog-json = "2.6.1"
 slog-term = "2.9.0"
 tokio-rustls = "0.23.4"
 toml = "0.5.9"
+tungstenite = "0.17.3"
```
luqmana, line 36: A lil unfortunate we have to pull tungstenite + all its deps just for one function (`derive_accept_key`). It's small enough and not really subject to change that I almost want to say we should just pull it in rather than the whole crate.

luqmana (commented):
I left a couple small comments but otherwise looks good, thanks @lifning!

dropshot/src/websocket.rs
```diff
@@ -0,0 +1,336 @@
...
+            None => Err(HttpError::for_internal_error(
+                "spawned websocket handler twice".to_string(),
+            )),
+            Some(WebsocketUpgradeInner {
+                upgrade_fut,
+                accept_key,
+                route,
+                ws_log,
+            }) => {
+                tokio::spawn(async move {
```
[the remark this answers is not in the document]
lifning: Sure! I think I'm more inclined to just make a struct for it to make it a little more ergonomic/readable/explicit than a plain tuple and having people remember to `.0` and `.1`

dropshot/src/api_description.rs
```diff
@@ -1579,6 +1588,41 @@ pub struct TagExternalDocs {
...
+ * Dropshot/Progenitor features used by endpoints which are not a part of the base OpenAPI spec.
+ */
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum ExtensionMode {
+    None,
+    Paginated,
+    Websocket,
+}
+
+impl Default for ExtensionMode {
```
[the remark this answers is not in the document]
lifning: Ahh, I wrote this before the rust-toolchain.toml bump to 1.62 and didn't think to revisit it!

dropshot/src/websocket.rs
```diff
@@ -0,0 +1,411 @@
...
+            })?;
+
+        let method = request.method().as_str().to_string();
+        let route = request.uri().to_string();
+        let upgrade_fut = upgrade::on(request);
+        let ws_log = rqctx.log.new(o!(
+            "upgrade" => "websocket".to_string(),
+            "request_id" => rqctx.request_id.to_string(),
+            "method" => method,
+            "uri" => route.clone(),
```
luqmana, line 174: These should be already be inherited from `rqctx` I thought. +1 on the upgrade one tho, useful to differentiate which logs are from what component.

lifning: gotcha -- also fwiw in case it's not clear, this one is just used by the `error!` calls in the wrapper around the user-provided future in `WebsocketUpgrade::handle`; if the user needs to log in their future, they can obtain it from rqctx the same way they do in any other endpoint & differentiate it however they wish

luqmana: ah, you're right. that makes sense!

luqmana (approved):

dropshot/src/websocket.rs
```diff
@@ -0,0 +1,410 @@
...
+pub use hyper::upgrade::Upgraded as HyperUpgraded;
+
+/**
+ * WebsocketUpgrade is an Extractor used to upgrade and handle an HTTP request
+ * as a websocket when present in a Dropshot endpoint's function arguments.
+ *
+ * The consumer of this *must* call [WebsocketUpgrade::handle] (or if they do
+ * not wish to upgrade a given connection, [WebsocketUpgrade::do_not_upgrade])
+ * on the owned instance in the endpoint's fn, or else WebsocketUpgrade's
+ * implementation of [Drop::drop] will panic.
```
davepacheco, line 38: Why's that?  (why not implicitly call `do_not_upgrade()` on drop, say?)

It looks like it's a non-fatal error to handle the upgrade more than once (we produce an error) but it's fatal to not handle it at all.  I don't have a strong idea here but I just wonder if there's a way to make it harder to accidentally panic in production.  (One thought I had was that `WebsocketUpgrade` could be parametrized by an impl of some trait that would allow us to immediately and automatically invoke it exactly once for you...but then I'm not sure what the body of the function would do.  Also, I'm not sure if it's desirable to be able to do stuff or look at the request before the upgrade -- presumably it is (e.g., authentication).  Then I thought this could be a different kind of endpoint...but then I worry I'm overthinking it.)

dropshot/src/websocket.rs
```diff
@@ -0,0 +1,410 @@
...
+ */
+#[derive(Debug)]
+pub struct WebsocketUpgrade(Option<WebsocketUpgradeInner>);
+
+/**
```
davepacheco, line 43: Did you mean to put some docs here and on `WebsocketUpgradeResponse`?

davepacheco (commented):
Cool! I had a few questions.

Would you mind updating the changelog to describe the new feature?

dropshot/src/websocket.rs
lifning: for the interface to be more discoverable (because `.handle()` *has* to be called for any of this to work; this way nobody's wondering why their connection isn't getting upgraded, even if they weren't originally clear on the docs) and less prone to incorrect use (e.g. accidental inclusion of a WebsocketUpgrade in a param list due to copy/paste errors, causing us to eventually generate an confusing/incorrect client from the resulting schema)

if we're very concerned about panics making it into production, we can make it be an error log in Drop instead, but that's less likely to be caught at development time -- maybe panic if debug assertions are on, else log?

dropshot/src/websocket.rs
lifning: :woman_facepalming: sure did! thanks

dropshot/src/handler.rs
```diff
@@ -223,15 +224,21 @@ macro_rules! impl_extractor_for_tuple {
...
+            let mut extension_mode = ExtensionMode::None;
             #[allow(unused_mut)]
             let mut parameters = vec![];
             $(
                 let mut metadata = $T::metadata(_body_content_type.clone());
-                paginated = paginated | metadata.paginated;
+                extension_mode = match (extension_mode, metadata.extension_mode) {
+                    (ExtensionMode::None, x) | (x, ExtensionMode::None) => x,
+                    (x, y) if x != y => {
+                        panic!("incompatible extension modes in tuple: {:?} != {:?}", x, y);
```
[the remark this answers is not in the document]
lifning: it means the two tuple elements were indicators of different non-None extension modes (i.e. Paginated and Websocket at this point), in either order, which we currently take to be mutually exclusive things

ahl (commented):
Are other extractors permitted alongside `WebsocketUpgrade`? Or must it be the lone extractor?

What does this look like in the OpenAPI output? The OpenAPI spec folks have pretty much declared Websockets as "not HTTP APIs" and therefore out of scope.

I vaguely had intended to use [AsyncAPI](https://www.asyncapi.com/) to describe websocket interfaces, both the channels and the messages. AsyncAPI and OpenAPI are at least sort of compatible in that I think we could produce a single JSON file with both descriptions in there.

Does it make sense to use the same `endpoint` macro with the constraints that the method must be a `GET`, the `WebsocketUpgrade` must be present, and the return type is `Result<Response<Body>, HttpError>`? Or might we want some other interface such as `#[channel { path = ... } ]` or `#[endpoint { method = WEBSOCKET, ...`?

dropshot/src/lib.rs
```diff
@@ -664,6 +665,11 @@ pub use pagination::ResultsPage;
 pub use pagination::WhichPage;
 pub use server::ServerContext;
 pub use server::{HttpServer, HttpServerStarter};
+pub use websocket::HyperUpgraded;
+pub use websocket::SchemaWebsocketParams;
```
ahl, line 669: Why is this pub?

dropshot/src/websocket.rs
```diff
@@ -0,0 +1,410 @@
...
+        }
+    }
+}
+
+// To indicate websocket usage by the endpoint to code generators (i.e. Progenitor)
+pub(crate) const WEBSOCKET_EXTENSION: &str = "x-dropshot-websocket";
+pub(crate) const WEBSOCKET_PARAM_SENTINEL: &str = "x-dropshot-websocket-param";
+
+#[derive(JsonSchema)]
+pub struct SchemaWebsocketParams;
```
ahl, line 301: If this is just here for the `impl JsonSchema for WebsocketUpgrade` we can probably just inline it.

dropshot/src/websocket.rs
```diff
@@ -0,0 +1,410 @@
...
+        "WebsocketUpgrade".to_string()
+    }
+
+    fn json_schema(
+        gen: &mut schemars::gen::SchemaGenerator,
+    ) -> schemars::schema::Schema {
+        let mut schema = SchemaWebsocketParams::json_schema(gen).into_object();
+        schema
+            .extensions
+            .insert(WEBSOCKET_PARAM_SENTINEL.to_string(), json!(true));
```
ahl, line 314: is the schema really just to communicate this sentinel through?

dropshot/src/websocket.rs
```diff
@@ -0,0 +1,410 @@
...
+
+impl JsonSchema for WebsocketUpgrade {
+    fn schema_name() -> String {
+        "WebsocketUpgrade".to_string()
+    }
+
+    fn json_schema(
+        gen: &mut schemars::gen::SchemaGenerator,
+    ) -> schemars::schema::Schema {
+        let mut schema = SchemaWebsocketParams::json_schema(gen).into_object();
```
ahl, line 311: I don't think this schema is ever really used in which case I think we could just replace this with
```suggestion
        let mut schema = schemars::schema::SchemaObject::default();
```

dropshot/src/api_description.rs
```diff
@@ -1579,6 +1588,17 @@ pub struct TagExternalDocs {
     pub url: String,
 }
 
+/**
+ * Dropshot/Progenitor features used by endpoints which are not a part of the base OpenAPI spec.
+ */
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
+pub enum ExtensionMode {
+    #[default]
```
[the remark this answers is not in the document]
lifning: i actually changed to this from a proper bare `impl Default` after an earlier PR comment pointed out that the rust-toolchain.toml for this project declares rust 1.62 - simple enough to revert to what i originally had if it's a concern at all

lifning:
> Are other extractors permitted alongside `WebsocketUpgrade`? Or must it be the lone extractor?

They shouldn't be mutually exclusive (or else we wouldn't be able to do much useful with it, like say addressing a specific org/proj/inst in nexus by `Path<>` params)

> What does this look like in the OpenAPI output? The OpenAPI spec folks have pretty much declared Websockets as "not HTTP APIs" and therefore out of scope.

It looks like a GET endpoint with a `"x-dropshot-websocket": true` stashed in it, just like the `"x-dropshot-pagination": true` we already have. Here's an example (with the unrelated formatting condensed a little for brevity)
```json
{
  "/organizations/{organization_name}/projects/{project_name}/instances/{instance_name}/serial-console": {
    "get": {
      "tags": ["instances"],
      "summary": "Connect to an instance's serial console stream",
      "operationId": "instance_serial_console",
      "parameters": [
        { "in": "path", "name": "instance_name", "required": true,
          "schema": { "$ref": "#/components/schemas/Name" }, "style": "simple" },
        { "in": "path", "name": "organization_name", "required": true,
          "schema": { "$ref": "#/components/schemas/Name" }, "style": "simple" },
        { "in": "path", "name": "project_name", "required": true,
          "schema": { "$ref": "#/components/schemas/Name" }, "style": "simple" }
      ],
      "responses": { "default": { "description": "", "content": { "*/*": {"schema": {}}}}},

      "x-dropshot-websocket": true
    }
  }
}
```

> I vaguely had intended to use [AsyncAPI](https://www.asyncapi.com/) to describe websocket interfaces, both the channels and the messages. AsyncAPI and OpenAPI are at least sort of compatible in that I think we could produce a single JSON file with both descriptions in there.

Would it be reasonable (in the interest of providing a websockets interface in the nearer term for serial & VNC API work) to punt on the full extent of that for the moment with the "x-dropshot-websocket" sentinel in an OpenAPI path's "get" method, and then follow up by implementing the proper AsyncAPI version?

For the use cases I'm thinking of, AsyncAPI's specifications are a bit heavier weight than we need, but that might be a blessing in this regard - I think we often want to forward *other* protocols through raw binary messages, so most of the work that I think this would end up unblocking would have a fairly trivial expression as far as just annotating everywhere it's used with an effective "yep, it's binary" in such a way that our macro could accept before having a "real" AsyncAPI implementation backing it.

> Does it make sense to use the same `endpoint` macro with the constraints that the method must be a `GET`, the `WebsocketUpgrade` must be present, and the return type is `Result<Response<Body>, HttpError>`? Or might we want some other interface such as `#[channel { path = ... } ]` or `#[endpoint { method = WEBSOCKET, ...`?

I did have the thought of perhaps making `endpoint { method = WEBSOCKET, ...` decorating an async fn in which you're required to put an upgraded connection parameter (much like requiring path variables), and not have an intermediate function at all... but I shyed away from the former part of that approach at first because I didn't want to use a non-HTTP-verb there, and now the latter part too because would mean we wouldn't have a good way to expose the JoinHandle of what will likely be a longer-running future than most other kinds of endpoint.

I also had the use case of "what if we want to leave a path open for hybrid GET endpoints requests that may or may not be upgraded depending on how it's accessed," but I am not sure this use case would be worth the additional complexity.

`#[channel...` would be a nice way to both get a little closer to AsyncAPI terminology and possibly isolate some of the additional websocket/AsyncAPI-specific complexity from the `endpoint` decorator's implementation when it comes to things like describing message formats. My current inclination in light of this is to perhaps make such a thing as a declarative macro for now, that expands to the currently-equivalent `#[endpoint = GET` decorator.

dropshot/src/api_description.rs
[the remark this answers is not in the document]
ahl: I should have clarified. I don't know if we have an MSRV! Yes, I know we specify the toolchain; that's at least in part to ensure repeatable builds as we have several tests for the `endpoint` macros that are a bit.. finicky with regard to version dependence.

I love `#[default]`; I don't know if using it would be annoying for our internal consumers (or maybe for external consumers; 👋 I don't know if you exist, but if you do, we love you!)

ahl:
> > Are other extractors permitted alongside `WebsocketUpgrade`? Or must it be the lone extractor?
> 
> They shouldn't be mutually exclusive (or else we wouldn't be able to do much useful with it, like say addressing a specific org/proj/inst in nexus by `Path<>` params)

Is it compatible with `Query` or `Body` extractors?

> > What does this look like in the OpenAPI output? The OpenAPI spec folks have pretty much declared Websockets as "not HTTP APIs" and therefore out of scope.
> 
> It looks like a GET endpoint with a `"x-dropshot-websocket": true` stashed in it, just like the `"x-dropshot-pagination": true` we already have. Here's an example (with the unrelated formatting condensed a little for brevity)

Cool!

> Would it be reasonable (in the interest of providing a websockets interface in the nearer term for serial & VNC API work) to punt on the full extent of that for the moment with the "x-dropshot-websocket" sentinel in an OpenAPI path's "get" method, and then follow up by implementing the proper AsyncAPI version?

Absolutely, and I apologize if I seemed to be implying otherwise. I don't want to stand in the way of urgent progress; I also want to have line of sight to how we want this to work in the future.

> For the use cases I'm thinking of, AsyncAPI's specifications are a bit heavier weight than we need, but that might be a blessing in this regard - I think we often want to forward _other_ protocols through raw binary messages, so most of the work that I think this would end up unblocking would have a fairly trivial expression as far as just annotating everywhere it's used with an effective "yep, it's binary" in such a way that our macro could accept before having a "real" AsyncAPI implementation backing it.

Neat. I haven't used AsyncAPI. I inferred that it lets you specify the protocol such that e.g. we could generate a client that had the messages as Rust types... but that's just imagineering.

> > Does it make sense to use the same `endpoint` macro with the constraints that the method must be a `GET`, the `WebsocketUpgrade` must be present, and the return type is `Result<Response<Body>, HttpError>`? Or might we want some other interface such as `#[channel { path = ... } ]` or `#[endpoint { method = WEBSOCKET, ...`?
> 
> I did have the thought of perhaps making `endpoint { method = WEBSOCKET, ...` decorating an async fn in which you're required to put an upgraded connection parameter (much like requiring path variables), and not have an intermediate function at all... but I shyed away from the former part of that approach at first because I didn't want to use a non-HTTP-verb there, and now the latter part too because would mean we wouldn't have a good way to expose the JoinHandle of what will likely be a longer-running future than most other kinds of endpoint.
> 
> I also had the use case of "what if we want to leave a path open for hybrid GET endpoints requests that may or may not be upgraded depending on how it's accessed," but I am increasingly unsure this use case would be worth the additional complexity.

Should we validate at least that folks are using a `GET` rather than some other method? I would be happy to help or to do this as a follow-on PR.

> `#[channel...` would be a nice way to both get a little closer to AsyncAPI terminology and possibly isolate some of the additional websocket/AsyncAPI-specific complexity from the `endpoint` decorator's implementation when it comes to things like describing message formats. My current inclination in light of this is to perhaps make such a thing as a trivial macro for now, that expands to the currently-equivalent `#[endpoint = GET` decorator.

That's a neat idea for a follow-on.

Thanks for the thorough discussion!

dropshot/src/api_description.rs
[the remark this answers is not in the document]
luqmana: I did mention the derive Default but we don't have an MSRV I could find so depends on how conservative we wanna be here I guess.

In terms of our own internal consumers, of the ones I know of:
- propolis lists an [MSRV of 1.62](https://github.com/oxidecomputer/propolis/#minimum-supported-rust-version-msrv)
- omicron is pinned on an [old nightly](https://github.com/oxidecomputer/omicron/blob/9bbf2a0fa436542e6bfe2f403114ba280fe3730d/rust-toolchain.toml#L13) (due to unstable features): nightly-2022-04-27
- crucible is also pinned on an [even older nightly](https://github.com/oxidecomputer/crucible/blob/1eac1be81e0b3d624f2d09aecc40601a7aaef5f1/rust-toolchain.toml#L2): nightly-2021-11-24
- buildomat doesn't have a rust-toolchain or note about version
- cio is pinned on an [old nightly](https://github.com/oxidecomputer/cio/blob/209300a352cb653fac660c0234f80749b5974a02/rust-toolchain#L1): nightly-2022-03-13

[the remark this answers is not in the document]
ahl: Do the old nightly's support #[default]? If so I say we go ahead and keep it. If not... I guess maybe roll them forward first?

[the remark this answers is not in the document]
lifning: i think i'm not quite as inclined to have us drag everything forward for something as trivial as some very small syntax sugar (vs. say, implementing an actual *feature* in such a way that otherwise wouldn't have been possible), but i wouldn't  say no if other folks felt strongly otherwise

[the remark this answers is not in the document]
luqmana: Yea, I don't think that that should hold this up.

lifning:
so the little "panic in Drop if it's not handled" scheme is kinda invalidated when any of the other extractors fail to parse from the request - it never reaches the user's code for them to call `handle`, which would probably be pretty inconvenient even when the panic is downgraded to a `warn!` when built in release mode. i'm leaning toward just trying to writing the wrapper for it in the proc macro, and if the user needs a JoinHandle there's theoretically nothing stopping them from `tokio::spawn`ing inside the handler, though it would theoretically be a slight perf hit.

dropshot/src/websocket.rs
```diff
@@ -0,0 +1,410 @@
...
+    * client that we are accepting the upgrade.
+    *
+    * `handler` is a closure that accepts a [`HyperUpgraded`] connection
+    * and returns a future that will be spawned by this function,
+    * in which the `Upgraded` connection may be used with your choice of
+    * websocket-handling code operating over an [`tokio::io::AsyncRead`] and
+    * [`tokio::io::AsyncWrite`] type (e.g. `tokio_tungstenite`).
+    *
+    * ```
+      #[dropshot::endpoint { method = GET, path = "/my/ws/endpoint" }]
```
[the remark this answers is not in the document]
lifning: i think this is resolved by the new `#[channel...]` approach, as we don't let `method` be specified at all in that macro (it's always GET for protocol=WEBSOCKETS)

dropshot/src/websocket.rs
```diff
@@ -0,0 +1,410 @@
...
+      async fn my_ws_endpoint(
+          rqctx: std::sync::Arc<dropshot::RequestContext<()>>,
+          websock: dropshot::WebsocketUpgrade,
+      ) -> Result<http::Response<hyper::Body>, dropshot::HttpError> {
```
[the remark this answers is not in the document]
lifning: we don't explicitly check for it in proc-macro-land, but in the latest patch i did introduce some friendlier type aliases for what's going on - and the user need not concern themselves with the `Result<Response<Body>, HttpError>` one any more; that's all abstracted away by `#[channel]` and all their socket-handling function has to return is a very generic `Result<(), Box<dyn Error+Send+Sync+'static>` (chosen so we can log their errors, if any, and let them use the ergonomic `?`)

dropshot/src/api_description.rs
```diff
@@ -1579,6 +1588,17 @@ pub struct TagExternalDocs {
     pub url: String,
 }
 
+/**
+ * Dropshot/Progenitor features used by endpoints which are not a part of the base OpenAPI spec.
```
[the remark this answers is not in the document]
lifning: per our discussion on matrix, i'll file some issues after this merges about turning this AsyncAPI support (and perhaps coming up with a general.. philosophy? on how to approach other extensions, like pagination or anything we come up with in the future that's outside of the scope of AsyncAPI in particular)

dropshot/src/handler.rs
```diff
@@ -202,7 +203,7 @@ pub trait Extractor: Send + Sync + Sized {
  * the associated endpoint is paginated.
  */
 pub struct ExtractorMetadata {
-    pub paginated: bool,
+    pub extension_mode: ExtensionMode,
```
ahl, line 206: @lifning it looks like this was a breaking change:

```
error[E0560]: struct `dropshot::ExtractorMetadata` has no field named `paginated`
  --> nexus/src/authn/external/cookies.rs:51:29
   |
51 |         ExtractorMetadata { paginated: false, parameters: vec![] }
   |                             ^^^^^^^^^ `dropshot::ExtractorMetadata` does not have this field
   |
   = note: available fields are: `extension_mode`, `parameters`
```

https://github.com/oxidecomputer/omicron/pull/1626

Could you please update the changelog?

lifning: https://github.com/oxidecomputer/dropshot/pull/428

PR #445, opened 2022-09-22
update to 1.64 and use workspace dependencies

justinmmott:
currently `trybuild` does not work properly with workspace defined dependencies. So, `serde` and `proc-macro2` have been left alone in the `dropshot/Cargo.toml` file. I have opened an [issue](https://github.com/dtolnay/trybuild/issues/187) on the `trybuild` repo to try and resolve this.

### DIFF
--- a/dropshot/tests/fail/bad_endpoint10.stderr
+++ b/dropshot/tests/fail/bad_endpoint10.stderr
@@ -14,18 +14,17 @@ note: required by a bound in `validate_result_error_type`
    |      ^^^^^^ required by this bound in `validate_result_error_type`
 
 error[E0277]: the trait bound `fn(Arc<RequestContext<()>>) -> impl Future<Output = Result<HttpResponseOk<()>, String>> {<impl From<bad_error_type> for ApiEndpoint<<Arc<RequestContext<()>> as RequestContextArgument>::Context>>::from::bad_error_type}: dropshot::handler::HttpHandlerFunc<_, _, _>` is not satisfied
-  --> tests/fail/bad_endpoint10.rs:14:10
+  --> tests/fail/bad_endpoint10.rs:10:1
    |
 10 | / #[endpoint {
 11 | |     method = GET,
 12 | |     path = "/test",
 13 | | }]
-   | |__- required by a bound introduced by this call
-14 |   async fn bad_error_type(
-   |            ^^^^^^^^^^^^^^ the trait `dropshot::handler::HttpHandlerFunc<_, _, _>` is not implemented for `fn(Arc<RequestContext<()>>) -> impl Future<Output = Result<HttpResponseOk<()>, String>> {<impl From<bad_error_type> for ApiEndpoint<<Arc<RequestContext<()>> as RequestContextArgument>::Context>>::from::bad_error_type}`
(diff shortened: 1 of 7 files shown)


justinmmott:
> currently `trybuild` does not work properly with workspace defined dependencies. So, `serde` and `proc-macro2` have been left alone in the `dropshot/Cargo.toml` file. I have opened an [issue](https://github.com/dtolnay/trybuild/issues/187) on the `trybuild` repo to try and resolve this.

No longer an issue as trybuild was updated

justinmmott:
https://github.com/dependabot/dependabot-core/issues/5315 seems that this wouldn't fix this issue as some of the test errors have changed slightly

davepacheco:
@justinmmott Thanks for doing this.  I see Renovate also opened a PR to update to 1.64 in #444.  We can close that one if we wind up moving forward with this one.

What does using workspace dependencies mean for consumers?  Does it mean that consumers must be using 1.64 or newer?

justinmmott:
@davepacheco  Sounds good to me. I'm not sure about what the workspace dependencies will do for customers, I was hoping the change to the `rust-toolchain.toml` would alleviate any issues with that. Also, [#typify/113](https://github.com/oxidecomputer/typify/pull/113) has some discussion on the matter as well. However, in this case I don't think that dependabot would be able to make the changes in the `.stderr` files.

davepacheco:
I believe rust-toolchain.toml is used only when people clone and build `dropshot` directly.  But if you've got an application (like our own omicron) that uses dropshot, then dropshot's rust-toolchain.toml is not used.  We use rust-toolchain in dropshot mainly to communicate (and enforce) that dropshot devs are using a consistent toolchain.  Dropshot itself is expected to work with a wider range of Rust versions (though we've never really said what that is, either).  I think we'd at least want to make sure our own projects won't break with this change.

davepacheco:
Ah I see @ahl also called out the dependency on https://github.com/dependabot/dependabot-core/issues/5315 -- that'd apply here too.

davepacheco:
FYI I went ahead and updated and landed #444 since it seems like there are still some open questions here.

justinmmott:
Ah no, I accidentally messed up while trying to merge main into my fork. I'll fix that asap

davepacheco:
Ah okay.  For what it's worth I think workspace inheritance is still promising here.  I think we just want to make sure it won't break consumers (either by verifying that it doesn't, or by updating the ones we know about and bumping the major rev of this crate).

justinmmott:
> Ah okay. For what it's worth I think workspace inheritance is still promising here. I think we just want to make sure it won't break consumers (either by verifying that it doesn't, or by updating the ones we know about and bumping the major rev of this crate).

Oh I totally agree, but if you guys are waiting on dependabot to make the updates automatically anyways. I can try and test to see if it will work down the line.

poliorcetics:
I opened a PR for it https://github.com/dependabot/dependabot-core/pull/5794, it should work but it's not merged yet.

It's probably possible to regenerate the docker container and use it locally to make the necessary changes but I have no idea how to do that. I'll gladly accept people testing it and telling me if it works has expected in a variety of configs though 👍

PR #451, opened 2022-09-28
add response types for various 300-level response codes

davepacheco:
My goal here is to add `HttpResponse*` types that you can use to have your Dropshot endpoint return each of these 300-level response codes:

* [302 Found](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302)
* [303 See Other](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303)
* [307 Temporary Redirect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307)

@ahl I'm interested in a bunch of your thoughts here.  The way I've done this, unlike the other response codes, these are _functions_ that return `HttpResponseHeaders`.  That way, the OpenAPI spec reflects the `Location` header.  It's a little weirdly not-parallel to the existing response types but I think the result is pretty good: you have to provide a location, we take care of putting it in the right place, and it lands in the spec.

**On the specific response codes:** these are all ways to redirect the client to a different URL.  People have traditionally used 302.  The MDN docs I linked to (and the linked HTTP specs) point out that you should really use 307 if you want the browser to basically retry this request at the other URL (which means the HTTP method and request body should be re-used for the second request), and you should really use 303 if you want the browser to make a "GET" to the other URL.  303 seems intended for the case where, say, you're completing a `POST` or `PUT` and you want to redirect somebody to a confirmation page.

**About the `Location` header and body:** I believe HTTP does not _require_ that responses with these status codes have a `Location` field, but I think at least for our use cases we should always have one, so I made it required here.  HTTP recommends a "small hypertext document" that conveys the gist of the redirect.  This seems irrelevant for an API so I just made these have an empty body.

**Use case:** This came up for me in omicron's login endpoints.  [These currently construct `Response<Body>` by hand](https://github.com/oxidecomputer/omicron/blob/54754a949840e2949af3a9c59785a0cbc645c056/nexus/src/external_api/console_api.rs#L84-L94), which is error-prone (and there are errors) and also doesn't describe the result in the OpenAPI spec.  I want to make these use a first-class dropshot-provided response type.  There are two cases: when you `GET` some login-related URLs, you may get redirected to the single sign-on page at another server (the external authentication provider).  In this case, any of these response codes is probably fine because the request came in as a GET (which has nobody) and you want it to be redirected as a GET (with no body).  303 seems most precise.  The other case is when you `POST` to an endpoint that accepts credentials, and the `POST` query parameters or request body _also_ provide a URL that you're supposed to be redirected to on successful login.  (The usual flow is that you access resource X, that sends you over to login, then you login, and we want to send you back to X.)  This should probably be a 303, since we want you to `GET` this URL, not `POST` it.


davepacheco:
I'm using these new interfaces in oxidecomputer/omicron#1757, in case you want to see what it looks like.

dropshot/src/handler.rs
```diff
@@ -1436,6 +1436,172 @@ impl From<HttpResponseUpdatedNoContent> for HttpHandlerResult {
...
+ * `http_response_found` returns an HTTP 302 "Found" response with no response
+ * body.
+ *
+ * The sole argument will become the value of the `Location` header.  This is
+ * where you want to redirect the client to.
+ *
+ * Per MDN and RFC 9110 S15.4.3, you might want to use 307 ("Temporary
+ * Redirect") or 303 ("See Other") instead.
+ */
+pub fn http_response_found(
```
ahl, line 1478: why `http_response_found` rather than `http_found` to be consistent with the others? because `http_found` would be confusing? Fair enough.

dropshot/src/handler.rs
```diff
@@ -1436,6 +1436,172 @@ impl From<HttpResponseUpdatedNoContent> for HttpHandlerResult {
...
+    location: &str,
+) -> HttpError {
+    HttpError::for_internal_error(format!(
+        "error encoding redirect URL {:?}: {:#}",
+        location, error
+    ))
+}
+
+/**
+ * This internal type impls HttpCodedResponse.  External clients should use
```
ahl, line 1500:
```suggestion
 * This internal type impls HttpCodedResponse.  Clients should use
```

?

dropshot/src/handler.rs
```diff
@@ -1436,6 +1436,172 @@ impl From<HttpResponseUpdatedNoContent> for HttpHandlerResult {
...
+    let _ = http::HeaderValue::from_str(&location)
+        .map_err(|e| http_redirect_error(e, &location))?;
+    Ok(HttpResponseHeaders::new(
+        HttpResponseSeeOtherStatus,
+        RedirectHeaders { location },
+    ))
+}
+
+/**
+ * This internal type impls HttpCodedResponse.  External clients should use
```
ahl, line 1545: maybe as above re: "External clients"?

dropshot/src/handler.rs
```diff
@@ -1436,6 +1436,172 @@ impl From<HttpResponseUpdatedNoContent> for HttpHandlerResult {
...
+ * `HttpResponseFound` instead, which includes metadata about the `Location`
+ * header.
+ */
+#[doc(hidden)]
+pub struct HttpResponseFoundStatus;
+impl HttpCodedResponse for HttpResponseFoundStatus {
+    type Body = Empty;
+    const STATUS_CODE: StatusCode = StatusCode::FOUND;
+    const DESCRIPTION: &'static str = "redirect (302)";
```
ahl, line 1509: What would you think about "redirect (found)" since the status code is already in... the status code.

dropshot/src/handler.rs
```diff
@@ -1436,6 +1436,172 @@ impl From<HttpResponseUpdatedNoContent> for HttpHandlerResult {
...
+     * also a fair bit more work to implement.  We'd need to create a separate
+     * type for this field so that we can impl `Serialize` and `JsonSchema` on
+     * it, and we'd need to also impl serialization of byte sequences in
+     * `MapSerializer`.  Ugh.
+     *
+     * We just use `String`.  This might contain values that aren't valid in
+     * HTTP response headers.  But we can at least validate that at runtime, and
+     * it sure is easier to implement!
+     */
+    location: String,
```
ahl, line 1461: do we want to `#[serde(rename = "Location")]`?

ahl (commented):
This looks great. Thanks for doing it.

Related: I have no idea what I'd do in progenitor to handle this. `reqwest` claims to do redirections for us... in which case we'd have no idea what kind of data we might get back. My inclination is--I think--to omit operations that have a 3xx response code, but if you have any thoughts I'd love to hear them.

dropshot/src/handler.rs
davepacheco: I don't mind either way, but [HTTP header names are case-insensitive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers).  Chatted offline -- will skip this.

dropshot/src/handler.rs
davepacheco: Done in 74d5e18 (and updated the others too)

dropshot/src/handler.rs
davepacheco: Done in 74d5e18 (and updated the others too).

dropshot/src/handler.rs
davepacheco: Done in 74d5e18 (and updated the others too).

davepacheco:
> Related: I have no idea what I'd do in progenitor to handle this. reqwest claims to do redirections for us... in which case we'd have no idea what kind of data we might get back. My inclination is--I think--to omit operations that have a 3xx response code, but if you have any thoughts I'd love to hear them.

Hmm.  In an ideal world, I don't think we want to skip them altogether.  I think it would be useful to be able to call the login endpoints from a generated client.  I expect those to return 303s.  In these cases, it would be fine to ignore the Location header and treat the response code is a general success code.  That's probably true for most 303s, since the point is usually to show a confirmation page or the like.

A 302 or 307 is a little different.  The server is telling you "this thing is over there -- try again there".  On the one hand, I don't think any of the operations in Nexus that returns these codes is that useful to all from a progenitor client.  (They're basically the endpoints that redirect a user to a web page to sign in.)  On the other hand, I think it's plausible we'd want to use these for other purposes.  Examples:

- backwards compatibility: we moved this route over there
- it's one way to make an alias for a resource appear elsewhere -- e.g., imagine if /organizations/:org\_name/FOO 307'd you to /by-id/organizations/:that\_org\_id/FOO (I'm not sure this is a great idea but it's an approach I've seen)
- we could use these when a resource gets renamed or moved -- e.g., if you rename Foo to Bar, we could 307 requests to Foo over to Bar for you

But I see the problem that the generated client doesn't know what it's going to get back.  Maybe we could add some other bit of metadata in the OpenAPI spec for the operation id that we might redirect you to?  That won't really help for the case where we add a 302/307 for compatibility reasons.  In that case, presumably the 302/307 wouldn't be in the openapi spec at all, there'd be some 200-level response instead, and the expectation would presumably be that you're getting that 200-level response content at the new URL.

I can see how this gets pretty messy.

ahl:
Maybe we can just generate a client function with a completely generic response value.

PR #452, opened 2022-09-29
Remove max log level features on slog dependency

jgallagher:
With these features in place, it wasn't possible for a consumer of `dropshot` to build a release binary that includes `trace!()` logs. Cargo features are additive, meaning `release_max_level_debug` was set for any crate that had dropshot in its dependency tree, and slog's evaluation of this feature set stops at the highest max level limit.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -15,6 +15,10 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.8.0\...HEAD[Full list of commits]
 
+=== Other notable changes
+
+* https://github.com/oxidecomputer/dropshot/pull/452[#452] Dropshot no longer enables the `slog` cargo features `max_level_trace` and `release_max_level_debug`. Previously, clients were unable to set a release log level of `trace`; now they can. However, clients that did not select their own max log levels will see behavior change from the levels Dropshot was choosing to the default levels of `slog` itself (`debug` for debug builds and `info` for release builds).
+
 == 0.8.0 (released 2022-09-09)
 
 https://github.com/oxidecomputer/dropshot/compare/v0.7.0\...v0.8.0[Full list of commits]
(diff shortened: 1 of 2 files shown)


davepacheco:
This looks good.  It makes sense that libraries probably shouldn't pick a feature here (although it's surprising to me that slog picks the highest level -- I guess there's no good choice, the way it's structured).

Is it possible to keep "max_level_trace" in dev-dependencies?  I think the original reason for this was so that we'd have verbose logs in tests, examples, etc.  It'd be nice if we could keep that.

davepacheco (approved):

jgallagher:
> Is it possible to keep "max_level_trace" in dev-dependencies? I think the original reason for this was so that we'd have verbose logs in tests, examples, etc. It'd be nice if we could keep that.

Maybe? Definitely yes if we switch to [Feature resolver version 2](https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2) with the 2021 edition. From that link, one of the differences with the new resolver is:

> Features enabled on dev-dependencies will not be unified when those same dependencies are used as a normal dependency, unless those dev-dependencies are currently being built. For example...

but that section ends with

> Note that dev-dependencies in dependencies are always ignored, this is only relevant for the top-level package or workspace members.

So I think it's probably fine to move those features to a `dev-dependency` on slog, and they'd only affect examples/tests for people working directly in this repo. I don't know if there are downstream consumers that are going to suddenly have a different log level due to this change, though (either in tests or for real).

davepacheco:
> > Note that dev-dependencies in dependencies are always ignored, this is only relevant for the top-level package or workspace members.
> 
> So I think it's probably fine to move those features to a `dev-dependency` on slog, and they'd only affect examples/tests for people working directly in this repo. I don't know if there are downstream consumers that are going to suddenly have a different log level due to this change, though (either in tests or for real).

I think that's okay.  That seems like a necessary consequence -- we're transferring control of the maximum log level to the dependency.  Let's make sure to note it in the CHANGELOG though.  Would you mind adding a note there in this PR?

jgallagher:
CHANGELOG entry added, and restored the features under `dev-dependencies`.

davepacheco (approved):
Thanks!

PR #459, opened 2022-10-05
Update stderr files for trybuild 1.0.66

dtolnay:
With https://github.com/dtolnay/trybuild/issues/86 fixed.

Without this PR, the current stderr files would still pass against either version of trybuild, because trybuild compares them against all old implementations of the normalization heuristic in sequence. However, the current stderr files could fail if code in dropshot/src/handler.rs moves around and changes the number of digits in the line number. After this PR, these tests would no longer fail if code in handler.rs gets moved around.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1766,9 +1766,9 @@ checksum = "59547bce71d9c38b83d9c0e92b6066c4253371f15005def0c30d9657f50c7642"
 
 [[package]]
 name = "trybuild"
-version = "1.0.65"
+version = "1.0.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


ahl:
Wait: you're taking this fix I made two years ago and showing us how to properly apply it to our repo? Jeez! Thank you 😃 !!

ahl:
Wow! Talk about great follow-up. Thanks for fixing the issue **and** for chasing this down!

ahl:
@dtolnay can you try pushing an empty commit to unstick github? We can't seem to unwedge it.

dtolnay:
Likely related to GitHub Actions being borked during when the PR was opened. I've had that before.
https://www.githubstatus.com/incidents/smn1qtqvbsb6

Issue #491, opened 2022-11-24
TLS certificate and key should be able to be replaced at runtime


jclulow:
At present, the private key and certificate for TLS must be provided once at startup of the server.  Short-lived certificates are becoming increasingly common; Let's Encrypt certificates are valid for 90 days, but it is recommended to renew them more often than that (e.g., every 30-60 days) to avoid issues around the expiry period.

Certificates also need to be re-issued if more alternative names are added to a non-wildcard certificate.

It would be good if `dropshot::HttpServer` provided a routine for installing new certificates, probably by passing a new `dropshot::ConfigTls`.  (see #490 for why it should not just be a pair of `PathBuf` arguments.)

PR #502, opened 2022-12-06
[tls] Allow refreshing TLS configuration information at runtime

smklein:
Adds a `refresh_tls` method to `HttpServer`, which allows TLS information to be updated for a running server.

Fixes https://github.com/oxidecomputer/dropshot/issues/491

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -15,6 +15,7 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.8.0\...HEAD[Full list of commits]
 
+* https://github.com/oxidecomputer/dropshot/pull/502[#502] Dropshot exposes a `refresh_tls` method to update the TLS certificates being used by a running server.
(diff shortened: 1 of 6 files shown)


jclulow:
How does the mutex on the acceptor work if there are no inbound connections being made?

jclulow:
Yes that makes sense thanks.  I had missed that we are not blocking at that point waiting on a connection, because of the select loop it's inside that actually deals directly with the system-level sockets.

CHANGELOG.adoc
```diff
@@ -15,6 +15,7 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.8.0\...HEAD[Full list of commits]
 
+* https://github.com/oxidecomputer/dropshot/pull/502[#502] Dropshot exposes a `refresh_tls` method to update the TLS certificates being used by a running server.
```
ahl, line 18: I may be mistaken, but I think this is the first breaking change in this list. If that's the case, can you please create two sections as below?

ahl (commented):
ship it

CHANGELOG.adoc
smklein: Sure, updated in https://github.com/oxidecomputer/dropshot/pull/502/commits/219ff7af4890eed9227d1cb9d5cb9931a5ee16d0

smklein:
FYI @davepacheco , since this is a breaking change

Issue #489, opened 2022-11-22
Endpoint macro should support `deprecated`


zephraph:
The openapi spec supports a `boolean` field on the operation object that signifies the operation is deprecated. We should allow that to be specified via the endpoint macro. 

Example

```rust
#[endpoint {
    method = GET,
    path = "/organizations/{organization_name}/projects/{project_name}/instances",
    tags = ["instances"],
    deprecated = true,
}]
```

One thing to consider is that if we deprecate something we should always provide documentation to the user on what to use instead. That could either be baked in to the api by having `deprecated` actually take a string or enforced later via a lint rule.

PR #503, opened 2022-12-06
allow endpoints to be marked as deprecated

ahl:
This lets us specify an endpoint as deprecated. The alternative path I considered was to use the rust built-in `#[deprecated]` attribute, but that seemed too cute.

I took the opportunity to simplify some of the `serde` use.


zephraph (approved):
Hell yeah, that was fast. Looks great, thanks for tackling this!

smklein:
Fixes: https://github.com/oxidecomputer/dropshot/issues/489

PR #522, opened 2022-12-20
Bump usdt dependency to support stable toolchains

bnaecker:


davepacheco:
The CI failure is #523.  Looks like this needs to be sync'd with "main".

davepacheco (commented):
Could you update the CHANGELOG?

I don't totally grok the dropshot/build.rs change.  Maybe a comment there would be useful?

davepacheco:
I buried the lede.  I'm really excited about this big step towards stable!

davepacheco (approved):
This looks good...but also seems easy to mess up.  @ahl want to take a look too?

dropshot/build.rs
```diff
@@ -13,15 +13,21 @@
...
+    // `feature(asm)` is stable.
     if version_check::is_min_version("1.59").unwrap_or(false) {
         println!("cargo:rustc-cfg=usdt_stable_asm");
     }
 
-    // Once asm_sym is stablilized, add an additional check so that those
-    // building on macos can use the stable toolchain with any hassle.
-    //
-    // A matching rust-cfg option named `usdt_stable_asm_sym` seems appropriate.
+    // `feature(asm_sym)` is stable, which is required on macOS.
```
ahl, line 28: this is fine, but FWIW we do it a little differently in usdt:

https://github.com/oxidecomputer/usdt/blob/master/probe-test-build/build.rs#L23-L28

ahl (approved):
ship it

dropshot/build.rs
bnaecker: Hmm, good point. I _think_ this is all logically the same, just inverting everything, right?

davepacheco:
See also #561

davepacheco (approved):

bnaecker:
My fault for multiple PRs. I meant to merge these, but just forgot to do so. #561 does the build.rs updates for better detection of the features, and this PR updates the actual `usdt` dependency and adds changelog notes.

Issue #516, opened 2022-12-16
test failure in `test_config_bind_address_https`


davepacheco:
From [this CI run](https://github.com/oxidecomputer/dropshot/actions/runs/3706071235/jobs/6280812539) on #514:

```
---- test_config_bind_address_https stdout ----
log file: "/tmp/test_config-c204f22231b1f0ef-config_bind_address_https.14352.1.log"
note: configured to log to "/tmp/test_config-c204f22231b1f0ef-config_bind_address_https.14352.1.log"
error: test failed, to rerun pass `-p dropshot --test test_config`
thread 'test_config_bind_address_https' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 98, kind: AddrInUse, message: "Address already in use" }', dropshot/tests/test_config.rs:149:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

@smklein could this be related to the recent TLS changes?

bnaecker:
I'm seeing the same thing on [this test run](https://github.com/oxidecomputer/dropshot/actions/runs/3742954784/jobs/6354701742) too. I reran the test to check if it was transient, but it failed twice in a row.

PR #523, opened 2022-12-20
[tests] Avoid port collision

smklein:
This is a short-term fix to #516  .

`test_config_bind_server` -- a primary helper function for the config tests -- does the following:

- Tries to start a client before a server starts
- Starts a server, then immediately stops it
- Tries to connect to that server, and validates that it can't (because it was shut down?)
- Creates a server at `bind_port + 1`:
https://github.com/oxidecomputer/dropshot/blob/9e12b7074299f9f1c8cc7b1f0eac02ed04679d81/dropshot/tests/test_config.rs#L220-L225
- Tries to connect to the old bind port server (still failing)
- Closes the new server
- Tries to connect to the new bind port server (should fail again, because it has shut down)

This PR fixes this issue *expediently* by ensuring all the selected ports are at an offset of at least "2" away from each other to avoid trampling.

**I strongly think we should change the structure of this test, to avoid this hard-coded port selection**. However, this fix should stop the test flakes from occurring.

Fixes https://github.com/oxidecomputer/dropshot/issues/516

### DIFF
--- a/dropshot/tests/test_config.rs
+++ b/dropshot/tests/test_config.rs
@@ -410,7 +410,7 @@ async fn test_config_bind_address_https_buffer() {
         ConfigBindServerHttps { log, certs, serialized_certs, serialized_key };
 
     /* This must be different than the bind_port used in the http test. */
-    let bind_port = 12218;
+    let bind_port = 12219;
     test_config_bind_server::<_, ConfigBindServerHttps>(test_config, bind_port)
         .await;
 


davepacheco:
Thanks for taking a look at this.  This looks like the right fix to me and it makes sense what happened here.  We're assuming `test_config_bind_address_https` and `test_config_bind_address_https_buffer` ran concurrently and picked overlapping ports.

I believe this test is attempting to verify that if a user provides a specific port, then we use that port.  I'm not sure how we avoid a hardcoded port selection while verifying that.

davepacheco (approved):

Issue #536, opened 2023-01-02
Cannot "refresh_tls" while also waiting for server to terminate


smklein:
Let's suppose someone starts a dropshot server with the following expectations:

- They spin up the server, and would like to immediately monitor for it closing (currently, this is done via `.await`, since `HttpServer` implements [Future](https://docs.rs/dropshot/0.8.0/dropshot/struct.HttpServer.html#impl-Future-for-HttpServer%3CC%3E)).
- They'd also like to expose an endpoint to "refresh TLS certificates", which should be able to reference the [`HttpServer`](https://docs.rs/dropshot/0.8.0/dropshot/struct.HttpServer.html) and call `refresh_tls` on it to get new TLS certificates.

This is currently **not possible**.

- The `impl Future` for the server effectively takes ownership of the entire `HttpServer` structure...
- ... so after a caller starts "waiting for the server to shut down", they can no longer access the `HttpServer` object.

smklein:
I think it may be possible to resolve this by creating a separate "ServerCloser" object, distinct from the `HttpServer` object.

This would give the caller an object to consume (for `.await` purposes) while also letting them have a reference to the `HttpServer` object for online configuration updates.

PR #537, opened 2023-01-02
Add `wait_for_shutdown` method to await server termination

smklein:
- Allows callers to await completion of a server separate from owning the server object.

Fixes https://github.com/oxidecomputer/dropshot/issues/536

### DIFF
--- a/dropshot/examples/self-referential.rs
+++ b/dropshot/examples/self-referential.rs
@@ -1,0 +1,101 @@
+// Copyright 2020 Oxide Computer Company
+
+use dropshot::endpoint;
+use dropshot::ApiDescription;
+use dropshot::ConfigDropshot;
+use dropshot::ConfigLogging;
+use dropshot::ConfigLoggingLevel;
+use dropshot::HttpError;
+use dropshot::HttpResponseOk;
+use dropshot::HttpServerStarter;
+use dropshot::RequestContext;
+use schemars::JsonSchema;
+use serde::Deserialize;
+use serde::Serialize;
+use std::sync::atomic::AtomicU64;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+
+#[tokio::main]
+async fn main() -> Result<(), String> {
+    let config_dropshot: ConfigDropshot = Default::default();
+    let config_logging =
+        ConfigLogging::StderrTerminal { level: ConfigLoggingLevel::Info };
+    let log = config_logging
+        .to_logger("example-self-referential")
+        .map_err(|error| format!("failed to create logger: {}", error))?;
+
+    let mut api = ApiDescription::new();
+    api.register(example_api_get_counter).unwrap();
+
+    let api_context = Arc::new(ExampleContext::new());
+
+    // Set up the server.
+    //
+    // Note that we split the "closer" from the "server", so we can continue
+    // to reference the server while simultaneously waiting for it to shut down.
+    let server =
+        HttpServerStarter::new(&config_dropshot, api, api_context, &log)
+            .map_err(|error| format!("failed to create server: {}", error))?
+            .start();
+    let (server, closer) = server.get_closer();
+    let (server, joiner) = server.get_joiner();
+
+    tokio::task::spawn(async move {
+        loop {
+            // We can access the server's data while simultaneously awaiting
+            // its termination.
+            let value =
+                server.app_private().counter.fetch_add(1, Ordering::SeqCst);
+            const TIMEOUT: u64 = 5;
+            if value >= TIMEOUT {
+                println!("Terminating now");
+                break;
+            } else {
+                println!("Terminating in {} seconds", TIMEOUT - value);
+                tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+            }
+        }
+        // Once the timeout has been reached, we stop the server ourselves.
+        closer.close().await.unwrap();
+    });
+
+    // Wait for the server to stop.
+    joiner.await
+}
+
+/**
+ * Application-specific example context (state shared by handler functions)
+ */
+struct ExampleContext {
+    counter: AtomicU64,
+}
+
+impl ExampleContext {
+    pub fn new() -> ExampleContext {
+        ExampleContext { counter: AtomicU64::new(0) }
+    }
+}
+
+/*
+ * HTTP API interface
+ */
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+struct CounterValue {
+    counter: u64,
+}
+
+#[endpoint {
+    method = GET,
+    path = "/counter",
+}]
+async fn example_api_get_counter(
+    rqctx: Arc<RequestContext<Arc<ExampleContext>>>,
+) -> Result<HttpResponseOk<CounterValue>, HttpError> {
+    let api_context = rqctx.context();
+
+    Ok(HttpResponseOk(CounterValue {
+        counter: api_context.counter.load(Ordering::SeqCst),
+    }))
+}
--- a/dropshot/examples/self-referential.rs
+++ b/dropshot/examples/self-referential.rs
@@ -1,5 +1,7 @@
 // Copyright 2020 Oxide Computer Company
 
+//! An example which demonstrates a server that can manage its own lifecycle.
+
 use dropshot::endpoint;
 use dropshot::ApiDescription;
 use dropshot::ConfigDropshot;
--- a/dropshot/examples/self-referential.rs
+++ b/dropshot/examples/self-referential.rs
@@ -40,7 +40,6 @@ async fn main() -> Result<(), String> {
         HttpServerStarter::new(&config_dropshot, api, api_context, &log)
             .map_err(|error| format!("failed to create server: {}", error))?
             .start();
-    let (server, close_handle) = server.take_close_handle();
     let shutdown = server.wait_for_shutdown();
 
     tokio::task::spawn(async move {
@@ -59,13 +58,10 @@ async fn main() -> Result<(), String> {
             }
         }
         // Once the timeout has been reached, we stop the server ourselves.
-        close_handle.close().await.unwrap();
-
-        // This background task can also wait for the server to stop
-        let _ = server.await;
+        server.close().await.unwrap();
     });
 
-    // Wait for the server to stop.
+    // From a separate task, wait for the server to stop.
     shutdown.await
 }
 
--- a/dropshot/examples/self-referential.rs
+++ b/dropshot/examples/self-referential.rs
@@ -40,8 +40,8 @@ async fn main() -> Result<(), String> {
         HttpServerStarter::new(&config_dropshot, api, api_context, &log)
             .map_err(|error| format!("failed to create server: {}", error))?
             .start();
-    let (server, closer) = server.get_closer();
-    let (server, joiner) = server.get_joiner();
+    let (server, close_handle) = server.take_close_handle();
+    let shutdown = server.wait_for_shutdown();
 
     tokio::task::spawn(async move {
         loop {
@@ -59,11 +59,14 @@ async fn main() -> Result<(), String> {
             }
         }
         // Once the timeout has been reached, we stop the server ourselves.
-        closer.close().await.unwrap();
+        close_handle.close().await.unwrap();
+
+        // This background task can also wait for the server to stop
+        let _ = server.await;
     });
 
     // Wait for the server to stop.
-    joiner.await
+    shutdown.await
 }
 
 /**
--- a/dropshot/examples/self-referential.rs
+++ b/dropshot/examples/self-referential.rs
@@ -1,6 +1,6 @@
-// Copyright 2020 Oxide Computer Company
+// Copyright 2023 Oxide Computer Company
 
-//! An example which demonstrates a server that can manage its own lifecycle.
+//! An example that demonstrates server shutdown (and waiting for shutdown).
 
 use dropshot::endpoint;
 use dropshot::ApiDescription;
@@ -33,9 +33,6 @@ async fn main() -> Result<(), String> {
     let api_context = Arc::new(ExampleContext::new());
 
     // Set up the server.
-    //
-    // Note that we split the "closer" from the "server", so we can continue
-    // to reference the server while simultaneously waiting for it to shut down.
     let server =
         HttpServerStarter::new(&config_dropshot, api, api_context, &log)
             .map_err(|error| format!("failed to create server: {}", error))?
--- a/dropshot/src/server.rs
+++ b/dropshot/src/server.rs
@@ -28,6 +28,7 @@ use hyper::Response;
 use rustls;
 use std::convert::TryFrom;
 use std::future::Future;
+use std::marker::PhantomData;
 use std::net::SocketAddr;
 use std::num::NonZeroU32;
 use std::pin::Pin;
@@ -200,8 +201,9 @@ impl<C: ServerContext> HttpServerStarter<C> {
             probe_registration,
             app_state: self.app_state,
             local_addr: self.local_addr,
-            join_handle: Some(join_handle),
-            close_channel: Some(tx),
+            closer: Some(HttpServerCloser { close_channel: Some(tx) }),
+            joiner: Some(HttpServerJoiner { join_handle: Some(join_handle) }),
+            _phantom: PhantomData,
         }
     }
 }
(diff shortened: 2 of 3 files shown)


dropshot/examples/self-referential.rs
```diff
@@ -0,0 +1,101 @@
...
+            .start();
+    let (server, closer) = server.get_closer();
+    let (server, joiner) = server.get_joiner();
+
+    tokio::task::spawn(async move {
+        loop {
+            // We can access the server's data while simultaneously awaiting
+            // its termination.
+            let value =
+                server.app_private().counter.fetch_add(1, Ordering::SeqCst);
```
smklein, line 49: In this example, I'm poking at `app_private` while also `.await`-ing termination, but this same split would be necessary to `refresh_tls` certificates while awaiting termination of the server from another thread.

dropshot/src/server.rs
```diff
@@ -543,18 +545,83 @@ impl<C: ServerContext> Service<&TlsConn> for ServerConnectionHandler<C> {
...
+
+/// Implements [Joiner] and identifies that the server can be joined
+pub struct Joinable {}
+
+/// Implements [Joiner] and identifies that the server cannot be joined
+/// directly.
+pub struct Unjoinable {}
+
+impl Joiner for Joinable {}
+impl Joiner for Unjoinable {}
```
smklein, line 575: TECHNICALLY none of these traits are necessary. I *could* just drop them, and panic if you call "close" or "join" when you shouldn't.

However, I figured it would be cleaner to provide some compile-time guarantees -- you can't `.await` a server if you already called `.get_joiner` on it, and you can't `close` a server if you already called `.get_closer` on it.

dropshot/src/server.rs
```diff
@@ -578,15 +645,84 @@ impl<C: ServerContext> HttpServer<C> {
...
+                _phantom: PhantomData,
+            },
+            closer.unwrap(),
+        )
+    }
+}
+
+impl<C: ServerContext> HttpServer<C, Closeable, Joinable> {
+    /// Signals the currently running server to stop and waits for it to exit.
     pub async fn close(mut self) -> Result<(), String> {
```
smklein, line 715: This function is really doing "close and join", but for backwards compatibility, it's just called "close".

dropshot/examples/self-referential.rs
```diff
@@ -0,0 +1,101 @@
...
+                println!("Terminating in {} seconds", TIMEOUT - value);
+                tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+            }
+        }
+        // Once the timeout has been reached, we stop the server ourselves.
+        closer.close().await.unwrap();
+    });
+
+    // Wait for the server to stop.
+    joiner.await
```
smklein, line 64: Not that we're joining on the `joiner` here. It's not possible to call `server.await` after calling `server.get_joiner`, but doing so let's the "server" and "joiner" have different lifetimes.

dropshot/examples/self-referential.rs
```diff
@@ -0,0 +1,101 @@
...
+            if value >= TIMEOUT {
+                println!("Terminating now");
+                break;
+            } else {
+                println!("Terminating in {} seconds", TIMEOUT - value);
+                tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+            }
+        }
+        // Once the timeout has been reached, we stop the server ourselves.
+        closer.close().await.unwrap();
```
smklein, line 60: Not that we're closing the `closer` here. It's not possible to call `server.close` after calling `server.get_closer`, but doing so let's the "server" and "closer" have different lifetimes.

dropshot/src/server.rs
```diff
@@ -543,18 +545,83 @@ impl<C: ServerContext> Service<&TlsConn> for ServerConnectionHandler<C> {
     }
 }
 
-/**
- * A running Dropshot HTTP server.
- */
-pub struct HttpServer<C: ServerContext> {
+/// Marker trait used to control whether or not an HTTP server can be closed.
+pub trait Closer: Send + Sync + Unpin + 'static {}
```
jgallagher, line 549: I had a couple thoughts here:

1. Do we actually need the `Closer` / `Joiner` traits? There's no way for a caller to provide new implementors of these types, and we only ever have 2 concrete marker types for each one.
2. Could we move the `HttpServerCloser` / `HttpServerJoiner` into `Closeable` / `Joinable`, removing the need to hold `Option<_>`s at the `HttpServer` level?

I had to implement this to convince myself this would work; I think it does: https://github.com/oxidecomputer/dropshot/compare/cleanup...cleanup-tweak. A nice side effect is that our `Future`/`FusedFuture` impls can call down to the `HttpServerJoiner` implementations instead of copying them.

smklein: I like that suggestion a lot; pulled it into https://github.com/oxidecomputer/dropshot/pull/537/commits/a06a78a67828d9368dd6c45d6fbf94ba7fc06866

smklein: I went one step further and flattened the structures even more in b2ce08def4659b7c6694f0a87b04f5f22b5da858 , which IMO also simplifies the readability -- no need for tuple structs

dropshot/examples/self-referential.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright 2020 Oxide Computer Company
```
davepacheco, line 1: Can you add a comment here describing this example?

dropshot/src/server.rs
```diff
@@ -543,18 +543,55 @@ impl<C: ServerContext> Service<&TlsConn> for ServerConnectionHandler<C> {
...
+            .send(())
+            .expect("failed to send close signal");
+        Ok(())
+    }
+}
+
+/// Identifies that the server cannot be closed directly.
+pub struct Uncloseable {}
+
+/// Represents the ability to join a running HTTP server.
```
davepacheco, line 584:
```suggestion
/// Handle used to wait for a running HTTP server to shut down
///
/// Note that you can obtain this object while the server is still running, in which case `await`ing on this future will block until someone triggers a shutdown _and_ that shutdown completes. 
```

(I didn't like "join" here in the user-facing description because we're a few steps afar from abstractions where that verb is used -- tasks, threads -- and I thought someone might not understand what "join" means here.)

dropshot/src/server.rs
```diff
@@ -543,18 +543,55 @@ impl<C: ServerContext> Service<&TlsConn> for ServerConnectionHandler<C> {
...
+pub struct HttpServer<C: ServerContext, Closer = Closeable, Joiner = Joinable> {
     probe_registration: ProbeRegistration,
     app_state: Arc<DropshotState<C>>,
     local_addr: SocketAddr,
-    join_handle: Option<tokio::task::JoinHandle<Result<(), hyper::Error>>>,
+    closer: Closer,
+    joiner: Joiner,
+}
+
+/// Represents the ability to close a running HTTP server.
```
davepacheco, line 560:
```suggestion
/// Handle used to trigger shutdown of an [`HttpServer`]
```

dropshot/src/server.rs
```diff
@@ -543,18 +543,55 @@ impl<C: ServerContext> Service<&TlsConn> for ServerConnectionHandler<C> {
...
+///
+/// Returned from [HttpServer::get_closer].
+pub struct Closeable {
```
davepacheco, line 563:
```suggestion
pub struct CloseHandle {
```

I find "Closeable" and "Joinable", both described as "abilities", a bit confusing.  Those sound like they would be traits.  I've suggested "CloseHandle" and "ShutdownWaitHandle" instead and suggested some doc changes.  Let me know what you think.

dropshot/src/server.rs
```diff
@@ -543,18 +543,55 @@ impl<C: ServerContext> Service<&TlsConn> for ServerConnectionHandler<C> {
...
-/**
- * A running Dropshot HTTP server.
- */
-pub struct HttpServer<C: ServerContext> {
+/// A running Dropshot HTTP server.
+///
+/// The generic traits represent the following:
+/// - C: Caller-supplied server context
+/// - Closer: Identifies if the server is directly closeable.
+/// - Joiner: Identifies if the server is directly joinable.
```
davepacheco, line 551: Tell me if I've got this right: the way I figure this working is that:

- Callers will get an `HttpServer<C, Closeable, Joinable>` without having specified `Closeable` or `Joinable` directly.  (They _will_ specify `C`.)
- The functions that get the `Closeable` or `Joinable` consume the `HttpServer` and return a new one with a different type (different type parameters for `Closer` or `Joiner`, respectively).

As a result, consumers mostly don't have to think about these type parameters.  Is that right?

The exception I can think of is that if you want to write a function that accepts an `HttpServer`, I expect you usually won't care whether about the `Closer` or `Joiner`, but your function will probably have to be generic over theses parameters.  That seems not ideal but probably a fine tradeoff.  This doesn't seem super common.

dropshot/src/server.rs
```diff
@@ -543,18 +543,55 @@ impl<C: ServerContext> Service<&TlsConn> for ServerConnectionHandler<C> {
...
+            .take()
+            .expect("cannot close twice")
+            .send(())
+            .expect("failed to send close signal");
+        Ok(())
+    }
+}
+
+/// Identifies that the server cannot be closed directly.
+pub struct Uncloseable {}
```
davepacheco, line 582: When would this be used aside from a case where `close()` was already invoked?  Should this just be called `Closed`?

dropshot/src/server.rs
```diff
@@ -543,18 +543,55 @@ impl<C: ServerContext> Service<&TlsConn> for ServerConnectionHandler<C> {
...
+
+/// Represents the ability to join a running HTTP server.
+///
+/// Returned from [HttpServer::get_joiner].
+pub struct Joinable {
```
davepacheco, line 587:
```suggestion
pub struct ShutdownWaitHandle {
```

(see above -- I'm not sold on this but I think it's clearer than `Joinable`)

dropshot/src/server.rs
```diff
@@ -578,15 +615,78 @@ impl<C: ServerContext> HttpServer<C> {
...
-        self.close_channel
+    pub fn probe_registration(&self) -> &ProbeRegistration {
+        &self.probe_registration
+    }
+}
+
+impl<C: ServerContext, Closer> HttpServer<C, Closer, Joinable> {
+    /// Separates the [HttpServer] from the [Joinable] object which is
+    /// capable of joining it.
+    pub fn get_joiner(self) -> (HttpServer<C, Closer, Unjoinable>, Joinable) {
```
davepacheco, line 630:
```suggestion
    /// Wait for the [HttpServer] to shut down
    ///
    /// The returned `Future` completes when the server has shut down.  Note that
    /// that won't happen until someone actually starts shutting down the server, which
    /// may not have happened yet.
    pub fn wait_for_shutdown(self) -> (HttpServer<C, Closer, Unjoinable>, Joinable) {
```

dropshot/src/server.rs
```diff
@@ -578,15 +615,78 @@ impl<C: ServerContext> HttpServer<C> {
...
+            },
+            joiner,
+        )
+    }
+}
+
+impl<C: ServerContext, Joiner> HttpServer<C, Closeable, Joiner> {
+    /// Separates the [HttpServer] from the [Closeable] object which is
+    /// capable of closing it.
+    pub fn get_closer(self) -> (HttpServer<C, Uncloseable, Joiner>, Closeable) {
```
davepacheco, line 655:
```suggestion
    /// Returns a handle that can be used to close the server
    pub fn closer(self) -> (HttpServer<C, Uncloseable, Joiner>, Closeable) {
```

dropshot/src/server.rs
```diff
@@ -578,15 +615,78 @@ impl<C: ServerContext> HttpServer<C> {
...
+                local_addr,
+                closer,
+                joiner: Unjoinable {},
+            },
+            joiner,
+        )
+    }
+}
+
+impl<C: ServerContext, Joiner> HttpServer<C, Closeable, Joiner> {
```
davepacheco, line 652: Probably a silly question, but: why can't we just impl `close()` directly on the `HttpServer` in this block? (i.e., why have the impl of `close()` be on the `Closeable` instead of just using that as a marker?)

dropshot/src/server.rs
```diff
@@ -642,12 +733,34 @@ impl<C: ServerContext> Future for HttpServer<C> {
...
+
+impl<C: ServerContext, Closer> FusedFuture for HttpServer<C, Closer, Joinable>
+where
+    Closer: Unpin,
+{
+    fn is_terminated(&self) -> bool {
+        self.joiner.is_terminated()
+    }
+}
+
```
davepacheco, line 763: I wonder if we should just remove the impl of `Future` on `HttpServer` so that there's only one way to do all this instead of two.  What do you think?

dropshot/examples/self-referential.rs
```diff
@@ -0,0 +1,101 @@
...
+    // Note that we split the "closer" from the "server", so we can continue
+    // to reference the server while simultaneously waiting for it to shut down.
+    let server =
+        HttpServerStarter::new(&config_dropshot, api, api_context, &log)
+            .map_err(|error| format!("failed to create server: {}", error))?
+            .start();
+    let (server, closer) = server.get_closer();
+    let (server, joiner) = server.get_joiner();
+
+    tokio::task::spawn(async move {
```
davepacheco, line 44: It might make a more compelling example to have the example request trigger a shutdown.

davepacheco (commented):
Is this understanding right:

- Previously, `close()` consumed the `HttpServer` _and_ waited for it to stop.  It consumed the `HttpServer` so that you can't `close()` it again or do anything else with it (which is good).
- Similarly, awaiting the future to wait for the server to stop consumes the `HttpServer`, for the same reason -- once it's stopped, you aren't supposed to be able to do anything with it, so you shouldn't still have a handle to it.
- Now, we want to be able to (1) start waiting for the server to shut down in one task while (2) triggering a TLS refresh from the other task.  That means we need a separate handle for waiting on the shutdown.  Getting that handle must not consume the server.

If that's all right, that makes sense to me.  That said, is it important to ensure that only one thing can ever wait for the server to shut down?  If we relax that constraint, it seems a lot simpler.  Could we have a `HttpServer::wait_for_shutdown()` that returns `impl Future<Output=()>` whose implementation could be receiver for a oneshot or watch channel?  I think you could eliminate one of the new type parameters here.  If we do this, I wonder if we still need to separate the `Closer` at all.

This isn't that important.  I mention it because server lifecycle is already hard, getting it right can impact availability, and so if we can keep it simpler, that seems worthwhile.

dropshot/examples/self-referential.rs
smklein: Done

dropshot/examples/self-referential.rs
smklein: It could, but we'd need to place the server in the `ExampleContext`, which would spread the logic out a bit. I kinda wanted to keep the lifecycle separation example more condensed.

dropshot/src/server.rs
smklein: That's right - though with this latest PR, the `Joinable` generic parameter has been removed.

dropshot/src/server.rs
smklein: Done

dropshot/src/server.rs
smklein: Updated

dropshot/src/server.rs
smklein: It's a generic parameter placed into the `HttpServer` - it acts like `()`, but has a  more descriptive name.

Calling `HttpServer::take_close_handle` causes the server to have this property. It does *not* mean the server is closed -- just that the server object does not control the ability to close anymore.

dropshot/src/server.rs
smklein: This struct no longer exists; using `wait_for_shutdown` instead

dropshot/src/server.rs
smklein: I actually *did* impl `close()` on the `HttpServer`, but it wasn't nearby in the code. Re-located.

dropshot/src/server.rs
smklein: I opted to use the name `take_close_handle`, as I think it's more descriptive. It's a `take` operation because it changes the server, and it's more readable than `closer`.

dropshot/src/server.rs
smklein: In a vacuum, I'd be fine with this, but doing so would make this a backwards-incompatible change in a pretty significant way.

dropshot/examples/self-referential.rs
```diff
@@ -0,0 +1,102 @@
+// Copyright 2020 Oxide Computer Company
```
ahl, line 1: update the copyright year?

ahl (commented):
looks good; nothing useful to say; just a nit

dropshot/src/server.rs
```diff
@@ -200,8 +206,8 @@ impl<C: ServerContext> HttpServerStarter<C> {
             probe_registration,
             app_state: self.app_state,
             local_addr: self.local_addr,
-            join_handle: Some(join_handle),
-            close_channel: Some(tx),
+            closer: CloseHandle { close_channel: Some(tx) },
+            join_future: join_handle.boxed().shared(),
```
davepacheco, line 210: This is neat.  I'd never run across `shared()` before.  (I usually wind up using a channel and cloning the receiver.)

dropshot/src/server.rs
```diff
@@ -543,14 +549,23 @@ impl<C: ServerContext> Service<&TlsConn> for ServerConnectionHandler<C> {
...
-/**
- * A running Dropshot HTTP server.
- */
+type SharedBoxFuture<T> = Shared<Pin<Box<dyn Future<Output = T> + Send>>>;
+
+/// A running Dropshot HTTP server.
+///
+/// The generic traits represent the following:
+/// - C: Caller-supplied server context
+/// - Closer: Identifies if the server is directly closeable.
```
davepacheco, line 558: This is gone now.

dropshot/examples/self-referential.rs
```diff
@@ -0,0 +1,102 @@
...
+        .map_err(|error| format!("failed to create logger: {}", error))?;
+
+    let mut api = ApiDescription::new();
+    api.register(example_api_get_counter).unwrap();
+
+    let api_context = Arc::new(ExampleContext::new());
+
+    // Set up the server.
+    //
+    // Note that we split the "closer" from the "server", so we can continue
```
davepacheco, line 37: This comment needs an update for the latest changes.

dropshot/examples/self-referential.rs
```diff
@@ -0,0 +1,102 @@
+// Copyright 2020 Oxide Computer Company
+
+//! An example which demonstrates a server that can manage its own lifecycle.
```
davepacheco, line 3:
```suggestion
//! An example that demonstrates server shutdown (and waiting for shutdown)
```

dropshot/src/server.rs
```diff
@@ -604,6 +600,29 @@ impl<C: ServerContext> HttpServer<C> {
     pub fn probe_registration(&self) -> &ProbeRegistration {
         &self.probe_registration
     }
+
+    /// Returns a future which completes when the server has shut down.
+    ///
+    /// This function does not cause the server to shut down, it just waits for
```
davepacheco, line 606:
```suggestion
    /// This function does not cause the server to shut down.  It just waits for
```

davepacheco (approved):
Thanks for this.  Relaying from offline chat: I like that the latest version best because it eliminates both type parameters.  But I can also see wanting to separate the close handle in the future, and I'm cool with doing that now, too (with the extra type parameter or not).

My only comments here are nits on a few out of date comments.

dropshot/src/server.rs
smklein: Done

dropshot/examples/self-referential.rs
smklein: Done

dropshot/examples/self-referential.rs
smklein: Done

dropshot/src/server.rs
smklein: Done

dropshot/examples/self-referential.rs
smklein: Done

PR #539, opened 2023-01-04
[logging] wrap file loggers in a `LineWriter`

sunshowers:
I noticed while trying to log a large request that the logger was
`write(2)`ing one byte at a time(!)

Since `slog_bunyan` uses a newline-delimited JSON format, use a
`LineWriter` around the file.


smklein (approved):
Good catch!

PR #540, opened 2023-01-04
[logging] switch to camino Utf8PathBuf

sunshowers:
We repeatedly assert and assume that logging paths are valid UTF-8, and
use a `String` in the deserializable log even though that isn't the most
appropriate use.

Use [camino](https://docs.rs/camino/latest/camino/) to assert the use of `Utf8PathBuf` at the type-level.
This is a breaking change from the Rust API point of view, but
existing config files should work without changes since a `Utf8PathBuf`
is equivalent to a `String` with `PathBuf` semantics.


davepacheco (approved):
Nice!  Thanks.

davepacheco:
I imagine this is technically a breaking change because of the change in types on the config object.  Could you add a mention to the changelog?

PR #541, opened 2023-01-05
Add a new `ChunkedBody` extractor

sunshowers:
The current `UntypedBody` extractor writes data into a single `Vec<u8>`.
Consider what happens if the body is large (e.g. 100MB, which can happen
if uploading an artifact over HTTP). As each chunk (typically 10-100KB)
comes in, we'll have to both copy data from the incoming `Bytes`, and
reallocate the `Vec` over and over.

To avoid this issue, Eliza Weisman and I wrote buf-list, which
represents a list (really a queue) of chunks that can be operated on
using standard Tokio and other abstractions:
https://crates.io/crates/buf-list.

Add a `ChunkedBody` extractor that uses a `BufList` inside, avoiding
this issue. Also add `ChunkedBody` to usage suggestions.

One other change I did is to remove the nonexistent type parameter `J`
from `UntypedBody<J>` suggestions -- that didn't look right.

# Questions

- [ ] One consideration here is that `BufList` needs to be exposed as a public API. It's currently at 0.1 -- I could release a 1.0 if that would be helpful as far as exposing in the API goes. What do you think?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -109,6 +109,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "buf-list"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "deb213ab6aa87733e74428d8a33f0bb93181810b0ae09ae96b8cc1521d265283"
+dependencies = [
+ "bytes",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -285,6 +294,7 @@ dependencies = [
  "async-stream",
  "async-trait",
  "base64 0.20.0",
+ "buf-list",
  "bytes",
  "chrono",
  "dropshot_endpoint",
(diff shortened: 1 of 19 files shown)


ahl:
I haven't yet looked at the code, but I had a couple of questions about (my understanding of) the approach:

Do we need to expose a separate interface for large requests in addition to small ones? This seems like an additional burden to end-users we should be reluctant to impose.

Would it be feasible to change `UntypedBody` to expose `Stream` interface? Effectively by encapsulating `hyper::HttpBody`? Users could then read the contents into a `Bytes` or a `BufList` or whatever else they like.

ahl:
First, I don't think we need to be beholden to API compatibility.

Second, I think we could encapsulate an `hyper::HttpBody` in such a way that would retain compatibility. In particular--and this is without any experimentation, so I may be very wrong--by changing `UntypedBody::as_bytes` to effectively do the work currently done in `http_read_body`.

sunshowers:
Hmm, doing that work would be an async function and as_bytes/as_str are synchronous currently.

sunshowers:
I'll put up an alternative version of this PR which breaks BC.

sunshowers:
Put up #542 -- converting this to a draft since I agree that that's definitely a better, more orthogonal API.

sunshowers:
Abandoned in favor of #556.

PR #543, opened 2023-01-05
Add `multiple-servers` example

jgallagher:


dropshot/examples/multiple-servers.rs
```diff
@@ -0,0 +1,377 @@
...
+
+    /*
+     * We want to shut down `server`, but it might be the very server handling
+     * this request! Move the shutdown onto a background task to allow this
+     * request to complete; otherwise, we deadlock.
+     *
+     * We can safely discard the result of `close()` because `main()` also gets
+     * it via the shutdown handle it received when `server` was created.
+     */
+    tokio::spawn(server.close());
```
smklein, line 374: After https://github.com/oxidecomputer/dropshot/pull/537 , I wonder if we should no longer make `close` wait for the server to terminate.

We could add a convenience `close_and_wait` for anyone that wants backwards compatibility?

jgallagher: Changing the behavior of `close()` seems a little dicey from a backwards compatibility standpoint, but I do see your point. A couple other ideas, although I'm not sure I like these all that much more:

1. Deprecate `close()` and add _two_ new methods: `close_and_wait()` + `close_nowait()`.
2. Change `close()` to take an argument specifying whether or not to wait: `close(CloseBehavior::Wait)` or somesuch. This would be backwards incompatible in a way that old clients would have to update instead of silently getting different behavior.

dropshot/examples/multiple-servers.rs
```diff
@@ -0,0 +1,377 @@
...
+            ConfigLogging::StderrTerminal { level: ConfigLoggingLevel::Info };
+        let log = config_logging
+            .to_logger(format!("example-multiserver-{name}"))
+            .map_err(|error| format!("failed to create logger: {}", error))?;
+
+        /*
+         * Build a description of the API.
+         *
+         * TODO: Could `ApiDescription` implement `Clone`, or could we pass an
+         * `Arc<ApiDescription>` instead?
```
davepacheco, line 222: This seems like a good idea.

davepacheco (approved):
This is a cool idea for an example!

I wonder if the server lifetime mechanics get in the way of showing what's involved in sharing state (i.e., it's not exactly the simplest example of sharing state), but then again, it's also pretty valuable that it's non-trivial.

dropshot/examples/multiple-servers.rs
jgallagher: Implementing `Clone` for `ApiDescription` is doable but kinda gnarly; it requires a bunch of manual `impl Clone for ...` to avoid requiring the server contexts to be cloneable (which is not necessary). OTOH, `Arc<ApiDescription>` seems problematic because of [ApiDescription::into_router()](https://github.com/oxidecomputer/dropshot/blob/01263bc1e7074bae8f6eee063d9a34981949b369/dropshot/src/api_description.rs#L897-L899). Maybe `ApiDescription` could hold two `Arc<_>`s as its fields (making it easily cloneable), and `into_router()` could return an `Arc<HttpRouter<Context>>`? I haven't looked at this very much - do you want me to poke at it more before merging this?

jgallagher:
> I wonder if the server lifetime mechanics get in the way of showing what's involved in sharing state (i.e., it's not exactly the simplest example of sharing state), but then again, it's also pretty valuable that it's non-trivial.

I had the same concern. We could certainly have a much simpler example that only showed a couple static servers sharing context, but the thing I really wanted to understand was how to start / stop servers dynamically, hence landing where I did.

A few options:

1. Keep this as-is
2. Keep this example, but also add a simpler one
3. Remove this example in favor of a simple one

Maybe 2 makes the most sense? Rename this `multiple-servers-dynamic` and add a much smaller `multiple-servers`?

dropshot/examples/multiple-servers.rs
davepacheco: Definitely not a blocker on this!

davepacheco:
> > I wonder if the server lifetime mechanics get in the way of showing what's involved in sharing state (i.e., it's not exactly the simplest example of sharing state), but then again, it's also pretty valuable that it's non-trivial.
> 
> I had the same concern. We could certainly have a much simpler example that only showed a couple static servers sharing context, but the thing I really wanted to understand was how to start / stop servers dynamically, hence landing where I did.
> 
> A few options:
> 
>     1. Keep this as-is
> 
>     2. Keep this example, but also add a simpler one
> 
>     3. Remove this example in favor of a simple one
> 
> 
> Maybe 2 makes the most sense? Rename this `multiple-servers-dynamic` and add a much smaller `multiple-servers`?

Any of those seems fine.  I'd probably just stick with what's here (option 1) unless you really feel like doing more.

jgallagher:
Cool, thanks. I'll open a couple followup issues (`ApiDescription` cloneability and whether `close()` should block).

PR #553, opened 2023-01-09
switch block-style to line-style comments

davepacheco:
This is modeled after oxidecomputer/omicron#770.  Paraphrasing from there what applies here as well.

---

Dropshot currently has a mix of commenting styles -- both block style (C style, /* ... */) and line style (// ...). It seems like this has been a distraction for folks, and the line style is more popular.  After the success of oxidecomputer/omicron#770 I used rustfmt's `normalize_comments` option to switch everything to line style.  I don't think we should enforce this with rustfmt because that requires an unstable option, which we know from experience is painful.  I don't expect this to be a big problem because I think line style tends to be the more common expectation.

When this lands, it may be annoying to merge with outstanding work.  I'm happy to help with that.  Based on last time, when this commit lands on "main", I think the right strategy is to sync up other PRs by:

1. First merge with the parent commit of this commit on "main".
2. Make sure you have no uncommitted local changes.
3. Copy this to `rustfmt-new.toml`:

```
# ---------------------------------------------------------------------------
# Stable features that we customize locally
# ---------------------------------------------------------------------------
max_width = 80
use_small_heuristics = "max"
edition = "2021"

# Temp unstable features
unstable_features = true
normalize_comments = true
```
4. Run `cargo +nightly fmt -- --config-path=rustfmt-new.toml`
5. Run `git commit -am 'premerge with style update'`
6. Merge with this commit on main.  This should pretty straightforward. If you have conflicts, they should be because you've changed comment lines. I think you pretty much want to take your version at this point (since step 3 will already have changed your comments to line-style).

### DIFF
--- a/dropshot/src/websocket.rs
+++ b/dropshot/src/websocket.rs
@@ -189,20 +189,20 @@ impl WebsocketUpgrade {
     /// ```
     /// #[dropshot::endpoint { method = GET, path = "/my/ws/endpoint/{id}" }]
     /// async fn my_ws_endpoint(
-    /// rqctx: std::sync::Arc<dropshot::RequestContext<()>>,
-    /// websock: dropshot::WebsocketUpgrade,
-    /// id: dropshot::Path<String>,
+    ///     rqctx: std::sync::Arc<dropshot::RequestContext<()>>,
+    ///     websock: dropshot::WebsocketUpgrade,
+    ///     id: dropshot::Path<String>,
     /// ) -> dropshot::WebsocketEndpointResult {
-    /// let logger = rqctx.log.new(slog::o!());
-    /// websock.handle(move |upgraded| async move {
-    /// slog::info!(logger, "Entered handler for ID {}", id.into_inner());
-    /// use futures::stream::StreamExt;
-    /// let mut ws_stream = tokio_tungstenite::WebSocketStream::from_raw_socket(
-    /// upgraded.into_inner(), tokio_tungstenite::tungstenite::protocol::Role::Server, None
-    /// ).await;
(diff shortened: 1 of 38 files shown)


david-crespo (approved):

dropshot/src/websocket.rs
```diff
@@ -184,45 +172,43 @@ impl Extractor for WebsocketUpgrade {
...
+    /// websock.handle(move |upgraded| async move {
+    /// slog::info!(logger, "Entered handler for ID {}", id.into_inner());
+    /// use futures::stream::StreamExt;
+    /// let mut ws_stream = tokio_tungstenite::WebSocketStream::from_raw_socket(
+    /// upgraded.into_inner(), tokio_tungstenite::tungstenite::protocol::Role::Server, None
+    /// ).await;
+    /// slog::info!(logger, "Received from websocket: {:?}", ws_stream.next().await);
+    /// Ok(())
+    /// })
+    /// }
```
ahl, line 206: lost the formatting

ahl (commented):
checked everywhere for a similar problem but just found this one
`git log --no-color -p main..  | grep '^-[^*]*$'`

dropshot/src/websocket.rs
davepacheco: Great catch!  For my future reference, I think it's because it wasn't formatted the way standard block-style comments are formatted (missing leading asterisks on those lines).  I fixed this by reverting this file, fixing the C-style formatting, then rerunning `cargo +nightly fmt -- --config-path=rustfmt-new.toml`.

PR #554, opened 2023-01-09
move extractors and generic schema utilities out of handler.rs

davepacheco:
I felt like handler.rs was getting a bit big and full of unrelated stuff.  Here I'm separating the extractors into a submodule with the eventual intent on putting different ones into different files within `extractors`.  This involved moving some functions used by both handler.rs and the new extractor module into schema_util.rs (analogous to http_util.rs).

I believe nothing has changed in this PR except for moving code from one file to another and adjusting imports / "mod" statements.

This depends on #553 -- will update the base branch once that one lands and I'll keep this in "draft" for the meantime.


davepacheco:
Sorry for the force push and general noise.

ahl (approved):
I really like this reorganization. I said "oooh" aloud when I saw that extractor was its own mod--like it a lot.

api_description.rs has a bunch of `j2oas_` prefixed functions; would those make sense to move into schema_util.rs? No or Not now are fine answers.

Issue #555, opened 2023-01-10
provide access to raw `hyper::Request` via an extractor


davepacheco:
Currently we provide every request handler with an `rqctx: Arc<RequestContext>`, which contains `rqctx.request: Arc<Mutex<hyper::Request>>`.

(Why? Access to the raw `hyper::Request` is useful as an escape hatch for stuff that Dropshot currently doesn't support.  The two use cases I know of are websockets (now supported) and streaming bodies (work in progress).  Although we expect these both to be supported in the not-too-distant future, this escape hatch remains useful.)

[But we've always known that this way of exposing it is janky.](https://github.com/oxidecomputer/dropshot/blob/f3cf19ecf50c9c938acd6f4ab86a01e6d1f8f851/dropshot/src/handler.rs#L81-L87)  There's only one thing with access to the request, so it's annoying that it's wrapped in an Arc+Mutex.  Worse: under #542 and related PRs, we're considering support for streaming request bodies.  These also need access to the `hyper::Request` (well, specifically, the body, which can only be gotten by consuming the request).  By allowing access to all handlers via the Arc+Mutex, it's possible to wind up with a deadlock, which means imposing implicit locking rules on consumers that we have no way to enforce.

Instead, let's remove the raw `hyper::Request` from the `RequestContext` and create a new extractor for getting it.  This way, if we can also ensure that no endpoint has two extractors that want access to the request (#556), there's no Arc+Mutex, so less awkwardness and no locking issues.

Note that there have been two attempts at this:

* #106 kind of ran out of steam after I dropped the ball.
* #548 prototypes this as well.  After some discussion, I'm working on a similar prototype that would enforce the new constraints at compile-time rather than runtime.

I've got a prototype of shared/exclusive extractors in #556 that's looking promising.  With that, I think we can make `TypedBody`, `UntypedBody`, and a new `RawRequest` into `ExclusiveExtractors`, which means you can only have one of these as an argument to your endpoint function.

PR #556, opened 2023-01-10
make it a compile error to accept more than one body extractor

davepacheco:
See #555 for motivation.  We currently verify at runtime that you don't have two extractors that both try to read the body.  We could make this a compile-time error instead.

This currently depends on #554 so I'm going to leave this as "draft".  Once #554 lands, I will update the base branch of this PR.  We may also want to wait for #557 (which implements #555), and then also wait for a prototype of that in Omicron, before landing this.  Until we have all that, this could all be a bust.

Status:

- [x] get consensus that this seems promising
- [x] confirm it's not a disaster in consumers like Omicron.  (It was very simple.  There's just one custom extractor and it's easily converted to a SharedExtractor.)  I'm happy to check any other consumers people would like.
- [x] get consensus on the rest of this path (#557 at least, and maybe #558)
- [x] polish up this PR
    - [x] changelog update
    - [x] fix error messages so they're as useful as they were before when your endpoint function arg isn't an extractor
    - [x] make sure we have tests
    - [x] any other XXXs
- [x] fix up websocket argument order (and changelog for that)


dropshot_endpoint/src/lib.rs
```diff
@@ -935,14 +935,6 @@ mod tests {
             const _: fn() = || {
                 struct NeedRequestContext(<Arc<RequestContext<std::i32> > as dropshot::RequestContextArgument>::Context) ;
             };
-            const _: fn() = || {
-                fn need_extractor<T>()
-                where
-                    T: ?Sized + dropshot::Extractor,
```
ahl, line 941: could this change to `RequestExtractor`?

ahl: ... or maybe `SharedExtractor`?

dropshot_endpoint/src/lib.rs
```diff
@@ -173,7 +173,8 @@ fn do_channel(
         ChannelProtocol::WEBSOCKETS => {
             // here we construct a wrapper function and mutate the arguments a bit
```
ahl, line 174:
```suggestion
            // Here we construct a wrapper function and mutate the arguments a bit
```

ahl (commented):
looks good

dropshot/src/extractor/common.rs
```diff
@@ -0,0 +1,198 @@
...
+/// Extractors that require exclusive access to the underyling `hyper::Request`
+///
+/// These extractors usually need to read the body of the request or else modify
+/// how the server treats the rest of it (e.g., websocket upgrade).  There may
+/// be at most one of these associated with any request.
+#[async_trait]
+pub trait ExclusiveExtractor: Send + Sync + Sized {
+    /// Construct an instance of this type from a `RequestContext`.
+    async fn from_request<Context: ServerContext>(
+        rqctx: &RequestContext<Context>,
```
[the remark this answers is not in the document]
davepacheco: Counter-intuitively, this works with just `&RequestContext<Context>`.  That's because what exclusive extractors need mutable access to is the `hyper::Request`, which is in the `RequestContext` but wrapped with `Arc<Mutex<...>>`.  So you can take the lock to access a `&mut hyper::Request`.

After #558, the `hyper::Request` won't even be in the `RequestContext` any more.

So given that it's not necessary to have a `&mut RequestContext` and that the caller is going to pass the same `RequestContext` to the request handler function after calling this one, I prefer giving the extractor just a `&RequestContext`.  Hopefully that all makes sense.

dropshot/src/extractor/common.rs
```diff
@@ -0,0 +1,198 @@
+// Copyright 2022 Oxide Computer Company
```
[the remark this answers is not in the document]
davepacheco: Done in 8ee3aa9a41f830618622f65992a5fc17989a69be.

dropshot/tests/fail/bad_endpoint3.stderr
```diff
@@ -1,32 +1,3 @@
-error[E0277]: the trait bound `String: Extractor` is not satisfied
-  --> tests/fail/bad_endpoint3.rs:17:12
-   |
-17 |     param: String,
-   |            ^^^^^^ the trait `Extractor` is not implemented for `String`
```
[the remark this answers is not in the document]
davepacheco: Totally.  Since you're on board with this general idea, the next step is to figure out how to ensure good error messages when (1) an argument is not an extractor at all, or (2) you attempt to use two exclusive extractors.  I think (1) is pretty easy because I made `SharedExtractor` impl `ExclusiveExtractor`, so we can change the existing check to check `ExclusiveExtractor`.  I have not yet figured out a good way to do (2) but Rain suggested an approach I'm going to look at.

dropshot_endpoint/src/lib.rs
davepacheco: I think it could change to `ExclusiveExtractor` (shared extractors can be treated as exclusive, but not the other way around).  (see my other comment -- I do intend to restore a helpful message for this case, one way or another)

dropshot_endpoint/src/lib.rs
davepacheco: Done in fb3183179de98c070dab4dc17be9a177f58e64d3.

dropshot/src/extractor/common.rs
```diff
@@ -0,0 +1,198 @@
...
+        body_content_type: ApiEndpointBodyContentType,
+    ) -> ExtractorMetadata {
+        <S as SharedExtractor>::metadata(body_content_type)
+    }
+}
+
+/// Top-level extractor for a given request
+///
+/// During request handling, we wind up needing to call a function with a
+/// variable number of arguments whose types are all extractors (either
```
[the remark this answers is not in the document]
ahl: looks good

dropshot/tests/fail/bad_endpoint3.stderr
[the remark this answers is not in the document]
davepacheco: Done.  The basic approach is to verify that:

- the first argument is a `RequestContext` (like we did before)
- the last argument is an `ExclusiveExtractor` (slight variation on what we did before)
- all other arguments are `SharedExtractor`

This has the side effect of verifying that you don't have two exclusive extractors because they don't impl SharedExtractor.

The naming of the tests is a little weird.  I kept the existing dense integer naming but it's a little confusing that test 19 is a slight variant of test 3.  (That is, the behavior tested by the old bad_endpoint3.rs is now tested by a combination of bad_endpoint3.rs and bad_endpoint19.rs.)  The alternatives seemed like separating into 3a and 3b or renaming everything to be more semantic. 🤷

dropshot_endpoint/src/lib.rs
```diff
@@ -219,6 +220,20 @@ fn do_channel(
...
+            // `WebsocketUpgrade` argument that we put in its place) appears
+            // last.  We replaced the type above, but now we need to put it in
+            // the right spot.
+            sig.inputs = {
+                let mut input_pairs =
+                    sig.inputs.clone().into_pairs().collect::<Vec<_>>();
+                let second_pair = input_pairs.remove(1);
+                input_pairs.push(second_pair);
+                input_pairs.into_iter().collect()
+            };
```
davepacheco, line 235: I'm curious what people think about this.  The comment basically summarizes it: we previously said that a `WebsocketConnection` has to be the first extractor.  As part of this PR, I decided that exclusive extractors (which would include this one) have to be _last_.  I did that because I think it's natural for the extractors to appear in the order they look at parts of the HTTP request, and exclusive extractors generally look at the request body (the last thing).

There's not actually a conflict here because this macro already translates the user's argument list into what the `endpoint` macro expects, so we can shuffle the arguments around like this.  But it's kind of annoying that we have to do this, and a user might notice the weirdness that their `WebsocketConnection` has to appear first, but other exclusive extractors have to appear last.

We could:

- do nothing (what we're doing here).  Downside: this ugliness, and maybe slightly confusing for people?
- make `WebsocketConnection` appear last.  Downside: breaking change, though trivial to fix.
- make exclusive extractors appear first instead.  Downside: less intuitive for the much more common case

Thoughts @ahl @lifning ?

davepacheco:
I've sync'd this change up with "main", cleaned up the tests and other XXXs, and written migration instructions in the changelog.  I think it's ready to go except maybe for [this one question](https://github.com/oxidecomputer/dropshot/pull/556/files#r1068543395).

@ahl Unfortunately I think the way I sync'd this up will make it annoying to do an incremental review.  Sorry.  Let me know if I can help.

dropshot_endpoint/src/lib.rs
sunshowers: I like making `WebsocketConnection` appear last -- it's a breaking change but I think it's justified by the overall direction of the PR.

PR #557, opened 2023-01-10
remove `hyper::Request` from `RequestContext`

### LINKED ISSUE #555: provide access to raw `hyper::Request` via an extractor

davepacheco:
Depends on #556.  Leaving "draft".  This is still prototype-quality.  It needs tests, docs, etc.

Fixes #555.

Status:

- [x] get consensus that this seems promising
- [x] confirm it's not a disaster in consumers like Omicron (see oxidecomputer/omicron#2156)
- [x] maybe: get consensus on #558
- [x] polish up this PR
    - [x] changelog update
    - [x] any other XXXs
- [x] land #556 first


dropshot/src/dtrace.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2023 Oxide Computer Company
```
davepacheco, line 1: This was moved (but otherwise unchanged) from dropshot/src/lib.rs

CHANGELOG.adoc
```diff
@@ -25,7 +25,20 @@ There are a number of breaking changes in this release but we expect they will b
...
-3. If you have your own type that impls `Extractor`, you will also need to change the type signature of the `from_request` method to accept a `&RequestContext<T>` instead of `Arc<RequestContext<T>>`.  (This should not be a problem unless your extractor was hanging on to a reference via the Arc.  We don't know a reason this would be useful.  If you were doing this, please reach out.  You can probably instead copy whatever information you need out of the `RequestContext` instead.)
+3. If you have your own type that impls `Extractor`, you will also need to change the type signature of the `from_request` method to accept a `&RequestContext<T>` instead of `Arc<RequestContext<T>>`.  (This should not be a problem unless your extractor was hanging on to a reference via the Arc.  We don't know a reason this would be useful.  If you were doing this, please reach out.  You can probably copy whatever information you need out of the `RequestContext` instead.)
```
ahl, line 28: please file an issue?

ahl (approved):
love it

PR #558, opened 2023-01-10
remove Arc around RequestContext

davepacheco:
This depends on #557.  Leaving in "draft".  This is a prototype.

This change removes the `Arc` around `RequestContext` in all of the request handler functions.  This is something I've been meaning to revisit for a while but it came up while doing #557, etc.

On the one hand, I think this is a good change in isolation: nobody else has a reference to this object, and in fact the whole point of this object is to encapsulate _for the request handler_ whatever Dropshot wants to pass to it _for this request_.  So it doesn't make sense for Dropshot to hold its own reference to it.  Dropshot can keep copies or references to whatever things inside the RequestContext that it wants.  It's simpler for everybody to get rid of this `Arc` and it makes it less likely that people will accidentally hang onto copies of it.

The main downside is that this change has a big blast radius for consumers -- literally every endpoint function has to change.  But the change is trivial and any mistakes would be caught at compile-time.  I used this vim command: `%s/Arc<RequestContext<\([^>]*\)>>/RequestContext<\1>/gc`.  Another option would be to explore supporting both forms, but that feels more confusing going forward.

I also considered making it `&RequestContext`.  I'm not sure what the right lifetime for that would be, if you wanted to make it explicit.  And again, my thought was: dropshot has created this object for you, so why not let you have it?  In retrospect I wish most of the fields themselves were references (or that we only exposed references).  Maybe if we did that _and_ only gave you `&RequestContext<'a>`, then we could actually prevent leakage of these fields without you explicitly cloning them.  That'd be a bigger change for consumers using these fields.

Thoughts? (CC @ahl)


davepacheco:
I updated Omicron to this change in oxidecomputer/omicron#2156.  It's predictably an annoying amount of noise, but it was mostly mechanical and we can document how to do it (and offer to help).

ahl (commented):
looks good

PR #560, opened 2023-01-11
move some api_description functions into new schema_util

davepacheco:
[As suggested here](https://github.com/oxidecomputer/dropshot/pull/554#pullrequestreview-1243181779).


ahl (approved):
mucho gusto!

PR #570, opened 2023-01-19
CI builds should test with and without features

davepacheco:
Our CI builds do not test with the "usdt-probes" feature enabled.  It's also not enabled by default so it would be easy to not notice while testing locally if you broke it.

This change:

- updates the Helios build to test with all features
- updates the GitHub Actions builds to test all OS's twice: once without features and once with.  This might be overkill but it was easy to do and covers everything.

We're not running clippy with features right now but that seems less critical.

### DIFF
--- a/.github/buildomat/jobs/build-and-test.sh
+++ b/.github/buildomat/jobs/build-and-test.sh
@@ -14,7 +14,7 @@ cargo --version
 rustc --version
 
 banner build
-ptime -m cargo build --locked --all-targets --verbose
+ptime -m cargo build --all-features --locked --all-targets --verbose
 
 banner test
-ptime -m cargo test --locked --verbose
+ptime -m cargo test --all-features --locked --verbose
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -40,6 +40,10 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, windows-2022, macos-12]
+        features: [ all, default ]
+        include:
+          - features: all
+            feature_flags: --all-features
     steps:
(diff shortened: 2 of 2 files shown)


davepacheco:
I've verified that the changes seemed to do what I expect:

- The buildomat build ran with `--all-features`.
- We have "all" and "default" feature builds for the other three OSes.  I checked that the "default" build for ubuntu passes no feature-related flags and the "all" build passes `--all-features`.

(I wouldn't mind another check.)

In order to land this I'll need to adjust the require checks for the repo, since the previously-required checks no longer exist.  I won't do that until I'm ready to land it.

.github/buildomat/jobs/build-and-test.sh
```diff
@@ -14,7 +14,7 @@ cargo --version
 rustc --version
 
 banner build
-ptime -m cargo build --locked --all-targets --verbose
+ptime -m cargo build --all-features --locked --all-targets --verbose
```
bnaecker, line 17: So this looks like it'll only test on illumos with the `usdt-probes` feature, whereas the GitHub workflow below tests both with and without that feature. Is that intentional? Did you want another run of `cargo build` / `cargo test` but without `--all-features`?

bnaecker (commented):
The GitHub workflow looks good, but it looks like the Buildomat job tests only _with_ the probes unless I'm missing something.

.github/buildomat/jobs/build-and-test.sh
davepacheco: Yeah, that's accurate.  It's not a strong feeling.  It just felt like I was already adding a lot more tests and I wasn't sure how necessary it was, given that we have the non-feature tests on other systems.

bnaecker (approved):

PR #577, opened 2023-01-29
Simplify tls.type example in README

epompeii:
All of the other examples are set to just a single value, and then all potential values are enumerated as a "Valid values include ..." statement in the description.


davepacheco:
Thanks1

PR #586, opened 2023-02-08
add operation_id to scalar validation message

ahl:


zephraph:
Thank you so much! This is a huge QoL change.

PR #597, opened 2023-02-27
Manually outline ApiDescription::register

steveklabnik:
This function often gets called in generic contexts, and so a *lot* of monomorphization happens. By making this small change, you can help the compiler not do as much work.

In my tests, this results in five seconds being knocked off of Omicron's build times.

### DIFF
--- a/dropshot/src/api_description.rs
+++ b/dropshot/src/api_description.rs
@@ -279,11 +279,21 @@ impl<Context: ServerContext> ApiDescription<Context> {
     {
         let e = endpoint.into();
 
-        self.validate_tags(&e)?;
-        self.validate_path_parameters(&e)?;
-        self.validate_named_parameters(&e)?;
+        // manually outline, see https://matklad.github.io/2021/09/04/fast-rust-builds.html#Keeping-Instantiations-In-Check
+        fn _register<C: ServerContext>(
+            s: &mut ApiDescription<C>,
+            e: ApiEndpoint<C>,
+        ) -> Result<(), String> {
+            s.validate_tags(&e)?;
+            s.validate_path_parameters(&e)?;
+            s.validate_named_parameters(&e)?;
 
-        self.router.insert(e);
+            s.router.insert(e);
+
+            Ok(())
+        }
+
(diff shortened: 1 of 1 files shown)


davepacheco:
Wow -- that's impressive.  Do you happen to have a link about this behavior or why this sort of change helps?  I'm also really curious how you found it.

In terms of the PR, would you mind just adding a comment (just thinking a sentence or two) explaining this?  I'm just worried a future reader will try to simplify this by removing the extra function, not realizing why it's there.

steveklabnik:
As it turns out, https://matklad.github.io/2021/09/04/fast-rust-builds.html#Keeping-Instantiations-In-Check basically answers all of your questions, including how I found it, so I've added it in a comment :)

(I have been meaning to write about this myself for years, see the bottom of https://steveklabnik.com/writing/are-out-paramters-idiomatic-in-rust)

steveklabnik:
(Oh and to put some small numbers on this, [this function](https://github.com/oxidecomputer/omicron/blob/9a86346c45a3d7a1512144565ed93ee1ae4ff6a4/nexus/src/external_api/http_entrypoints.rs#L81) ends up generating 12,000 lines of LLVM-IR. The next largest function generates 6,000. I didn't actually double check the LLVM line length right this moment, eventually I will have a more comprehensive report about omicron compile times but I'm not ready to share all of that just yet.)

davepacheco:
Thanks!

davepacheco (approved):

ahl:
This is awesome. The `register` method is absolutely not performance sensitive at all. Would it make sense to avoid making it generic e.g. by accepting an `ApiEndpoint<Context>`? We could change the `#endpoint` macro to invoke the `into()`. I could imagine that your change already reaps the bulk of the benefit, but just curious. Thanks!

steveklabnik:
> Would it make sense to avoid making it generic e.g. by accepting an ApiEndpoint<Context>?

I may investigate doing that as well, it may help, it may not, we'll see! This is the kind of thing where you generally need to do it and measure, rather than guess.

PR #598, opened 2023-02-28
add path information to body deserialization failure

ahl:
for https://github.com/oxidecomputer/omicron/issues/1848


dropshot/src/extractor/body.rs
```diff
@@ -81,20 +81,28 @@ where
...
-        (UrlEncoded, UrlEncoded) => serde_urlencoded::from_bytes(&body)
-            .map_err(|e| {
+
+    let content = match (expected_content_type, body_content_type) {
+        (Json, Json) => {
+            let jd = &mut serde_json::Deserializer::from_slice(&body);
+            serde_path_to_error::deserialize(jd).map_err(|e| {
+                HttpError::for_bad_request(
+                    None,
+                    format!("unable to parse JSON body: {}", e),
```
david-crespo, line 91: If we're going this far, I think "unable to parse JSON body" should be changed as well. IMO from the point of view of an API consumer, the fact that the error arises as a failure to parse is a bit of an implementation detail. From the API's point of view the point is that it's an _invalid_ body, just as negative disk size would be invalid, even though it's technically parseable.

ahl: I see you point. What would you prefer to see here?

david-crespo: “Invalid request body: …” seems ok. I think in the case where it’s not even JSON, you’ll get a reasonably suggestive message after that about “expected { at line 1”, right?

davepacheco (approved):

PR #617, opened 2023-03-21
Add support for a new StreamingBody extractor

sunshowers:
Per RFD 353, add support for a new extractor that can provide data as a
stream of Bytes chunks. This means that large request payloads won't
have to buffer everything into a single, contiguous chunk of memory.

Also add some examples showing how to use the extractor to operate on
chunks in a streaming fashion. (There are also other potential examples
we can add here, such as compressing and/or decompressing data on the
fly). I haven't implemented those here.

Finally, switch over the other extractors to be implemented on top of
`StreamingBody`. In effect, the other extractors just provide a
differently-shaped API  on top of `StreamingBody`.

This PR doesn't implement support for the other determination in RFD
353, which is to allow some endpoints' max payload sizes to be set to a
higher limit. That's in #618.

### DIFF
--- a/dropshot/src/extractor/body.rs
+++ b/dropshot/src/extractor/body.rs
@@ -231,26 +231,6 @@ impl StreamingBody {
         Self { body, cap }
     }
 
-    /// Converts `self` into a [`BytesMut`], buffering the entire response in memory.
-    ///
-    /// If payloads are expected to be large, consider using [`Self::into_stream`] to
-    /// avoid buffering in memory if possible.
-    ///
-    /// # Errors
-    ///
-    /// Returns an [`HttpError`] if any of the following cases occur:
-    ///
-    /// * A network error occurred.
-    /// * `request_body_max_bytes` was exceeded for this request.
-    pub async fn into_bytes_mut(self) -> Result<BytesMut, HttpError> {
-        self.into_stream()
-            .try_fold(BytesMut::new(), |mut out, chunk| {
-                out.put(chunk);
-                futures::future::ok(out)
-            })
-            .await
-    }
-
     /// Converts `self` into a stream.
     ///
     /// The `Stream` produces values of type `Result<Bytes, HttpError>`.
@@ -266,9 +246,7 @@ impl StreamingBody {
     ///
     /// Buffer a `StreamingBody` in-memory, into a
     /// [`BufList`](https://docs.rs/buf-list/latest/buf_list/struct.BufList.html)
-    /// (a segmented list of [`Bytes`] chunks). This is similar to
-    /// [`Self::into_bytes_mut`], except it avoids copying memory into a single
-    /// large allocation.
+    /// (a segmented list of [`Bytes`] chunks).
     ///
     /// ```
     /// use buf_list::BufList;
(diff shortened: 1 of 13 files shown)


dropshot/tests/test_demo.rs
```diff
@@ -727,6 +730,57 @@ async fn test_untyped_body() {
...
+        .unwrap();
+    let json: DemoStreaming = read_json(&mut response).await;
+    assert_eq!(json.nbytes, 1024);
+
+    // Error case: body too large.
+    let big_body = vec![0u8; 1025];
+    let error = client
+        .make_request_with_body(
+            Method::PUT,
+            "/testing/untyped_body",
```
davepacheco, line 772: Was this supposed to be `streaming_body`?

dropshot/src/extractor/body.rs
```diff
@@ -184,38 +189,246 @@ impl UntypedBody {
...
+    /// #    let body = StreamingBody::__from_bytes(bytes::Bytes::from("foobar"));
+    /// #    let mut writer = vec![];
+    /// #    write_all(body, &mut writer).await.unwrap();
+    /// #    assert_eq!(writer, &b"foobar"[..]);
+    /// # }
+    /// ```
+    ///
+    /// ---
+    ///
+    /// An alternative way to write data to an `AsyncWrite`, using
```
davepacheco, line 331: What's the advantage of this?  (just thinking: did we pull in an extra dependency just to add an example that looks equivalent to the previous example?  That's not bad but it'd be useful to know why this is better.)

dropshot/src/extractor/body.rs
```diff
@@ -184,38 +189,246 @@ impl UntypedBody {
...
+    /// If payloads are expected to be large, consider using [`Self::into_stream`] to
+    /// avoid buffering in memory if possible.
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`HttpError`] if any of the following cases occur:
+    ///
+    /// * A network error occurred.
+    /// * `request_body_max_bytes` was exceeded for this request.
+    pub async fn into_bytes_mut(self) -> Result<BytesMut, HttpError> {
```
davepacheco, line 245: What do you think about not exposing this publicly?  For the reason you mentioned it's a (small) footgun, and if you want that you can use `UntypedBody` already, right?

davepacheco (commented):
Nice!

dropshot/src/extractor/body.rs
sunshowers: Yeah, on second thought I think what we should do is recommend you use `UntypedBody`, and let you convert *that* into a `BytesMut`. Can do that in a followup.

dropshot/src/extractor/body.rs
sunshowers: That is a very good question. I figured we'd want to include this example for completeness, but I had a chat with Eliza Weisman who says that the tokio codec stuff is mostly historical:

1. Codecs were originally a way to guarantee cancel safety, but `write_all_buf` is just a much clearer way to do that -- if it fails or gets canceled, it still advances the buffer with the amount written by.
2. Back in the day the tokio devs assumed people would use `Sink` a lot more. That turned out not to be the case so much.

These days there's no reason for anyone to use codecs or `Sink`, really, so I'm going to just remove this example.

sunshowers: Removed this example.

dropshot/tests/test_demo.rs
sunshowers: Whoops yeah! Fixed.

PR #619, opened 2023-03-21
improve style of macro generation

ahl:
noticed this when doing some syn 2.0 cleanup; I didn't know how to do this when I wrote the code originally.


sunshowers (approved):
lgtm!

PR #651, opened 2023-04-17
include address of remote peer in RequestInfo

### LINKED ISSUE #316: RequestContext should expose remote address of connection
In #148 (for #46) we added the remote address information to the logs emitted while handling a request.  In addition to this, we should stash the remote address on the `RequestContext` object so that request handlers can examine the remote IP address from which the request originates.

jclulow:
This is an attempt to fix #316.

### DIFF
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -96,22 +96,21 @@ pub struct RequestInfo {
     uri: http::Uri,
     version: http::Version,
     headers: http::HeaderMap<http::HeaderValue>,
+    remote_addr: std::net::SocketAddr,
 }
 
-impl<B> From<&hyper::Request<B>> for RequestInfo {
-    fn from(request: &hyper::Request<B>) -> Self {
+impl RequestInfo {
+    pub(crate) fn new<B>(
+        request: &hyper::Request<B>,
+        remote_addr: std::net::SocketAddr,
+    ) -> Self {
         RequestInfo {
             method: request.method().clone(),
             uri: request.uri().clone(),
             version: request.version(),
             headers: request.headers().clone(),
+            remote_addr,
         }
-    }
-}
-
-impl<B> From<hyper::Request<B>> for RequestInfo {
-    fn from(request: hyper::Request<B>) -> Self {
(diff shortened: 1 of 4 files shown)


dropshot/src/handler.rs
```diff
@@ -132,6 +131,10 @@ impl RequestInfo {
         &self.headers
     }
 
+    pub fn remote_addr(&self) -> &std::net::SocketAddr {
+        &self.remote_addr
+    }
```
davepacheco, line 136:
```suggestion
    pub fn remote_addr(&self) -> std::net::SocketAddr {
        self.remote_addr
    }
```

`SocketAddr` returns `Copy` so I think it's worthwhile to just return it directly rather than a reference.  That'll also be consistent with [`local_addr()` on the Server](https://docs.rs/dropshot/latest/dropshot/struct.HttpServer.html#method.local_addr).

davepacheco (commented):
Thanks for doing this.  Would you mind just adding a test somewhere?  (I would hope this could be just a few lines in some existing test that just verifies that the address returned by the new function matches the address of the client that we used to make the request.  I know it'll be loopback anyway so it's not the most interesting test.  But the port numbers should match up and that'll make sure we're grabbing the right thing, both now and if things change in the future.)

jclulow:
> Thanks for doing this. Would you mind just adding a test somewhere? (I would hope this could be just a few lines in some existing test that just verifies that the address returned by the new function matches the address of the client that we used to make the request. I know it'll be loopback anyway so it's not the most interesting test. But the port numbers should match up and that'll make sure we're grabbing the right thing, both now and if things change in the future.)

I added a test!  Let me know if it is what you had in mind, or if you would like something different.

davepacheco (approved):
This looks great.  Thanks.  (Too bad we can't get the client port out, but I guess it's possible it's not determined until we make the request and even then could be transient.)

I'd also like to add a changelog entry.  I did that in #652.  That PR targets *this* branch right now.  So we could either land that into this PR and then land this PR, or else we could land this PR as-is and then land that PR into "main".  Either way's fine with me.

PR #660, opened 2023-04-28
Add required-without-page_token information to pagination extension

ahl:
Previously `x-dropshot-pagination` was simply the value `true`. Now it's an object with a field `required` that is an array of strings that tell us what query parameters are required to fetch the first page (i.e. when `page_token` is not specified).

This also cleans up some stray doc nits.

### DIFF
--- a/dropshot/src/lib.rs
+++ b/dropshot/src/lib.rs
@@ -392,7 +392,7 @@
 //!
 //! ```json
 //! {
-//!     "page_token": "abc123...",
+//!     "next_page": "abc123...",
 //!     "items": [
 //!         {
 //!             "name": "aardvark",
@@ -413,7 +413,7 @@
 //! The page token `"abc123..."` is an opaque token to the client, but typically
 //! encodes the scan parameters and the value of the last item seen
 //! (`"name=badger"`).  The client knows it has completed the scan when it
-//! receives a response with no `page_token` in it.
+//! receives a response with no `next_page` in it.
 //!
 //! Our API endpoint can also support scanning in reverse order.  In this case,
 //! when the client makes the first request, it should fetch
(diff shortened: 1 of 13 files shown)


dropshot/src/lib.rs
```diff
@@ -507,15 +507,35 @@
 //! this ought to work, but it currently doesn't due to serde_urlencoded#33,
 //! which is really serde#1183.
 //!
-//! ### DTrace probes
+//! Note that any parameters defined by `MyScanParams` are effectively encoded
+//! into the page token and need not be supplied invocations when `page_token`
```
davepacheco, line 511:
```suggestion
//! into the page token and need not be supplied with invocations when `page_token`
```

dropshot/src/lib.rs
```diff
@@ -507,15 +507,35 @@
...
+//! is specified. That is not the case for required parameters defined by
+//! `MyExtraQueryParams`--those must be supplied on each invocation.
+//!
+//! ### OpenAPI extension
+//!
+//! In generated OpenAPI documents, Dropshot adds the `x-dropshot-paginated`
+//! extension to paginated operations. The value is currently a structure
+//! with this format:
+//! ```json
```
davepacheco, line 520: Does this render right without a newline here?

davepacheco (approved):
Do I grok correctly that there's no impact on dropshot consumers here (unless they're using the openapi extension)?

dropshot/src/lib.rs
ahl: image

Yes it looks good to me.

ahl:
> Do I grok correctly that there's no impact on dropshot consumers here (unless they're using the openapi extension)?

Yes; that's accurate.

PR #665, opened 2023-05-05
fix #664: Generalize dtrace support to allow other tracing tools

### LINKED ISSUE #664: Generalize dtrace support for other tracing tools
Hi! I really like the dtrace support built into dropshot. Obviously there are personal and historical reasons why dropshot has first-class support for dtrace ;) But I'd like to include other tracing mechanisms in my Dropshot server. 

It would be great if the dtrace support wasn't a "special case" built into the framework, but rather, the framework allowed a "tracing trait" which dtrace impls. Then users could implement that trait for

- sentry
- jaegar
- etc

@jessfraz and I would be happy to open a PR for this to explore a prototype, if y'all agree that this general idea is desirable :)

adamchalmers:
Currently on main, dropshot supports tracing functions with dtrace probes. Some users, including KittyCAD, would like to add other forms of tracing (e.g. sentry or their own custom tooling). 

This PR adds a new trait `Tracing` which matches the API used by dropshot's dtrace support. It adds two implementations of that trait:

- A dtrace implementation, which runs the same functionality as the current dtrace feature flag
    - This is only available behind the existing dtrace Cargo feature flag for the crate
- A no-op implementation

Users can specify which tracing they want by passing a `Tracing` value into the Dropshot server starters (see updated examples). 

**Changes to user experience**
 - Users now have to pass a value that implements `Tracing` when creating a Dropshot server. This is a minor breaking change -- users can either pass the dtrace implementation, or the no-op implementation, or write their own implementation and use that.

**Implementation changes**
 - This generalizes a lot of Dropshot's core server structs, which were previously generic over `C: ServerContext`. This PR further generalizes many (not all) of them over `Tr: Tracing`. 

**Next steps**
I'm going to experiment with integrating this into some KittyCAD services, to make sure the `Tracing` trait is right for us. I'm opening this PR to Dropshot now, to see if y'all have any objections to the overall idea, or notes about my implementation. I'm going to mark this PR as a draft until I confirm that it suits our needs within KittyCAD first :)

### DIFF
--- a/dropshot/examples/basic.rs
+++ b/dropshot/examples/basic.rs
@@ -2,6 +2,7 @@
 //! Example use of Dropshot.
 
 use dropshot::endpoint;
+use dropshot::tracing::Noop;
 use dropshot::ApiDescription;
 use dropshot::ConfigDropshot;
 use dropshot::ConfigLogging;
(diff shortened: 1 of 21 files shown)


adamchalmers:
I don't think we're going to need this, "use plain old function calls" is working out so far.

PR #670, opened 2023-05-12
Do not use chrono's "oldtime" feature

adamchalmers:
This feature is enabled by default, but 
a. isn't needed
b. introduces [RUSTSEC-2020-0071](https://rustsec.org/advisories/RUSTSEC-2020-0071)

So, let's not use it.


adamchalmers:
Taggin @davepacheco for review and to authorize me for running CI :)

davepacheco:
Thanks.  I'm happy to +1 and authorize the buildomat CI but it looks from the ubuntu job that already ran like this needs a Cargo.lock update as well.

adamchalmers:
OK, Cargo.lock updated.

davepacheco (approved):

PR #701, opened 2023-06-12
Add server config option to spawn handlers detached from client disconnect behavior

jgallagher:
Adds a new `default_handler_disposition` field to both `DropshotConfig` and `ServerConfig`. The default value for this (in `DropshotConfig`'s `Default` impl) is `HandlerDisposition::CancelOnDisconnect`, which matches the current behavior of dropshot: if a client disconnects, the future returned by the endpoint handler is dropped (and therefore cancelled). If the config is instead `HandlerDisposition::DetachFromClient`, all endpoint handler futures are `tokio::spawn()`'d, and they will therefore always run to completion.

I phrased this field as `default_handler_disposition` under the assumption that we'd (eventually) provide a way for individual endpoints to override this choice, but am very much not wed to the naming here - any suggestions are welcome.


dropshot/src/config.rs
```diff
@@ -49,6 +49,32 @@ pub struct ConfigDropshot {
     pub bind_address: SocketAddr,
     /// maximum allowed size of a request body, defaults to 1024
     pub request_body_max_bytes: usize,
+    /// Default behavior for HTTP handler functions with respect to clients
+    /// disconnecting early.
+    pub default_handler_disposition: HandlerDisposition,
```
davepacheco, line 54:
```suggestion
    pub default_handler_disposition: HandlerDisposition,
```

Thinking out loud: how about: `on_client_disconnect` with the enum being called `OnClientDisconnect` with variants `CancelHandler` and `DetachHandler`?  I like being more concrete than the word "disposition".  On the other hand, it goes further down the confusing path of suggested we detach _when_ the client disconnects.

What about: `default_handler_task_mode` with `HandlerTaskMode` with variants `Detached` and `CancelOnDisconnect`?

dropshot/src/config.rs
```diff
@@ -78,6 +104,7 @@ impl Default for ConfigDropshot {
         ConfigDropshot {
             bind_address: "127.0.0.1:0".parse().unwrap(),
             request_body_max_bytes: 1024,
+            default_handler_disposition: HandlerDisposition::CancelOnDisconnect,
```
davepacheco, line 107:
```suggestion
            // The default is `CancelOnDisconnect` for compatibility.  This is arguably a less safe
            // default.  After we've got some operational experience with `DetachFromClient`, we may
            // want to make that the default.
            default_handler_disposition: HandlerDisposition::CancelOnDisconnect,
```

dropshot/src/api_description.rs
```diff
@@ -33,7 +34,7 @@ use std::collections::HashSet;
 #[derive(Debug)]
 pub struct ApiEndpoint<Context: ServerContext> {
     pub operation_id: String,
-    pub handler: Box<dyn RouteHandler<Context>>,
+    pub handler: Arc<dyn RouteHandler<Context>>,
```
davepacheco, line 37: Why did this change to `Arc`?

davepacheco (commented):
For a change like this, I think it'd be useful to test-integrate it into Omicron before we land it here.  What do you think?  (I'm afraid you're going to wind up undoing some changes I did recently to remove `..Default::default()` from a whole bunch of `ConfigDropshot` structures.  Sorry about that.)

For the record: I confirmed that [the current behavior is expected from hyper](https://github.com/hyperium/hyper/issues/2728).  I wanted to confirm that because it does seem counter-intuitive to a bunch of us and if we're going so far out of our way to do things differently, I wanted to make sure we weren't just holding it wrong.  (I'm not sure _why_ it's the behavior and I remain frustrated that cancellation appears to be undocumented in the canonical Rust and async/await references.)

dropshot/tests/test_config.rs
```diff
@@ -345,3 +380,361 @@ async fn test_config_bind_address_https_buffer() {
...
+    };
+    let bind_port = 12221;
+
+    // Make sure there is not currently a server running on our expected
+    // port so that when we subsequently create a server and run it we know
+    // we're getting the one we configured.
+    let client = test_config.make_client();
+    let error = client.get(test_config.make_uri(bind_port)).await.unwrap_err();
+    assert!(error.is_connect());
+
```
davepacheco, line 510: I think this stuff (and probably other stuff in this test) is not necessary here.  It's used in the tests that are actually testing binding to specific ports.  You're not testing that here, so I would specify port 0 (let the kernel decide) and leave a lot of this out.

dropshot/tests/test_config.rs
```diff
@@ -345,3 +380,361 @@ async fn test_config_bind_address_https_buffer() {
...
+    let result = client_task.await.unwrap_err();
+    assert!(result.is_cancelled());
+
+    // The handler _should still be running_; send it a message to continue. If
+    // the handler was cancelled, this send would hang.
+    block_until_tx.send(()).await.unwrap();
+
+    // Hit the endpoint again without cancelling; it should return a drop count
+    // of 0.
+    // cancelled.
```
davepacheco, line 716: leftover comment?

dropshot/src/api_description.rs
davepacheco: Got it -- in the `spawn` case, we need the handler to be `'static`, so it can't be a reference back to something.

dropshot/src/config.rs
jgallagher: > Thinking out loud: how about: on_client_disconnect with the enum being called OnClientDisconnect with variants CancelHandler and DetachHandler? I like being more concrete than the word "disposition". On the other hand, it goes further down the confusing path of suggested we detach when the client disconnects.

This is almost word-for-word what I had in my first draft, and I scratched it for this exact reason.

> What about: default_handler_task_mode with HandlerTaskMode with variants Detached and CancelOnDisconnect?

That sounds good to me. :+1:

dropshot/src/server.rs
```diff
@@ -850,8 +856,30 @@ async fn http_request_handle<C: ServerContext>(
...
+            // Spawn the handler so if we're cancelled, the handler still runs
+            // to completion.
+            let (tx, rx) = oneshot::channel();
+            tokio::spawn(async move {
+                let result = handler.handle_request(rqctx, request).await;
+
+                // Ignore errors from `send()`: if it fails, it's because the
+                // outer `http_request_handle` future was cancelled while
+                // waiting for this result.
+                _ = tx.send(result);
```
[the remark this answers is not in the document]
jgallagher: > Probably a silly question, but why not use the Future's output as provided by awaiting on the spawn handle?

I... have no idea. I think it's just habit to use oneshot channels like this; normally when spawning I think I'm not in a position where I can await the future directly.

I do think adding a warning here would be handy, and I think easy? I'll give it a shot.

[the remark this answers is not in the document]
jgallagher: Added logging (both here and around the `.unwrap()` below) in efef6f9.

dropshot/tests/test_config.rs
```diff
@@ -345,3 +380,361 @@ async fn test_config_bind_address_https_buffer() {
...
+        // Notify test that we're running.
+        ctx.increment_on_drop_tx.send(()).unwrap();
+
+        // Block until the test tells us to return.
+        () = ctx.block_until_rx.recv().await.unwrap();
+
+        // We weren't cancelled: mem::forget() drop_counter so its drop impl
+        // doesn't run. This leaks a reference to our drop_count, but we're in a
+        // unit test so won't worry about it.
+        mem::forget(drop_counter);
```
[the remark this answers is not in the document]
jgallagher: I took a stab at simplifying them in 8946d44 - I didn't go with the two counters idea, but I think what I did makes what is being tested much more obvious. LMK what you think.

CHANGELOG.adoc
```diff
@@ -19,6 +19,7 @@ https://github.com/oxidecomputer/dropshot/compare/v0.9.0\...HEAD[Full list of co
...
 * https://github.com/oxidecomputer/dropshot/pull/651[#651] The address of the remote peer is now available to request handlers via the `RequestInfo` struct. With this change we've removed the related `From<hyper::Request<B>>` implementation; instead use `RequestInfo::new<B>(&hyper::Request<B>, std::net::SocketAddr)`.
+* https://github.com/oxidecomputer/dropshot/pull/701[#701] adds a new `default_handler_task_mode` field to `ConfigDropshot` and `ServerConfig`. A server configured with `HandlerTaskMode::CancelOnDisconnect` will continue to behave as it has up to this point: a client disconnect will cancel future spawned by its request. `HandlerTaskMode::Detached` will change all handlers to always run to completion, regardless of client disconnects. `ConfigDropshot::default()` will choose `HandlerTaskMode::Detached`, which is a change in behavior if using the default configuration.
```
davepacheco, line 22:
```suggestion
* https://github.com/oxidecomputer/dropshot/pull/701[#701] changes how Dropshot manages the tasks that are used to handle requests.  There are two modes, now configurable server-wide using `HandlerTaskMode`.  Prior to this change, the behavior matched what's now called `HandlerTaskMode::CancelOnDisconnect`: the Future associated with a request handler could be cancelled if, for example, the client disconnected early.  After this change, the default behavior is what's now called `HandlerTaskMode::Detached`, which causes Dropshot to use `tokio::spawn` to run the request handler.  That task will never be cancelled.  This is useful for consumers whose request handlers may not be cancellation-safe.
```

It occurred to me as I was writing this that this might be the first runtime dependency on tokio in Dropshot.  I think when we wrote the bulk of Dropshot, we just assumed consumers were using tokio, but I don't know if we depended in any way on tokio being the executor.  I'm not sure it's possible to use a different executor today.  I don't think it's worth spending much time on this now but at some point we should probably figure out what our dependency really is and document that better.

dropshot/src/server.rs
```diff
@@ -850,8 +856,48 @@ async fn http_request_handle<C: ServerContext>(
...
+            // The only way we can fail to receive on `rx` is if `tx` is
+            // dropped before a result is sent, which can only happen if
+            // `handle_request` panics. We will propogate such a panic here,
+            // just as we would have in `CancelOnDisconnect` mode above (where
+            // we call the handler directly).
+            match rx.await {
+                Ok(result) => result?,
+                Err(_) => {
+                    error!(request_log, "handler panicked");
+                    panic!("handler panicked before sending response");
```
davepacheco, line 896: I'm not sure how to handle this.  It'd be nice to report the error in these cases (or else is it lost completely?).  I'm not clear that we can because we don't have anything usefully printable.

This example is interesting:
https://docs.rs/tokio/latest/tokio/task/struct.JoinError.html#examples-1

Maybe that's the way to go?

davepacheco (commented):
One thing I mentioned above but I think got lost: I think it could be quite valuable if, when using `CancelOnDisconnect` we emitted a warning message when the Future was being dropped while in a request handler.  This can definitely be a follow-on change (and not asking you to do that) but I figured I'd mention it.  The easiest way I can think to do this is to store a bit (AtomicBool?) that starts false, gets set to `true` after we finish the request handler, and then create something in the same scope with a `Drop` handler (maybe `scope_guard` can help here?) that checks if the boolean is false?  Seems kind of cheesy though.

dropshot/src/server.rs
jgallagher: Nice, that looks great. It's a little more work to get the error out; added in 204c111

jgallagher:
> One thing I mentioned above but I think got lost: I think it could be quite valuable if, when using CancelOnDisconnect we emitted a warning message when the Future was being dropped while in a request handler. This can definitely be a follow-on change (and not asking you to do that) but I figured I'd mention it. The easiest way I can think to do this is to store a bit (AtomicBool?) that starts false, gets set to true after we finish the request handler, and then create something in the same scope with a Drop handler (maybe scope_guard can help here?) that checks if the boolean is false? Seems kind of cheesy though.

I added a TODO to this effect, but yeah, I'm not sure how best to do this either at the moment. With the default changing it doesn't seem super urgent at the moment.

> It occurred to me as I was writing this that this might be the first runtime dependency on tokio in Dropshot. I think when we wrote the bulk of Dropshot, we just assumed consumers were using tokio, but I don't know if we depended in any way on tokio being the executor. I'm not sure it's possible to use a different executor today. I don't think it's worth spending much time on this now but at some point we should probably figure out what our dependency really is and document that better.

(Lifting this up since github resolved the conversation when I took your CHANGELOG edit) I don't think that's true: we're using at least `tokio::net::{TcpListener, TcpStream}` and `tokio_rustls`; I assume both of those require a tokio executor?

davepacheco:
> (Lifting this up since github resolved the conversation when I took your CHANGELOG edit) I don't think that's true: we're using at least `tokio::net::{TcpListener, TcpStream}` and `tokio_rustls`; I assume both of those require a tokio executor?

That's possible.  It's not obvious to me that they would.

davepacheco (approved):

PR #702, opened 2023-06-13
Make server shutdown wait on any Detached handler futures

jgallagher:
I ended up using a [`WaitGroup`](https://docs.rs/waitgroup/latest/waitgroup/struct.WaitGroup.html) to track outstanding handler futures instead of trying to accumulate them into something like a `FuturesUnordered` or `JoinSet`, because we don't actually care about their results: we just want to know when they're all done.

Testing this was a little weird: we have to force a situation where we have a detached handler future still running after server shutdown has been requested. It uses a similar pattern to the tests added #701, where the handler will wait for a message on a channel from the test driver before returning, allowing us to construct a client request and then cancel it, leaving the detached handler still running.

### DIFF
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -16,6 +16,7 @@ async-trait = "0.1.68"
 base64 = "0.21.2"
 bytes = "1"
 camino = { version = "1.1.4", features = ["serde1"] }
+debug-ignore = "1.0.5"
 form_urlencoded = "1.1.0"
 futures = "0.3.28"
 hostname = "0.3.0"
@@ -37,6 +38,7 @@ slog-json = "2.6.1"
 slog-term = "2.9.0"
 tokio-rustls = "0.24.0"
 toml = "0.7.4"
+waitgroup = "0.1.2"
 
 [dependencies.chrono]
 version = "0.4.26"
--- a/dropshot/src/server.rs
+++ b/dropshot/src/server.rs
@@ -12,6 +12,7 @@ use super::router::HttpRouter;
 use super::ProbeRegistration;
 
 use async_stream::stream;
+use debug_ignore::DebugIgnore;
 use futures::future::{
     BoxFuture, FusedFuture, FutureExt, Shared, TryFutureExt,
 };
@@ -28,6 +29,7 @@ use hyper::Response;
 use rustls;
 use std::convert::TryFrom;
 use std::future::Future;
+use std::mem;
 use std::net::SocketAddr;
 use std::num::NonZeroU32;
 use std::panic;
@@ -39,6 +41,7 @@ use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::oneshot;
 use tokio_rustls::{server::TlsStream, TlsAcceptor};
 use uuid::Uuid;
+use waitgroup::WaitGroup;
 
 use crate::config::HandlerTaskMode;
 use crate::RequestInfo;
@@ -69,6 +72,9 @@ pub struct DropshotState<C: ServerContext> {
     pub local_addr: SocketAddr,
     /// Identifies how to accept TLS connections
     pub(crate) tls_acceptor: Option<Arc<Mutex<TlsAcceptor>>>,
+    /// Worker for the handler_waitgroup associated with this server, allowing
+    /// graceful shutdown to wait for all handlers to complete.
+    pub(crate) handler_waitgroup_worker: DebugIgnore<waitgroup::Worker>,
 }
 
 impl<C: ServerContext> DropshotState<C> {
@@ -96,6 +102,7 @@ pub struct HttpServerStarter<C: ServerContext> {
     app_state: Arc<DropshotState<C>>,
     local_addr: SocketAddr,
     wrapped: WrappedHttpServerStarter<C>,
+    handler_waitgroup: WaitGroup,
 }
 
 impl<C: ServerContext> HttpServerStarter<C> {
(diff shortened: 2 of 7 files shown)


dropshot/tests/test_detached_shutdown.rs
```diff
@@ -0,0 +1,94 @@
...
+    let teardown_fut = testctx.teardown();
+    tokio::pin!(teardown_fut);
+
+    // Actually tearing down should time out, because it's waiting for the
+    // handler to return (which in turn is waiting on us to signal it!).
+    if tokio::time::timeout(Duration::from_secs(2), &mut teardown_fut)
+        .await
+        .is_ok()
+    {
+        panic!("server shutdown returned while handler running");
```
jgallagher, line 86: Without the addition of 

```rust
handler_waitgroup.wait().await;
```

when constructing `join_handle`, the test fails here.

dropshot/tests/test_detached_shutdown.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    // Signal the handler to complete.
+    release_endpoint_tx.send(()).await.unwrap();
+
+    // Now we can finish waiting for server shutdown.
+    teardown_fut.await;
```
jgallagher, line 93: Without the explict

```rust
mem::drop(self.app_state);
```

in `close()`, this test hangs forever here, because the wait group is still waiting for the last worker (the one held in `DropshotState`) to be dropped.

dropshot/src/server.rs
```diff
@@ -12,6 +12,7 @@ use super::router::HttpRouter;
 use super::ProbeRegistration;
 
 use async_stream::stream;
+use debug_ignore::DebugIgnore;
```
smklein, line 15: TIL, this crate seems useful!

smklein (approved):
Thanks for the solid test - always quirky to test these ordering things, but I appreciate that you put in the effort to make this more robust.

dropshot/Cargo.toml
```diff
@@ -37,6 +38,7 @@ slog-json = "2.6.1"
 slog-term = "2.9.0"
 tokio-rustls = "0.24.0"
 toml = "0.7.4"
+waitgroup = "0.1.2"
```
davepacheco, line 41: This is fine.  I see this pulls in two deps.  It's a shame if there's nothing in tokio or futures that can already do this. :shrug:

dropshot/src/server.rs
```diff
@@ -147,11 +157,13 @@ impl<C: ServerContext> HttpServerStarter<C> {
                         api,
                         private,
                         log,
+                        handler_waitgroup.worker(),
```
davepacheco, line 160: This is probably not something to worry about right now, but: I wonder how debuggable this is, either _in situ_ or _post mortem_.  Like if you walk up to a server that's shutting down and stuck waiting on one of these, would you have any way to know which request it was waiting on?  I imagine eventually we'll want to elevate this to an API but that's probably way down the road.

Just to show what I mean, in the past I built something like this:
https://github.com/TritonDataCenter/node-vasync#barrier-coordinate-multiple-concurrent-operations
In that thing, each outstanding operation has a distinct name.  You can take the whole object and expose that over a debug HTTP API to ask the server "what are the operations you're waiting on".  This proved incredibly useful.  But that was more for stuff like our own RSS, where you've got complicated long-running things that could get stuck somewhere.  This particular case seems less likely to be a problem.  It would just be neat to have a thing like waitgroup but where it was easy to ask it what it was waiting on.

davepacheco (approved):
Ugh, now the websocket case makes me wonder whether it's a problem that we _don't_ cancel these things now?  Do you happen to know what used to happen before #701 if you shut down a server with request handlers running?  What if there was a websocket stream attached?

dropshot/src/server.rs
jgallagher: > This is probably not something to worry about right now, but: I wonder how debuggable this is, either in situ or post mortem. Like if you walk up to a server that's shutting down and stuck waiting on one of these, would you have any way to know which request it was waiting on?

Probably not! Internally the waitgroup is just a wrapper around an [AtomicWaker](https://docs.rs/atomic-waker/1.1.1/atomic_waker/struct.AtomicWaker.html), which itself is just a glorified `AtomicUsize`; I think you could pretty quickly find the _count_, but assuming it's something like `1`, I don't know how you'd find the guilty party.

jgallagher:
> Ugh, now the websocket case makes me wonder whether it's a problem that we don't cancel these things now? Do you happen to know what used to happen before https://github.com/oxidecomputer/dropshot/pull/701 if you shut down a server with request handlers running? What if there was a websocket stream attached?

I _think_ we did the right thing, at least as far as we can - the web socket upgrade handler happens within the normal handler's future, right? And graceful shutdown waits for all handler futures to complete already.

If in practice most websocket handlers do their own `tokio::spawn()` and move the upgraded websocket off to a background task, we wouldn't (and still can't) do anything to wait on that.

davepacheco:
> > Ugh, now the websocket case makes me wonder whether it's a problem that we don't cancel these things now? Do you happen to know what used to happen before #701 if you shut down a server with request handlers running? What if there was a websocket stream attached?
> 
> I _think_ we did the right thing, at least as far as we can - the web socket upgrade handler happens within the normal handler's future, right? And graceful shutdown waits for all handler futures to complete already.
> 
> If in practice most websocket handlers do their own `tokio::spawn()` and move the upgraded websocket off to a background task, we wouldn't (and still can't) do anything to wait on that.

The thing I'm (low-level) worried about is that someone starts up a WebSocket for something like a serial console (just as an example of something that's a best-effort background thing) and now the server cannot be shut down.  It seems like the consumer would probably want to cancel that rather than wait for the client to determine when they can shut down.  The same applies to ordinary HTTP requests but those are usually short-lived and should usually have aggressive timers on them so they seem less likely to cause a problem.

If we previously blocked on WebSockets closing, there's no behavior change here, and this is definitely fine.

PR #703, opened 2023-06-14
Add `#[allow(clippy::unused_async)]` to emitted endpoints

jgallagher:
We require endpoints to be async, but it's not uncommon to have endpoints that themselves do not need to `.await` anything. If a user wants to enable the (off-by-default) `clippy::unused_async` lint, they would need to add this annotation by hand to each such endpoint, or we could just do it for them.

This came from trying clippy in omicron with `-D clippy::unused_async`; the first (several) things it tripped over are fixed by this.


ahl (approved):
looks great!

PR #716, opened 2023-06-30
Add multi part

jessfraz:
This adds support for `multipart/form-data`. Thanks! Let me know if you want anything to be changed but it came out super clean, better than I expected. Added tests as well :)

ping @ahl @davepacheco

### DIFF
--- a/dropshot/tests/test_multipart.rs
+++ b/dropshot/tests/test_multipart.rs
@@ -134,3 +134,35 @@ async fn no_content_type() {
 
     testctx.teardown().await;
 }
+
+#[tokio::test]
+async fn weird_content_type() {
+    let api = api();
+    let testctx = common::test_setup("multipart_client", api);
+
+    let uri = testctx.client_testctx.url("/upload");
+    let request = hyper::Request::builder()
+        .method(Method::POST)
+        .uri(uri)
+        .header("Content-Type", vec![0xf0, 0x28, 0x8c, 0xbc])
+        .body(
+            "--Y-BOUNDARY\r\n\
+                Content-Disposition: form-data; name=\"my_text_field\"\r\n\
+                \r\n\
+                hello\r\n"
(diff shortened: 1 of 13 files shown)


dropshot/tests/test_multipart.rs
```diff
@@ -0,0 +1,72 @@
...
+}]
+async fn api_multipart(
+    _rqctx: RequestContext<usize>,
+    mut body: MultipartBody,
+) -> Result<Response<Body>, HttpError> {
+    // Iterate over the fields, use `next_field()` to get the next field.
+    if let Some(field) = body.content.next_field().await.unwrap() {
+        // Process the field data chunks e.g. store them in a file.
+        if let Ok(bytes) = field.bytes().await {
+            return Ok(Response::builder()
```
adamchalmers, line 34: Now that I think about it, maybe this test should concatenate all the fields together, so that we know it doesn't crash after the first field. I think that would increase test coverage.

ahl:
thanks, @jessfraz--might take us a little bit as we're in crunch time over here; apologies in advance for any delay

jessfraz:
@ahl understandable! If it helps I figured you all would likely need this for a blob storage api or if you allow customers to upload their own isos to boot in vms :)

adamchalmers:
Merge conflicts fixed!

ahl:
@adamchalmers are you folks on the same page with regard to your comments above?

adamchalmers:
Yeah, I updated the tests to increase the coverage. Not really sure if we keep the Copyright Oxide header on test_multipart.rs but I also don't really care! Jess and I spoke last night and we're on the same page.

jessfraz:
LGTM thanks!

ahl (approved):
cool; thanks for this

adamchalmers:
<3 Thanks y'all!

Issue #348, opened 2022-05-17
support for HEAD method


jclulow:
It would be good to allow endpoints to support the HEAD HTTP method in addition to the existing GET, et al.

davepacheco:
We definitely should do this.

Also, it'd be nice if there were an option to implement some kind of support without having to implement it entirely yourself.  I wonder if we could run the `GET` handler and throw away the response (and tweak whatever response headers need to change)?  Maybe give you a flag you can check to see if it's `GET` or `HEAD` so that if you want to avoid some work, you could?

jclulow:
This was fixed in #755.

PR #755, opened 2023-08-25
oxidecomputer/dropshot#348 support for HEAD method

jclulow:


davepacheco:
This seems like a good step.  How about a test?

I still wonder if it'd be useful to be able to opt into Dropshot implementing this for you if you have a `GET` implementation (or some other way to avoid people having to implement these two methods completely separately).  Or maybe the answer to that is "just use functions to commonize the code between your GET and HEAD method".  Anyway, just being _able_ to implement `HEAD` as you've done here seems like a fine step.

jclulow:
> This seems like a good step. How about a test?

Can you point me at an existing test that I should modify to cover this case?

> I still wonder if it'd be useful to be able to opt into Dropshot implementing this for you if you have a `GET` implementation (or some other way to avoid people having to implement these two methods completely separately). Or maybe the answer to that is "just use functions to commonize the code between your GET and HEAD method". Anyway, just being _able_ to implement `HEAD` as you've done here seems like a fine step.

I think all of that is valuable, but obviously more work -- and I think even if we did do that, it would still be valuable to be able to escape and directly handle the request this way as well.

davepacheco:
> > This seems like a good step. How about a test?
> 
> Can you point me at an existing test that I should modify to cover this case?

It's a little cheesy but this file has a bunch of endpoints and some tests that exercise them, which is at least a basic check:
https://github.com/oxidecomputer/dropshot/blob/35d44088fdb263030ef1295f990b87f2c0a4b93c/dropshot/tests/test_demo.rs#L811-L821
I'd add a `HEAD` one analogous to the `DELETE` one.


> > I still wonder if it'd be useful to be able to opt into Dropshot implementing this for you if you have a `GET` implementation (or some other way to avoid people having to implement these two methods completely separately). Or maybe the answer to that is "just use functions to commonize the code between your GET and HEAD method". Anyway, just being _able_ to implement `HEAD` as you've done here seems like a fine step.
> 
> I think all of that is valuable, but obviously more work -- and I think even if we did do that, it would still be valuable to be able to escape and directly handle the request this way as well.

:100:

jclulow:
I have added a test here.  It's a bit more complex than the DELETE one, because I decided to verify that we are actually able to generate a zero-length response with the appropriate non-zero **Content-Length** for a HEAD request, and that it all matches up with a GET request for the same URL in our mock server.

davepacheco:
Offline, we discussed that it's a little easy to misuse this in that one could try to return a `HEAD` response with a body.  We could have the macro check for this.  But urgency is more of a priority right now.  We can still add better checks later.

Issue #56, opened 2020-10-07
Discussion of alternative crates is not publicly available


mikeyhew:
Hey Oxiders!

I heard about this project from your [post on twitter](https://twitter.com/oxidecomputer/status/1308636868859195392). Since we're creating a public API at my work, I was interested to know what you thought of the alternative crates in the ecosystem, and why you decided to roll your own. I was excited to see that [in the documentation for this crate](https://docs.rs/dropshot/0.3.0/dropshot/), you wrote:

>For a discussion of alternative crates considered, see Oxide RFD 10.

Unfortunately it looks like that's an internal discussion that I can't see. I was wondering if you could share any of the specifics about what other crates you looked at and what you thought about them?

davepacheco:
Hi @mikeyhew, sorry for the link to the non-public doc.  (We weren't really sure whether anybody else would be interested in dropshot!)  I'll summarize the key points here.  I really don't want to knock the other crates out there -- many of them look quite good.  Also, keep in mind that we did this research back in January, and some of this is definitely out of date.

Our primary goal was to find or build a foundational component for serving APIs over HTTP in the control plane that we're building.  This is a little unusual for a few reasons.  We have a small team of generalists working on the whole product.  We don't expect to have a dedicated API team.  One person might do the work of implementing SSD firmware upgrade in the OS, in the control plane, _and_ in the public API.  So we wanted the programming interface for this crate to be easy to use and hard to misuse, with as much checked at compile-time as possible.  Also, the components we're building will be running on-prem in somebody else's datacenter.  We won't have an SRE team familiar with the components to debug them.  (We'll have support, of course, but that's a different kind of engagement.)  This is another reason we want stuff to fail at compile-time more than runtime, and it also means we're prioritizing good debuggability and predictable behavior.

As we mentioned in the talk, another important goal for us was to build something with strong OpenAPI support, and particularly where the code could be the source of truth and a spec could be generated from the code that thus could not diverge from the implementation.  We weren't sure that would be possible, but it seemed plausible and worthwhile if we could do it.  None of the crates we found had anything like this.  If I recall correctly, there was once some code (middleware, maybe?) for generating OpenAPI from Rocket, but it wasn't as integrated as we wanted and I believe the code fell out of date.

---

I spent a while developing a prototype atop actix-web, and Dropshot's interface is definitely inspired by it.  But I found several things frustrating about working with actix-web.  These may well have been my own issues.  I was very new to Rust at the time.

[...]

PR #814, opened 2023-10-31
Reference the public summary of RFD 10

Kinrany:
Documentation references Oxide RFD 10 which is not public, as discussed in https://github.com/oxidecomputer/dropshot/issues/56.

This PR adds a link to the summary of RFD 10 from that discussion.


davepacheco (approved):

davepacheco:
Thanks!  We could as well copy the contents of that comment into a section of the README or other docs but for now I'll just add this link.

PR #828, opened 2023-11-22
[schema_util] better error messages for enum variants used as params

sunshowers:
I've hit this twice in the last few months and forgot how to fix it already after the first time scratching my head against it -- enums need to be wrapped in a struct. Provide a more helpful message, and add tests.


dropshot/src/schema_util.rs
```diff
@@ -116,25 +138,110 @@ pub(crate) fn schema2struct(
...
+                    schema
+                )
+            }
+            Schema2StructError::ComplexEnum(schema) => {
+                write!(
+                    f,
+                    "invalid type {:#?}\n\
+                     (hint: this appears to be an enum -- \
+                     if your enum only has unit variants, \
+                     it can be used in this position but needs to be wrapped in a struct)",
```
sunshowers, line 235: I think in some cases complex variants can also be used here, so maybe this needs to be updated to reflect that.

ahl:
Would it be possible to add a test that shows this along the lines of what we have in dropshot/tests/fail/bad_endpoint*?

sunshowers:
Ah so this isn't a compilation failure similar to `bad_endpoint`, this is a failure that occurs when the schema is generated. I can go ahead and add some tests for that though.

dropshot/src/schema_util.rs
```diff
@@ -814,6 +923,60 @@ mod test {
...
+            /// Alice
+            Alice,
+
+            /// Bob
+            Bob,
+
+            // Mallory doesn't get a doc comment but still gets reflected as a
+            // complex variant, because the other two have comments.
+            Mallory,
+        }
```
ahl, line 950: The shape of the derived schemas are going to change depending on serde attributes such as `#[serde(tag = "tag")]` or `#[serde(tag = "tag", content = "content")]`

dropshot/src/schema_util.rs
```diff
@@ -116,25 +138,110 @@ pub(crate) fn schema2struct(
...
+                        ))
+                    }
                 }
             }
 
-            results
+            Ok(results)
         }
 
+        // Unit enums that do not have doc comments are presented as enum_values.
```
ahl, line 163: yes; nit: assumes externally tagged enums

ahl: also not sure but "presented" -> "represented"? or "result in a schema with..."

dropshot/src/schema_util.rs
ahl, line 235: is it important to produce distinct errors here? Would it be sufficient to say "put your enum in a struct, dummy!" for both cases?

ahl (approved):
ship it

dropshot/src/schema_util.rs
sunshowers: I'll add tests to cover the other cases.

dropshot/src/schema_util.rs
ahl: I do like that you're picking them apart... but I think the advice to the user is going to be basically the same.

dropshot/src/schema_util.rs
sunshowers: Done.

dropshot/src/schema_util.rs
sunshowers: It looks like all of the non-simple schemas look exactly the same. I combined them anyway, since as you pointed out the advice is the same. (It also simplifies the test somewhat.)

dropshot/src/schema_util.rs
```diff
@@ -202,38 +203,35 @@ fn schema2struct_impl(
...
-            Schema2StructError::UnitEnum(schema) => {
+            Schema2StructError::InvalidType(schema) => {
                 write!(
                     f,
-                    "invalid type {:#?}\n\
-                     (hint: this appears to be an enum with unit variants, \
-                     and needs to be wrapped in a struct)",
-                    schema
+                    "invalid type: {}",
+                    serde_json::to_string_pretty(schema).unwrap(),
```
sunshowers, line 227: A change I snuck in was to use `serde_json::to_string_pretty`, as Adam recommended in chat.

sunshowers (commented):
Per discussion, I didn't end up implementing an integration test because it would be difficult to test for panics and the unit test is good enough.

PR #862, opened 2023-12-29
WIP: dropshot support for multiple API versions

ahl:
I think these are most of the steps:

- [x] `#[endpoint]` macro parsing
- [ ] macro creation of versioned endpoints
- [ ] route tree creation of multiple versions including detection of overlapping version ranges
- [ ] enumeration of versions
- [ ] proper routing to versioned handlers
- [ ] extraction of OpenAPI document for a particular version

---
Here's my proposal for how we parse versions with the proc macro:

```rust
#[endpoint {
    method = GET,
    path = "/testing/demo1",
    // versions = "1.2.3"..,
    // versions = .."1.2.5",
    versions = "1.2.3".."1.2.5",
}]
async fn demo_handler_args_1(..) {}
```

### DIFF
--- a/dropshot/src/api_description.rs
+++ b/dropshot/src/api_description.rs
@@ -47,6 +47,7 @@ pub struct ApiEndpoint<Context: ServerContext> {
     pub extension_mode: ExtensionMode,
     pub visible: bool,
     pub deprecated: bool,
+    pub versions: ApiVersions,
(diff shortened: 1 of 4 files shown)


davepacheco:
Elaborated on the plan here that we discussed a few weeks ago in #869.

PR #878, opened 2024-01-10
[test-utils] add log_prefix_for_test + store name of test in LogContext

sunshowers:
While trying to diagnose https://github.com/oxidecomputer/omicron/issues/4779,
which only occurs in CI, I'm trying to collect ClickHouse logs. To do that we
need a way to uniquely identify the ClickHouse logs per test.

It is possible to derive the log prefix from `log_file_for_test`, but it seems
cleaner to do it this way.

Not entirely sure about this design though -- a `LogContext` represents a single log file, but in this case we're also using it to generate a second path, kinda.


ahl (approved):
this looks good to me, but @davepacheco is more likely to have a useful opinion here than am I

davepacheco (approved):

davepacheco:
Feel free to land when ready!

Issue #901, opened 2024-02-08
Support Multiple Key Types


kevinmeziere:
Currently Dropshot ConfigTls expects a key that matches Item::Pkcs8Key.  Some certificate authorities do not provide a key pair that matches this.

A concrete example is using a key pair obtained from tailscale cert. Today the only key proved by tailscale is an EC key.

davepacheco:
Thanks.  In the latest release, I think you could use `ConfigTls::Dynamic` with your own `rustls::ServerConfig` to do this, right?  (The fix in #902 looks simple enough, too.)

kevinmeziere:
Yea thats the idea, get the key/cert and create a Some ConfigTls::AsBytes. I didn't go the ConfigTls::Dynamic route as it looked like that was maybe looking for its own TLS handler... but I could misunderstanding the usage.

davepacheco:
Dropshot always constructs a `rustls::ServerConfig`.  When you use `ConfigTls::Dynamic`, we use the one the caller provides directly:
https://github.com/oxidecomputer/dropshot/blob/59f102ee0d44ee324da1eb7a6804c2af55043e9e/dropshot/src/server.rs#L481

`ConfigTls::AsBytes` or `ConfigTls::AsFile` are just shortcuts for a common case.  They boil down to:
https://github.com/oxidecomputer/dropshot/blob/59f102ee0d44ee324da1eb7a6804c2af55043e9e/dropshot/src/server.rs#L483-L537

So if you need a workaround, you can make your own `rustls::ServerConfig` using similar code, modified in the way you did in #902 to accept the kind of key you need.

#902 still seems worth doing.  I just wanted to let you know in case this unblocks you.

davepacheco:
Closing this because I think you can do this today with `ConfigTls::Dynamic`.  We can still do #902 but we don't need this open to track it.

PR #902, opened 2024-02-08
allow multiple key types

kevinmeziere:
Addresses https://github.com/oxidecomputer/dropshot/issues/901

### DIFF
--- a/dropshot/src/server.rs
+++ b/dropshot/src/server.rs
@@ -518,8 +518,7 @@ impl TryFrom<&ConfigTls> for rustls::ServerConfig {
             .map_err(|err| {
                 io_error(format!("failed to load certificate: {err}"))
             })?;
-        let keys = rustls_pemfile::pkcs8_private_keys(&mut key_reader)
-            .collect::<Result<Vec<_>, _>>()
+        let keys = rustls_pemfile::private_key(&mut key_reader)
             .map_err(|err| {
                 io_error(format!("failed to load private key: {err}"))
             })?;


davepacheco:
@kevinmeziere it looks like this is failing a few CI checks.

@ahl can you take a look at this one?  It seems okay but I think you've been in this area more recently.

kevinmeziere:
Should be cleaned up now. not really sure what to make of the the failing 'cargo fmt -- --check' as I cannot reproduce locally.

dropshot/src/server.rs
```diff
@@ -518,20 +518,20 @@ impl TryFrom<&ConfigTls> for rustls::ServerConfig {
             .map_err(|err| {
                 io_error(format!("failed to load certificate: {err}"))
             })?;
-        let keys = rustls_pemfile::pkcs8_private_keys(&mut key_reader)
-            .collect::<Result<Vec<_>, _>>()
+        let keys = rustls_pemfile::private_key(&mut key_reader)
+            //.collect::<Result<Vec<_>, _>>()
```
ahl, line 522: `rustls_pemfile::private_key` docs say "Yields the first PEM section describing a private key (of any type), or an error if a problem occurs while trying to read PEM sections."

Github isn't letting me make this suggestion for the full span, but from 521-529 I'd suggest doing:

```rust
        let private_key = rustls_pemfile::private_key(&mut key_reader)
            .map_err(|err| {
                io_error(format!("failed to load private key: {err}"))
            })?
            .ok_or_else(|| io_error("no private key found".into()))?;
```

ahl (changes requested):
please see my suggestion; would it be easy enough to add a test to validate this new behavior?

ahl:
@kevinmeziere checking in on this; how would you like to proceed?

ahl:
> Sorry for the slowness... I have taken your changes (which work nicely) and have yet to make tests. I hope to get to that this week.

No apology needed! Just checking in. Thank you for the submission. With regard to tests, I don't want to create an unnecessary hurdle... and I don't want functionality you're depending on to regress! If testing is too onerous, let us know!

kevinmeziere:
After spending sometime poking around at this it is not clear to me that it is possible to have rcgen create an ec key pair. It would seem this is necessary to do in (or a duplicate of) tests/common/mod.rs:62. I do not have a good enough grasp of cryptography to be comfortable to try and implement a new test with a different certificate generating library. I see a couple ways forward:
1 - Leave untested. The existing test do check for a pkcs8 pair, so it would just be additional key types that are not tested.
2 - Withdraw. For my own purposes I would be ok managing my own local patch.
3 - Create some terrible test that either has a cert file that it loads from a file or shells out to create a pair and loads. Seems terrible.

Please let me know if you have some input on this.

ahl:
I think leaving it untested would be fine. No need to withdraw it -- this is a good fix. I think some static data in a file or thunked into a string might not be so horrible? Let me know what you think. Thanks.

kevinmeziere:
I was mainly concerned that the certificate had to be valid for rustls to load it. In the other test dropshot uses "localhost" so that seems like a fine assumption, as we could just set the expiration to 20 years from now. If thats ok with you I'll write it up.

ahl:
> I was mainly concerned that the certificate had to be valid for rustls to load it. Also because we would be loading the certificate it would be testing more than just "can dropshot load cert type x", so potentially changes in rustls in the future would case fails that have nothing to do with what we are "testing". In the other test dropshot uses "localhost" so that seems like a fine assumption, also we could just set the expiration to 20 years from now. If thats ok with you I'll write it up.

I think that's fine. A comment on the test that's like: "welcome to the year 2044. You're still using dropshot? Presumably from your Martian colony? Was 2038 a big deal... or meh? Anyhow, here's how you renew what we ancients called 'a cert'"

ahl:
Can you update the PR with the code suggestions above if those indeed seem like the right approach? We can figure out testing later.

PR #919, opened 2024-02-26
Use `writeln!` over `eprintln!` to avoid a potential panic

### LINKED ISSUE #811: mgs panics at startup when out of disk space for stderr
A core was retrieved from the field with this (demangled) stack:
```
> $C ! demangle
fffff5ffffdf99e0 libc.so.1`_lwp_kill+0xa()
fffff5ffffdf9a10 libc.so.1`raise+0x22(6)
fffff5ffffdf9a60 libc.so.1`abort+0x58()
fffff5ffffdf9a70 panic_abort::__rust_start_panic::abort::h0cfb9e9813446f8b+9()
fffff5ffffdf9a80 __rust_start_panic+9()
fffff5ffffdf9b20 rust_panic+0x10()
fffff5ffffdf9bd0 std::panicking::rust_panic_with_hook::ha18432c291108e47+0x2a2()
fffff5ffffdf9c20 std::panicking::begin_panic_handler::{{closure}}::h4645986a55c3ef2b+0xc6()
fffff5ffffdf9c30 std::sys_common::backtrace::__rust_end_short_backtrace::h6107fb71e173f9a5+9()
fffff5ffffdf9c70 rust_begin_unwind+0x71()
fffff5ffffdf9cb0 core::panicking::panic_fmt::h863016252fdb1147+0x33()
fffff5ffffdf9d40 std::io::stdio::_eprint::h4074584434e31527+0xea()
fffff5ffffdf9e00 dropshot::logging::log_drain_for_file::h583cfa5d5cfb4a33+0x134()
fffff5ffffdfa150 dropshot::logging::ConfigLogging::to_logger::h4dc5d413269a715f+0x1a6()
fffff5ffffdfbb90 omicron_gateway::start_server::{{closure}}::h111cb597d07e89a5+0xa8()
fffff5ffffdfd760 tokio::runtime::park::CachedParkThread::block_on::h2eef03e43bc93363+0x7bf()
fffff5ffffdfe310 tokio::runtime::context::runtime::enter_runtime::hc471aafe41f806ae+0x71()
fffff5ffffdfeeb0 tokio::runtime::runtime::Runtime::block_on::h05c4ceb4a7406867+0x64()
fffff5ffffdffb70 mgs::main::h77a49bf9a1bccd53+0x93()
fffff5ffffdffb80 std::sys_common::backtrace::__rust_begin_short_backtrace::h557fdc0f2339b596+6()
fffff5ffffdffb90 std::rt::lang_start::_$u7b$$u7b$closure$u7d$$u7d$::h22d2e8743e6db155 (.llvm.5752716022703285789)+0xc()
fffff5ffffdffc90 std::rt::lang_start_internal::h3a9b87d1cf4661d4+0x2c4()
fffff5ffffdffcb0 main+0x2c()
fffff5ffffdffce0 _start_crt+0x87()
fffff5ffffdffcf0 _start+0x18()
```
This looks at first similar to oxidecomputer/omicron#4352 but here we're still just trying to get started and `slog` is not directly on the scene.  Helpfully, `std::io::stdio::_eprint` is undocumented, but is known to be an implementation detail of `eprint!`, which is documented to panic if we can't write to stderr (https://doc.rust-lang.org/std/macro.eprint.html).  For that reason, using `eprint!` seems unsuitable for a daemon; panicking is not an appropriate response to running out of disk space as it can result in the service ending up in the maintenance state, from which automated recovery is not possible.  Either the record should be dropped or, more dangerously, the daemon could block until disk space becomes available (see oxidecomputer/omicron#1014 for some additional background discussion on logging configuration tradeoffs).  Worse, though, this seems to be happening from a library (dropshot); I can't tell whether this is a bug in the way dropshot's logging facility has been configured by mgs or a bug in dropshot itself (in which case this bug should be moved).

jgallagher:
On startup when logging to a file, we use(d) `eprintln!` to leave a note to the user with the path where logs are emitted. This can cause processes using dropshot to panic, so this PR replaces `eprintln!` with `writeln!`.

Fixes #811.

---

I would have liked to add an automated test for this, but explicitly testing `eprintln!` seems hard; there's no standard library way to reopen stderr, for example. I tested this manually by patching the `basic` example like so:

```diff
--- a/dropshot/examples/basic.rs
+++ b/dropshot/examples/basic.rs
@@ -29,7 +29,11 @@ async fn main() -> Result<(), String> {
     // For simplicity, we'll configure an "info"-level logger that writes to
     // stderr assuming that it's a terminal.
     let config_logging =
-        ConfigLogging::StderrTerminal { level: ConfigLoggingLevel::Info };
+        ConfigLogging::File {
+            level: ConfigLoggingLevel::Info,
+            path: "basic.log".into(),
+            if_exists: dropshot::ConfigLoggingIfExists::Append,
+        };
     let log = config_logging
         .to_logger("example-basic")
         .map_err(|error| format!("failed to create logger: {}", error))?;
```

and then running it like this:

```console
./target/debug/examples/basic 2> >(true)
```

On this branch, the example continues to run and contains this in `basic.log`:

```
21:19:54.931Z WARN example-basic: failed to report log path on stderr
    err = Broken pipe (os error 32)
```

On `main`, running the same command immediately exits with code 101 (i.e., a Rust panic).

### DIFF
--- a/dropshot/src/logging.rs
+++ b/dropshot/src/logging.rs
@@ -112,15 +112,15 @@ impl ConfigLogging {
                 // isn't writeable (e.g., if stderr has been redirected to a
                 // file on a full disk). Our options seem to be:
                 //
-                // a) Return an error if we fail to emit this message b)
-                // Silently swallow errors c) If an error occurs, try to relay
-                // that fact
+                // a) Return an error if we fail to emit this message
+                // b) Silently swallow errors
+                // c) If an error occurs, try to relay that fact
                 //
                 // (a) doesn't seem great, because it's possible we're able to
                 // run just fine (and possibly even use the logger we're about
                 // to create, as we don't know that it will suffer the same fate
                 // that writing to stderr does). (b) is defensible since this is
-                // just an informational message. We go with (c) to attempt to
+                // just an informational message. We go with (c) and attempt to
                 // leave a breadcrumb in the log itself.
                 if let Err(err) = writeln!(
                     io::stderr(),
--- a/dropshot/src/logging.rs
+++ b/dropshot/src/logging.rs
@@ -11,6 +11,7 @@ use slog::Level;
 use slog::Logger;
 use std::fs::OpenOptions;
 use std::io::LineWriter;
+use std::io::Write;
 use std::{io, path::Path};
 
 /// Represents the logging configuration for a server.  This is expected to be a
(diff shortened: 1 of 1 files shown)


davepacheco (approved):
`eprintln!()` panics?!  Oops.

PR #924, opened 2024-02-29
Serialize page token with MessagePack instead of as JSON string

### LINKED ISSUE #923: It would be nice if page tokens were shorter
This is helpful but not essential for https://github.com/oxidecomputer/console/issues/1102, putting page tokens in console URLs. The real blocker is #436.

---

Page tokens are base64ed JSON and they come out pretty long. Here's an example:

```
> atob('eyJ2IjoidjEiLCJwYWdlX3N0YXJ0Ijp7InNvcnRfYnkiOiJuYW1lX2FzY2VuZGluZyIsInByb2plY3QiOiJhbGFuIiwibGFzdF9zZWVuIjoienp6LWluc3QtMTE0In19')
'{"v":"v1","page_start":{"sort_by":"name_ascending","project":"alan","last_seen":"zzz-inst-114"}}' 
```

`oxiderack.com/projects/mock-project/instances?page=eyJ2IjoidjEiLCJwYWdlX3N0YXJ0Ijp7InNvcnRfYnkiOiJuYW1lX2FzY2VuZGluZyIsInByb2plY3QiOiJhbGFuIiwibGFzdF9zZWVuIjoienp6LWluc3QtMTE0In19` would work fine _I guess_, but to me it looks a bit silly. It should be pretty easy to encode the token in a more efficient format like [MessagePack](https://docs.rs/rmp-serde/latest/rmp_serde/#).

https://github.com/oxidecomputer/dropshot/blob/6b410861c105052faa30536edc9d271b2abf1be7/dropshot/src/pagination.rs#L417-L430

david-crespo:
Proof of concept showing how easy this is to change. Would close #923 if we like it. Considerations:

* We should get a better sense than I did in #923 of how much of an improvement this is in the real world
* Should we try compression also/instead?
* Do we need to rev the `PaginationVersion`? Currently, Nexus deploys involve shutting down all nodes, so the problem of old and new formats being used at the same time isn't likely to be a big problem, plus we'd have to add logic for handling both formats
  https://github.com/oxidecomputer/dropshot/blob/6b410861c105052faa30536edc9d271b2abf1be7/dropshot/src/pagination.rs#L387-L404


david-crespo:
This seemed fine, but @ahl talked me out of putting these cursors in the query string in console, so will close for now.

PR #942, opened 2024-03-22
Remove unused dependency: proc-macro2

jmqd:
I discovered this via `cargo-udeps`. It seems to be completely unused, and the project builds and tests green with it removed.

It seems like the dependency was introduced here, but is no longer used: https://github.com/oxidecomputer/dropshot/commit/15acd1a0f8b78f3af9098e4839d570589b483fda.


ahl:
Thanks for this!

Issue #914, opened 2024-02-23
end-of-request USDT probes, log entries don't always happen


davepacheco:
After #701, there are potentially two async tasks involved in request handlers:

- the one in which hyper invokes us when a new request comes in.  This one also runs the request handler if using `HandlerTaskMode::CancelOnDisconnect`.
- if using `HandlerTaskMode::Detached`, the one in which the request handler runs

If the request is cancelled (e.g., if the client disconnects), the first of these tasks gets cancelled.  With `HandlerTaskMode::Detached`, the request handler still runs to completion.  Unfortunately, the "request completed" log entry USDT probes happen in the first of these contexts.  So you don't get these events when the client request was cancelled.  This makes it much harder to write correct DTrace scripts and also makes it much harder to debug from logs since you can't assume there's a "request completed" log entry for all requests that did essentially complete.

(This was noticed while debugging oxidecomputer/omicron#5111.)

PR #958, opened 2024-04-02
Log warning when request dropped with `CancelOnDisconnect`

deathbyknowledge:
Greetings!

I wanted to start contributing to the repo and found #914 as a good first issue so I read the discussion and tried to implement it.

I did add the scopeguard dependency just for this one scenario... not sure if that's OK. I'm happy to make any changes necessary, this is just the first try.
I've updated the test case to make sure the log gets written as well.

Let me know if I should update anything!


dropshot/src/test_util.rs
```diff
@@ -451,6 +451,12 @@ impl LogContext {
         &self.test_name
     }
 
+    /// Returns the log path.
+    #[inline]
```
davepacheco, line 455:
```suggestion
    #[inline]
```

Can we make this non-`pub`?

dropshot/tests/test_config.rs
```diff
@@ -542,6 +544,14 @@ async fn test_config_handler_task_mode_cancel() {
 
     server.close().await.unwrap();
 
+    // Wait for the log to finish.
+    sleep(Duration::from_millis(100)).await;
```
davepacheco: This is not great -- it's going to lead to flaky tests.  Elsewhere we use [this function](https://github.com/oxidecomputer/omicron/blob/2efb6148774bd56668e9322e180d9cd793c8d620/test-utils/src/dev/poll.rs#L32-L91) to keep checking a condition until some maximum time has elapsed.  That's a lot more robust.  But it's not easy to use here.  You could duplicate the pattern here (it's basically just a loop until some maximum timeout has elapsed).  I'm not sure it's worth testing this though -- I don't think we have other tests that verify log entries.  (Not that it would be a bad idea!)

dropshot/src/server.rs
```diff
@@ -901,11 +902,24 @@ async fn http_request_handle<C: ServerContext>(
...
-            //
-            // TODO-robustness: We should log a warning if we are dropped before
-            // this handler completes; see
-            // https://github.com/oxidecomputer/dropshot/pull/701#pullrequestreview-1480426914.
-            handler.handle_request(rqctx, request).await?
+            // In case it is indeed cancelled, we keep track of the handler
+            // progress with `done`, so the scopeguard below can log a warning
+            // if it's dropped before completion.
+            let done = AtomicBool::new(false);
+            let request_log = rqctx.log.clone();
```
davepacheco, line 909: Since this is common with L928, you could move it up to L899 and delete this one and the one on L928.

davepacheco (changes requested):
Thanks for taking a look at this.  This seems like an improvement in that we'll get the log message in the cancellation case.  But this is not quite what I had in mind with #914.  It doesn't fix the USDT probes, for example.

This code was originally structured with `http_request_handle_wrap()` taking care of all of the before/after request stuff.  That's currently the USDT probes and log message.  (In the future, we might put something like counters or tracing events there, too.)  I like putting this all in one place because it's clear that these things will all happen together (or all not happen) and it's also obvious where to put new things that need to happen before or after request handling.

Of course, the problem is that function isn't async cancel-safe -- we didn't realize when we wrote it that async cancellation means the "after" stuff won't necessarily run.  But I think maybe the more general solution here is to use scopeguard in that function to ensure that we run the "after" stuff.  That probably means the big match at the end will be over three variants: handler returned success, handler returned an error, or the handler was cancelled.

Does that make sense?  You don't have to do that here -- as I said, this change is an improvement as-is (modulo the other feedback I've got here) -- but I wouldn't consider #914 fixed until we do the other stuff.

deathbyknowledge:
> Does that make sense? You don't have to do that here -- as I said, this change is an improvement as-is (modulo the other feedback I've got here) -- but I wouldn't consider #914 fixed until we do the other stuff.

Understood! I definitely got the scope of #914 wrong, I see it now. 
I'll leave this to be the tiny log only then. I'll try tackling the bigger issue with more time in a different PR.

Thanks for the crystal clear answer btw!

dropshot/tests/test_config.rs
deathbyknowledge: I agree, this is pretty janky. I hadn't seen the `wait_for_condition` function although it's probably not worth starting to verify log entries in this PR so I've removed that whole bit.

dropshot/src/test_util.rs
deathbyknowledge: I added the method so it could be used from within the test module to read the log file and check if the entry was there, so it had to be `pub` in order for it to be accessible (I think?). That being said, since the check is no longer there, I've removed this bit too.

davepacheco (approved):

davepacheco:
Sorry for the delay on this.  This looks good now.  But I wasn't sure if this is obviated by #965 -- will take a look at that one.

davepacheco:
Closing to focus on #965.

PR #988, opened 2024-05-03
Add a trailing newline to OpenAPI spec output

jgallagher:
Fixes a papercut - if one opens an OpenAPI spec JSON file in an editor that wants to save files with trailing newlines, that breaks later expectorate tests. This changes the output to include a trailing newline instead.


dropshot/src/api_description.rs
```diff
@@ -1058,9 +1059,11 @@ impl<'a, Context: ServerContext> OpenApiDefinition<'a, Context> {
         out: &mut dyn std::io::Write,
     ) -> serde_json::Result<()> {
         serde_json::to_writer_pretty(
-            out,
+            &mut *out,
             &self.api.gen_openapi(self.info.clone()),
-        )
+        )?;
+        writeln!(out, "").map_err(serde_json::Error::custom)?;
```
sunshowers, line 1065:
```suggestion
        writeln!(out).map_err(serde_json::Error::custom)?;
```

ahl:
I don't know if this makes me dumb or a genius, but my editor strips out the trailing newline in json documents. I'm sure I could configure it to preserve them... but it does seem that this may be an annoyance in either direction.

jclulow:
In general I believe most tools expect a UNIX text file to be either totally empty (0 bytes) or a whole number of lines, each terminated by a linefeed.  What text editor is _removing_ them?!

ahl:
Vim

sunshowers:
I think the general expectation is as @jclulow said, to always have a trialing newline yeah. Hence e.g. git shows the "No trailing newline at end of file" message.

ahl:
I see: "an industry standard best practice"!

ahl:
> I got something I'm interpreting as permission-to-merge from Adam offline.

I think my exact words were "over my dead body"

PR #994, opened 2024-05-07
Preserve extensions in the translation from JSON schema to OpenAPI

ahl:


dropshot/src/schema_util.rs
```diff
@@ -488,6 +488,14 @@ fn j2oas_schema_object(
         data.write_only = metadata.write_only;
     }
 
+    // Preserve extensions
+    data.extensions = obj
+        .extensions
+        .iter()
+        .filter(|(key, _)| key.starts_with("x-"))
```
bnaecker, line 495: Does this need to catch upper case `X-`?

ahl: It appears to require just the lowercase prefix:
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#specification-extensions

> The extensions properties are implemented as patterned fields that are always prefixed by "x-".

PR #1006, opened 2024-05-14
[dropshot-endpoint] move channel and endpoint code into their own files

sunshowers:
I'm going to add a `dropshot::server` macro next to these, and it's big enough
that it warrants its own file. The others should similarly live in their own
files as well.

I also moved some utility code shared between the endpoint and the server into
another file.

This is mostly big-hammer refactoring for now with no functional changes. The
server macro PR will do some more fine-grained refactoring on top of this.


ahl:
Do we want to consider renaming this crate to `dropshot-macro` now? I regret naming it `dropshot_endpoint`. No is a fine answer... it probably doesn't matter that much.

sunshowers:
Yeah, doesn't matter that much, so thinking "no" for now. It still is endpoints of various kinds ultimately.

ahl (approved):

PR #1007, opened 2024-05-15
[dropshot] add bad_channel fixtures

sunshowers:
I'm going to be touching a bunch of this code soon, and I realized we didn't
really have any fixtures to check that we were producing good error messages
for channels. So add some.

The current bad_channel tests provide a baseline -- we can definitely do better
for some of these tests. For now, we can ensure that we aren't regressing them.

PR #1008, opened 2024-05-15
[3/n] [dropshot-endpoint] factor out endpoint and parameter validation

sunshowers:
We're going to reuse a lot of this code for server endpoints. Refactor code to
prepare for that. I think it's overall a nice improvement :)

For successful generation, there are absolutely no changes in output. (This is
a standalone PR, so it's easy to see this.)

In error cases:

1. Since we encapsulate the different parts of parsing into their own types, there are some minor changes to the situations where we produce errors. Among the fixtures we have, there are a few outputs that change -- I think all for the better.
2. If the function is valid, but the metadata is not, we pass through the original function without doing any modifications to it. This leads to a much better experience for traits, and a slightly better one for functions.

Also add a little bit of error-handling infrastructure in the form of an
`ErrorStore`. This makes it reasonable to have this kind of nested
structure, where each component can determine whether it produced errors. We're
going to use this more extensively within the server macro, since that has
to maintain separate error contexts per endpoint.

As part of this, `do_endpoint` and `do_channel` no longer return `Result` types -- all errors are handled internally, and instead they return a pair of output and a list of errors. This resolves an inconsistency that I've been bothered by for a bit.

Depends on #1007.


dropshot/tests/fail/bad_endpoint2.stderr
```diff
@@ -18,23 +18,3 @@ error: Expected a non-receiver argument
...
-  --> tests/fail/bad_endpoint2.rs:13:23
-   |
-9  | #[endpoint {
-   | ---------- `Self` type implicitly declared here, by this `impl`
-...
-13 | async fn bad_endpoint(self) -> Result<HttpResponseOk<()>, HttpError> {
-   |                       ^^^^
-   |                       |
-   |                       use of generic parameter from outer item
-   |                       refer to the type directly here instead
```
sunshowers, line 40: I think these errors in particular are unhelpful, and detecting them syntactically is much cleaner.

sunshowers:
(Looks like the tests aren't running because this is a dependent PR not targeting main. Yay for github's terrible support for stacks.)

dropshot/tests/fail/bad_endpoint2.stderr
ahl: What  does "detecting them syntactically" mean and how does that relate to this change? Agreed that these errors are not helpful.

dropshot_endpoint/src/error_store.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn new(parent: Option<usize>) -> Self {
+        Self { parent, has_errors: false }
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct ErrorSink<'a, T> {
+    // It's a bit weird to use both a lifetime parameter and a RefCell, but it
+    // makes sense here. The lifetime parameter ensures that the error context
+    // doesn't outlive the collector, and the RefCell exists so we can
```
ahl, line 68: what's a collector?

dropshot_endpoint/src/error_store.rs
```diff
@@ -0,0 +1,105 @@
+// Copyright 2024 Oxide Computer Company
+
+//! Handle lists of errors in Dropshot.
```
ahl, line 3: Can you explain more of what you're doing and why. Obviously you have my attention with the use of unsafe and RefCell--I'd appreciate some context in order to give it a reasonable review.

dropshot_endpoint/src/error_store.rs
sunshowers: Ah, the collector is just the thing that collects and stores errors.

dropshot_endpoint/src/error_store.rs
sunshowers: Give me a bit of time to expand on this, will add a comment and also put in an explanation here. (But at a high level, this is basically an `Rc<RefCell<Vec<Error>>>` without the `Rc`. Doing this requires unsafe to perform lifetime extension, which is sound overall -- and actually a common pattern.)

dropshot/tests/fail/bad_endpoint2.stderr
sunshowers: By "syntactically" I just mean looking at the token stream and rejecting code within the proc macro, rather than "semantically", generating code that will fail to compile and relying on rustc's errors. (Is there a better term for this? I've always used syntactically/semantically for this in the context of proc macros)

dropshot_endpoint/src/error_store.rs
sunshowers: Managed to remove the unsafe (also added `#[forbid(unsafe_code)]`), and added a couple of comments to explain why the `RefCell` is necessary.

dropshot_endpoint/src/error_store.rs
sunshowers: Fixed this comment up.

dropshot_endpoint/src/error_store.rs
ahl: I'm probably too scared of `unsafe` but thank you for figuring out a way around it. I hope it's not much worse!

dropshot_endpoint/src/error_store.rs
```diff
@@ -1,105 +1,164 @@
 // Copyright 2024 Oxide Computer Company
 
-//! Handle lists of errors in Dropshot.
+//! Handle lists of errors that occur while generating the proc macro.
+//!
+//! See the documentation of [`ErrorStore`] for more information.
 
 use std::cell::RefCell;
 
+/// Top-level struct that holds all errors encountered during the invocation of
```
ahl, line 9: this is excellent

dropshot_endpoint/src/error_store.rs
```diff
@@ -1,105 +1,164 @@
...
+///   function signature, and vice versa.
+/// * Within the signature, each type should be considered separately -- so an
+///   issue in an extractor type shouldn't stop us from parsing the return type.
+/// * But within a section, if there are errors, we should avoid producing the
+///   corresponding output -- because doing so can lead to confusing errors.
+///
+/// So the typical Rust pattern of returning on the first error isn't quite good
+/// enough for us, and what we need is:
+///
+/// * a hierarchical way to collect errors
```
ahl, line 31: it's not clear to my why hierarchy is important or how it manifests to users... but this is great, and no need to block on my confusion!

ahl (approved):
ship it

dropshot_endpoint/src/error_store.rs
sunshowers: It's actually better!

dropshot_endpoint/src/error_store.rs
sunshowers: Basically the hierarchy is:

```
top level
|-> metadata
|-> params
      |-> request context
      |-> each extractor
      |-> return type
```

and errors in some node in the tree should not stop sibling nodes from being validated.

PR #1010, opened 2024-05-16
[2/n] [dropshot-endpoint] turn valid endpoint tests into snapshot tests

sunshowers:
Use expectorate to manage these tests. It's much easier to see the outputs, and
to make changes to them, this way.


dropshot_endpoint/src/endpoint.rs
```diff
@@ -465,82 +468,12 @@ mod tests {
             },
         )
         .unwrap();
-        let expected = quote! {
```
ahl, line 468: code that predates expectorate! how quaint!

ahl (approved):

PR #1011, opened 2024-05-16
[6/n] [dropshot-endpoint] add extra validation for parameters

sunshowers:
We are forced to do some of this validation for trait-based servers, and
catching things like non-static lifetime parameters early can make this a nicer
experience. We don't try and capture everything, just some of the things that
are either required for, or closely related to, trait-based servers.

Depends on #1009.


dropshot_endpoint/src/endpoint.rs
```diff
@@ -479,6 +491,204 @@ impl<'ast> EndpointParams<'ast> {
...
+    let mut visitor = Visitor { kind, name_str, errors: &errors };
+    visitor.visit_type(ty);
+
+    // Don't return the type if there were errors.
+    (!errors.has_errors()).then(|| ty)
+}
+
+/// A representation of the RequestContext type.
+#[derive(Clone, Eq, PartialEq)]
+enum RqctxTy<'ast> {
```
ahl, line 577: this is what we're going to add to for the trait-based approach?

dropshot_endpoint/src/endpoint.rs
```diff
@@ -834,6 +1044,57 @@ mod tests {
         );
     }
 
+    // These argument types are close to being invalid, but are actually valid.
```
ahl, line 1047: Is it that these are odd but valid types? or could you elaborate on this proximity to invalidity?

ahl (approved):
Is it the case today that we're not doing the appropriate checks for lifetimes?

dropshot_endpoint/src/endpoint.rs
sunshowers: > this is what we're going to add to for the trait-based approach?

Yeah.

sunshowers:
> Is it the case today that we're not doing the appropriate checks for lifetimes?

It won't succeed at compiling, but it'll produce some pretty rough error messages: https://gist.github.com/sunshowers/c83254c5f1002b410ae05f9426c78b2d

dropshot_endpoint/src/endpoint.rs
sunshowers: Expanded the comment here to describe the changes which would make these types invalid.

dropshot_endpoint/src/endpoint.rs
```diff
@@ -1082,7 +1082,14 @@ mod tests {
         );
     }
 
-    // These argument types are close to being invalid, but are actually valid.
+    // These argument types are close to being invalid, but are either okay or
+    // we're not sure.
+    //
+    // * `MyRequestContext` might be a type alias, which is legal for
+    //   function-based servers.
```
ahl, line 1089: will type aliases not valid for trait-based servers?

dropshot_endpoint/src/endpoint.rs
```diff
@@ -1082,7 +1082,14 @@ mod tests {
...
+    // * We don't support non-'static lifetimes, but 'static is fine.
+    // * We don't support generic types within extractors, but `<X as Y>::Z` is
+    //   actually a concrete, non-generic type.
```
ahl, line 1092: tricky!

dropshot_endpoint/src/endpoint.rs
```diff
@@ -1094,7 +1101,7 @@ mod tests {
                 /** handle "xyz" requests */
                 async fn handler_xyz(
                     _rqctx: MyRequestContext,
-                    query: Query<&'static str>,
+                    query: Query<QueryParams<'static>>,
                     path: Path<<X as Y>::Z>,
                 ) -> Result<HttpResponseOk<()>, HttpError> {
```
ahl, line 1106: this is a nit, but I'd love to get rid of all references to `HttpResponseOk<()>` as I **never** want people to do that. Find to defer that, but just saying it so one of us remembers.

dropshot_endpoint/src/endpoint.rs
```diff
@@ -1082,7 +1082,14 @@ mod tests {
         );
     }
 
-    // These argument types are close to being invalid, but are actually valid.
+    // These argument types are close to being invalid, but are either okay or
```
ahl, line 1085: very helpful!

ahl (approved):
thanks for the clarifications!

dropshot_endpoint/src/endpoint.rs
sunshowers: Yeah, for trait-based servers, type aliases can't be used for `RequestContext` specifically. We need to be able to replace the type within for our purposes.

dropshot_endpoint/src/endpoint.rs
sunshowers: Good point, I'll try and change future ones at least.

sunshowers: Updated all of the ones in #984 to use `HttpResponseUpdatedNoContent` btw.

dropshot_endpoint/src/params.rs
```diff
@@ -0,0 +1,302 @@
+// Copyright 2024 Oxide Computer Company
```
sunshowers, line 1: post-accept, moved param-related code into its own file because it was getting quite big and making `endpoint.rs` harder to read. This code is also going to be mostly shared with channels in the future.

PR #1022, opened 2024-05-24
[ci] freshen up

sunshowers:
* Use ubuntu-22.04, and macos-14 for M1 runners.
* Move clippy configuration to Cargo's config rather than lib.rs or CI.
* Copy over omicron's clippy config.
* Set `CARGO_TERM_COLOR=always` for easier to read GHA output.
* Cache artifacts for quicker builds.
* Update README.

Overall, it's a nice speedup. There's still more potential for improvement via things like hakari which are more invasive.

* On Ubuntu, compared to [this baseline](https://github.com/oxidecomputer/dropshot/actions/runs/9229117299/job/25394620676) which took 4m3s, [this build](https://github.com/oxidecomputer/dropshot/actions/runs/9229402240/job/25395542313?pr=1022) reused the cache and took 2m43s, around 1.5x as fast.
* On macOS, compared to [this baseline](https://github.com/oxidecomputer/dropshot/actions/runs/9229117299/job/25394621599) which took 6m15s, [this build](https://github.com/oxidecomputer/dropshot/actions/runs/9229402240/job/25395543919?pr=1022) reused the cache and took just 2m5s. This is 3x as fast.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -60,5 +60,3 @@ jobs:
         run: cargo build ${{ matrix.feature_flags }} --locked --all-targets --verbose
       - name: Run tests
         run: cargo test ${{ matrix.feature_flags }} --locked --all-targets --verbose
-      - name: Run doctests
-        run: cargo test ${{ matrix.feature_flags }} --doc --locked --verbose
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,7 +17,6 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@9b4c13b0bfa31b4514c14f74b5a166c2708f43c6
-      - uses: taiki-e/install-action@nextest
       - name: Report cargo version
         run: cargo --version
       - name: Report rustfmt version
@@ -49,7 +48,6 @@ jobs:
             feature_flags: --all-features
     steps:
       - uses: actions/checkout@9b4c13b0bfa31b4514c14f74b5a166c2708f43c6
-      - uses: taiki-e/install-action@nextest
       - name: Report cargo version
         run: cargo --version
       - name: Report rustc version
@@ -61,6 +59,6 @@ jobs:
       - name: Build
         run: cargo build ${{ matrix.feature_flags }} --locked --all-targets --verbose
       - name: Run tests
-        run: cargo nextest run ${{ matrix.feature_flags }} --locked --all-targets --cargo-verbose
+        run: cargo test ${{ matrix.feature_flags }} --locked --all-targets --verbose
       - name: Run doctests
         run: cargo test ${{ matrix.feature_flags }} --doc --locked --verbose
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -8,3 +8,38 @@ members = ["dropshot", "dropshot_endpoint" ]
...
+# Warn on some more style lints that are relatively stable and make sense.
+iter_cloned_collect = "warn"
+iter_next_slice = "warn"
+iter_nth = "warn"
+iter_nth_zero = "warn"
+iter_skip_next = "warn"
+len_zero = "warn"
+redundant_field_names = "warn"
+# `declare_interior_mutable_const` is classified as a style lint, but it can
+# identify real bugs (e.g., declarying a `const Atomic` and using it like
```
ahl, line 30:
```suggestion
# identify real bugs (e.g., declaring a `const Atomic` and using it like
```

I think that's what you meant?

README.adoc
```diff
@@ -17,12 +17,12 @@ $ cargo +nightly doc
 
 == Contributing
 
-You can **build and run the whole test suite** with `cargo test`.  The test
-suite runs cleanly and should remain clean.
+You can **build and run the whole test suite** with https://nexte.st/[nextest]:
+`cargo nextest run`.  The test suite runs cleanly and should remain clean.
```
ahl, line 21: Is this necessary? I'm reluctant to add a dependency on a required install unless there's some overwhelming benefit.

.github/workflows/rust.yml
```diff
@@ -29,28 +33,34 @@ jobs:
...
         run: rustc --version
+      - uses: Swatinem/rust-cache@23bce251a8cd2ffc3c1075eaa2367cf899916d84 # v2.7.3
+        with:
+          # Matrix instances other than OS need to be added to this explicitly
+          key: ${{ matrix.features }}
       - name: Build
         run: cargo build ${{ matrix.feature_flags }} --locked --all-targets --verbose
       - name: Run tests
-        run: cargo test ${{ matrix.feature_flags }} --locked --all-targets --verbose
+        run: cargo nextest run ${{ matrix.feature_flags }} --locked --all-targets --cargo-verbose
```
ahl, line 64: unless there's a tremendous benefit, can se just use cargo test?

README.adoc
sunshowers: It doesn't seem to be hugely impactful for test times, but it does make the CI a lot nicer to look at I think. My worry is that at some point cargo test might break — though it's not hugely likely to happen with dropshot because it has smaller/simpler tests than something like omicron.

ymmv, happy to change the readme back, not use it in CI, or both.

Cargo.toml
sunshowers: fixed (I'd copied this from omicron's Cargo.toml -- should fix this there too)

README.adoc
sunshowers: ok, changed this back

.github/workflows/rust.yml
sunshowers: changed this back as well

ahl (approved):
thanks and my apologies for my fear of change with regard to test runners!

PR #1028, opened 2024-05-28
upgrade to hyper v1

seanmonstar:
This defines a `dropshot::Body` type that is used for both requests and
responses, in place of the now-gone `hyper::Body`.

Most of the changes are just renames. The interesting files (at least) are:

- `body.rs`: This contains the new multi-purpose `Body` type. It has a few constructors to make things nice. Otherwise, it's not that complex, but could be tweaked or optimized with more usage and/or measurements.
- `server.rs`: hyper got rid of the `Accept` trait. So now an accept loop is in this file manually. It also changed how graceful shutdown is coordinated, with a less-magical helper. It allows for adding a "graceful period timeout" if you eventually want.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -15,6 +15,16 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.11.0\...HEAD[Full list of commits]
 
+=== Breaking Changes
+
+* https://github.com/oxidecomputer/dropshot/pull/1028[#1028] Updates Dropshot for `hyper` 1.0 and `http` 1.0.  Since consumers provide Dropshot with values from `hyper` and `http`, you'll need to update to `hyper` 1.0 and `http` 1.0 (or newer compatible versions), too.
++
+**What you need to do:**
+1. Update your crate's dependencies on `hyper` and `http` to 1.0 (or a newer compatible version) in Cargo.toml.
+2. Replace any references to `hyper::Body` with `dropshot::Body` instead.
++
+There are no other known breaking changes in these crates that affect Dropshot.  If you have any trouble with this upgrade, please let us know by filing an issue.
+
 == 0.11.0 (released 2024-08-21)
 
 https://github.com/oxidecomputer/dropshot/compare/v0.10.1\...v0.11.0[Full list of commits]
@@ -36,16 +46,6 @@ generated OpenAPI spec.
 
 * Trait-based API definitions. See https://docs.rs/dropshot_endpoint/latest/dropshot_endpoint/attr.api_description.html[the documentation] for details.
 * https://github.com/oxidecomputer/dropshot/pull/1049[#1049] Added `HttpResponse::status_code()`
-
-=== Breaking Changes
-
-* https://github.com/oxidecomputer/dropshot/pull/1028[#1028] Updates Dropshot for `hyper` 1.0 and `http` 1.0.  Since consumers provide Dropshot with values from `hyper` and `http`, you'll need to update to `hyper` 1.0 and `http` 1.0 (or newer compatible versions), too.
-+
-**What you need to do:**
-1. Update your crate's dependencies on `hyper` and `http` to 1.0 (or a newer compatible version) in Cargo.toml.
-2. Replace any references to `hyper::Body` with `dropshot::Body` instead.
-+
-There are no other known breaking changes in these crates that affect Dropshot.  If you have any trouble with this upgrade, please let us know by filing an issue.
 
 == 0.10.1 (released 2024-05-15)
 
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -18,11 +18,13 @@ https://github.com/oxidecomputer/dropshot/compare/v0.11.0\...HEAD[Full list of c
 === Breaking Changes
 
 * https://github.com/oxidecomputer/dropshot/pull/1028[#1028] Updates Dropshot for `hyper` 1.0 and `http` 1.0.  Since consumers provide Dropshot with values from `hyper` and `http`, you'll need to update to `hyper` 1.0 and `http` 1.0 (or newer compatible versions), too.
-+
-**What you need to do:**
+
+==== Upgrading to hyper 1.0
+
 1. Update your crate's dependencies on `hyper` and `http` to 1.0 (or a newer compatible version) in Cargo.toml.
 2. Replace any references to `hyper::Body` with `dropshot::Body` instead.
-+
+3. You may need to update your use of `dropshot::Body`; the `http-body-util` can be helpful.
+
 There are no other known breaking changes in these crates that affect Dropshot.  If you have any trouble with this upgrade, please let us know by filing an issue.
 
 == 0.11.0 (released 2024-08-21)
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -15,6 +15,10 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.10.1\...HEAD[Full list of commits]
 
+=== Breaking Changes
+
+* https://github.com/oxidecomputer/dropshot/pull/1028[#1028] Upgraded to hyper v1.0. This means types from the `http` must come from v1.x. Dropshot no longer expects requests and responses to use the `hyper::Body` type, but rather `dropshot::Body`. It behaves very similar.
+
 == 0.10.1 (released 2024-05-15)
 
 https://github.com/oxidecomputer/dropshot/compare/v0.10.0\...v0.10.1[Full list of commits]
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -13,7 +13,29 @@
 
 == Unreleased changes (release date TBD)
 
-https://github.com/oxidecomputer/dropshot/compare/v0.10.1\...HEAD[Full list of commits]
+https://github.com/oxidecomputer/dropshot/compare/v0.11.0\...HEAD[Full list of commits]
+
+== 0.11.0 (released 2024-08-21)
+
+https://github.com/oxidecomputer/dropshot/compare/v0.10.1\...v0.11.0[Full list of commits]
+
+=== Breaking Changes
+
+* For `ApiEndpoint::register`, the `Err` variant now returns a structured `ApiDescriptionRegisterError` rather than a string.
+* https://github.com/oxidecomputer/dropshot/pull/1087[#1087] The
+`RequestContext` type now contains the `operation_id`, the name of the endpoint
+handler. This is the name of the Rust handler function, if one uses the
+`dropshot::endpoint` macro, and the value of the `operationId` field in the
+generated OpenAPI spec.
+* `TagConfig` field names have changed, for consistency with tag configuration in API traits. The `Deserialize` implementation will still work with the old field names, but the `Serialize` implementation will always produce the new field names.
+** `endpoint_tag_policy` is now called `policy`.
+** `tag_definitions` is now called `tags`.
+* https://github.com/oxidecomputer/dropshot/pull/1060[#1060] Optionally include additional header values in request log
+
+=== Other notable changes
+
+* Trait-based API definitions. See https://docs.rs/dropshot_endpoint/latest/dropshot_endpoint/attr.api_description.html[the documentation] for details.
+* https://github.com/oxidecomputer/dropshot/pull/1049[#1049] Added `HttpResponse::status_code()`
 
 === Breaking Changes
 
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,7 +17,13 @@ https://github.com/oxidecomputer/dropshot/compare/v0.10.1\...HEAD[Full list of c
 
 === Breaking Changes
 
-* https://github.com/oxidecomputer/dropshot/pull/1028[#1028] Upgraded to hyper v1.0. This means types from the `http` must come from v1.x. Dropshot no longer expects requests and responses to use the `hyper::Body` type, but rather `dropshot::Body`. It behaves very similar.
+* https://github.com/oxidecomputer/dropshot/pull/1028[#1028] Updates Dropshot for `hyper` 1.0 and `http` 1.0.  Since consumers provide Dropshot with values from `hyper` and `http`, you'll need to update to `hyper` 1.0 and `http` 1.0 (or newer compatible versions), too.
++
+**What you need to do:**
+1. Update your crate's dependencies on `hyper` and `http` to 1.0 (or a newer compatible version) in Cargo.toml.
+2. Replace any references to `hyper::Body` with `dropshot::Body` instead.
++
+There are no other known breaking changes in these crates that affect Dropshot.  If you have any trouble with this upgrade, please let us know by filing an issue.
 
 == 0.10.1 (released 2024-05-15)
 
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -55,7 +55,7 @@ version = "^0.10.2-dev"
 path = "../dropshot_endpoint"
 
 [dependencies.hyper]
-version = "1"
+version = "1.3"
 features = [ "full" ]
 
 [dependencies.hyper-util]
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -23,7 +23,8 @@ debug-ignore = "1.0.5"
 form_urlencoded = "1.2.1"
 futures = "0.3.30"
 hostname = "0.4.0"
-http = "0.2.9"
+http = "1"
+http-body-util = "0.1.1"
 indexmap = "2.2.6"
 multer = "3.1.0"
 paste = "1.0.15"
@@ -54,7 +55,11 @@ version = "^0.10.2-dev"
 path = "../dropshot_endpoint"
 
 [dependencies.hyper]
-version = "0.14"
+version = "1"
+features = [ "full" ]
+
+[dependencies.hyper-util]
+version = "0.1.5"
 features = [ "full" ]
 
 [dependencies.openapiv3]
@@ -86,8 +91,8 @@ features = [ "uuid1" ]
 async-channel = "2.3.1"
 buf-list = "1.0.3"
 expectorate = "1.1.0"
-hyper-rustls = "0.25.0"
-hyper-staticfile = "0.9"
+hyper-rustls = "0.26.0"
+hyper-staticfile = "0.10"
 lazy_static = "1.5.0"
 libc = "0.2.155"
 mime_guess = "2.0.4"
@@ -120,3 +125,4 @@ version_check = "0.9.4"
 
 [features]
 usdt-probes = ["usdt/asm"]
+
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -15,7 +15,7 @@ workspace = true
 
 [dependencies]
 async-stream = "0.3.5"
-async-trait = "0.1.81"
+async-trait = "0.1.82"
 base64 = "0.22.1"
 bytes = "1"
 camino = { version = "1.1.9", features = ["serde1"] }
@@ -23,16 +23,16 @@ debug-ignore = "1.0.5"
 form_urlencoded = "1.2.1"
 futures = "0.3.30"
 hostname = "0.4.0"
-http = "1"
-http-body-util = "0.1.1"
-indexmap = "2.4.0"
+http = "1.1.0"
+http-body-util = "0.1.2"
+indexmap = "2.5.0"
 multer = "3.1.0"
 paste = "1.0.15"
 percent-encoding = "2.3.1"
 rustls = "0.22.4"
 rustls-pemfile = "2.1.3"
 scopeguard = "1.2.0"
-serde_json = "1.0.125"
+serde_json = "1.0.128"
 serde_path_to_error = "0.1.16"
 serde_urlencoded = "0.7.1"
 sha1 = "0.10.6"
@@ -56,11 +56,11 @@ version = "^0.11.1-dev"
 path = "../dropshot_endpoint"
 
 [dependencies.hyper]
-version = "1.3"
+version = "1.4.1"
 features = [ "full" ]
 
 [dependencies.hyper-util]
-version = "0.1.5"
+version = "0.1.9"
 features = [ "full" ]
 
 [dependencies.openapiv3]
@@ -68,7 +68,7 @@ version = "2.0.0"
 features = [ "skip_serializing_defaults" ]
 
 [dependencies.serde]
-version = "1.0.208"
+version = "1.0.210"
 features = [ "derive" ]
 
 [dependencies.tokio]
@@ -89,14 +89,14 @@ version = "0.8.21"
 features = [ "uuid1" ]
 
 [dev-dependencies]
-anyhow = "1.0.86"
+anyhow = "1.0.89"
 async-channel = "2.3.1"
 buf-list = "1.0.3"
 expectorate = "1.1.0"
 hyper-rustls = "0.26.0"
 hyper-staticfile = "0.10"
 lazy_static = "1.5.0"
-libc = "0.2.157"
+libc = "0.2.158"
 mime_guess = "2.0.5"
 subprocess = "0.2.9"
 tempfile = "3.10"
@@ -105,7 +105,7 @@ trybuild = "1.0.99"
 pem = "3.0"
 rcgen = "0.13.1"
 # Used in a doc-test demonstrating the WebsocketUpgrade extractor.
-tokio-tungstenite = "0.23.1"
+tokio-tungstenite = "0.24.0"
 
 [dev-dependencies.rustls-pki-types]
 version = "1.8.0"
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "dropshot"
 description = "expose REST APIs from a Rust program"
-version = "0.10.2-dev"
+version = "0.11.1-dev"
 authors = ["David Pacheco <dap@oxidecomputer.com>"]
 edition = "2021"
 license = "Apache-2.0"
@@ -15,34 +15,35 @@ workspace = true
 
 [dependencies]
 async-stream = "0.3.5"
-async-trait = "0.1.80"
+async-trait = "0.1.81"
 base64 = "0.22.1"
 bytes = "1"
-camino = { version = "1.1.7", features = ["serde1"] }
+camino = { version = "1.1.9", features = ["serde1"] }
 debug-ignore = "1.0.5"
 form_urlencoded = "1.2.1"
 futures = "0.3.30"
 hostname = "0.4.0"
 http = "1"
 http-body-util = "0.1.1"
-indexmap = "2.2.6"
+indexmap = "2.4.0"
 multer = "3.1.0"
 paste = "1.0.15"
 percent-encoding = "2.3.1"
 rustls = "0.22.4"
-rustls-pemfile = "2.1.2"
+rustls-pemfile = "2.1.3"
 scopeguard = "1.2.0"
-serde_json = "1.0.117"
+serde_json = "1.0.125"
 serde_path_to_error = "0.1.16"
 serde_urlencoded = "0.7.1"
 sha1 = "0.10.6"
-slog = "2.5.0"
+simple-mermaid = { version = "0.1.1", optional = true }
+slog = { version = "2.5.0", features = ["dynamic-keys"] }
 slog-async = "2.8.0"
 slog-bunyan = "2.5.0"
 slog-json = "2.6.1"
 slog-term = "2.9.1"
 tokio-rustls = "0.25.0"
-toml = "0.8.14"
+toml = "0.8.19"
 waitgroup = "0.1.2"
 
 [dependencies.chrono]
@@ -51,7 +52,7 @@ features = [ "serde", "std", "clock" ]
 default-features = false
 
 [dependencies.dropshot_endpoint]
-version = "^0.10.2-dev"
+version = "^0.11.1-dev"
 path = "../dropshot_endpoint"
 
 [dependencies.hyper]
@@ -67,7 +68,7 @@ version = "2.0.0"
 features = [ "skip_serializing_defaults" ]
 
 [dependencies.serde]
-version = "1.0.203"
+version = "1.0.208"
 features = [ "derive" ]
 
 [dependencies.tokio]
@@ -80,7 +81,7 @@ optional = true
 default-features = false
 
 [dependencies.uuid]
-version = "1.9.0"
+version = "1.10.0"
 features = [ "serde", "v4" ]
 
 [dependencies.schemars]
@@ -88,17 +89,18 @@ version = "0.8.21"
 features = [ "uuid1" ]
 
 [dev-dependencies]
+anyhow = "1.0.86"
 async-channel = "2.3.1"
 buf-list = "1.0.3"
 expectorate = "1.1.0"
 hyper-rustls = "0.26.0"
 hyper-staticfile = "0.10"
 lazy_static = "1.5.0"
-libc = "0.2.155"
-mime_guess = "2.0.4"
+libc = "0.2.157"
+mime_guess = "2.0.5"
 subprocess = "0.2.9"
 tempfile = "3.10"
-trybuild = "1.0.96"
+trybuild = "1.0.99"
 # Used by the https examples and tests
 pem = "3.0"
 rcgen = "0.13.1"
@@ -106,7 +108,7 @@ rcgen = "0.13.1"
 tokio-tungstenite = "0.23.1"
 
 [dev-dependencies.rustls-pki-types]
-version = "1.7.0"
+version = "1.8.0"
 # Needed for CertificateDer::into_owned
 features = ["alloc"]
 
@@ -121,8 +123,11 @@ features = [ "max_level_trace", "release_max_level_debug" ]
 # This is required for the build.rs script to check for an appropriate compiler
 # version so that `usdt` can be built on stable rust.
 [build-dependencies]
-version_check = "0.9.4"
+version_check = "0.9.5"
 
 [features]
 usdt-probes = ["usdt/asm"]
+internal-docs = ["simple-mermaid"]
 
+[package.metadata.docs.rs]
+features = ["internal-docs"]
--- a/dropshot/src/body.rs
+++ b/dropshot/src/body.rs
@@ -9,6 +9,7 @@ use hyper::body::{Body as HttpBody, Bytes, Frame};
 
 type BoxError = Box<dyn std::error::Error + Send + Sync>;
 
+/// A body type for both requests and responses in Dropshot.
 #[derive(Debug)]
 pub struct Body {
     inner: BoxBody<Bytes, BoxError>,
--- a/dropshot/src/body.rs
+++ b/dropshot/src/body.rs
@@ -27,8 +27,8 @@ impl Body {
         Body { inner }
     }
 
-    /// Create a full body from a specific buffer.
-    pub fn full(buf: impl Into<Bytes>) -> Self {
+    /// Create a body with content from a specific buffer.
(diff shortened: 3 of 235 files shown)


jclulow:
I currently make fairly extensive use of:


```rust
Body::wrap_stream(tokio_stream::wrappers::ReceiverStream::new(rx))
```

Will that still work with the new code?

seanmonstar:
It'd be pretty straight forward to add such a constructor to `dropshot::Body`.

dropshot/src/body.rs
```diff
@@ -0,0 +1,121 @@
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use http_body_util::combinators::BoxBody;
+use http_body_util::BodyExt;
+use hyper::body::{Body as HttpBody, Bytes, Frame};
+
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
```
ahl, line 9: probably need some explanatory docs here

dropshot/src/handler.rs
```diff
@@ -139,7 +141,7 @@ impl RequestInfo {
     ///
     /// This is provided for source compatibility.  In previous versions of
     /// Dropshot, `RequestContext.request` was an
-    /// `Arc<Mutex<hyper::Request<hyper::Body>>>`.  Now, it's just
+    /// `Arc<Mutex<hyper::Request<impl Body>>>`.  Now, it's just
```
ahl, line 144: I'm not sure how to handle this. It might not be relevant any more in terms of migration. I'd like @davepacheco to look at this PR generally, but this in particular.

dropshot/src/handler.rs
```diff
@@ -67,8 +67,10 @@ use std::marker::PhantomData;
 use std::num::NonZeroU32;
 use std::sync::Arc;
 
+pub type ResponseBody = Body;
```
ahl, line 70: the use of this confused me; what's the idea here?

dropshot/src/server.rs
```diff
@@ -188,18 +183,21 @@ impl<C: ServerContext> HttpServerStarter<C> {
...
                 https.start(rx, log_close)
             }
-        }
+        };
+        /*
         .map(|r| {
             r.map_err(|e| format!("waiting for server: {e}"))?
                 .map_err(|e| format!("server stopped: {e}"))
-        });
+        })*/
```
ahl, line 191: ?

dropshot/src/server.rs
```diff
@@ -281,9 +304,11 @@ impl<C: ServerContext> InnerHttpServerStarter<C> {
...
         log: &Logger,
         handler_waitgroup_worker: waitgroup::Worker,
-    ) -> Result<InnerHttpServerStarterNewReturn<C>, hyper::Error> {
-        let incoming = AddrIncoming::bind(&config.bind_address)?;
-        let local_addr = incoming.local_addr();
+    ) -> Result<InnerHttpServerStarterNewReturn<C>, std::io::Error> {
+        let std_listener = std::net::TcpListener::bind(&config.bind_address)?;
+        std_listener.set_nonblocking(true)?;
+        let incoming = TcpListener::from_std(std_listener)?;
+        let local_addr = incoming.local_addr()?;
```
ahl, line 312: can you talk about these changes or explain in a comment?

dropshot/src/server.rs
```diff
@@ -545,18 +566,43 @@ type InnerHttpsServerStarterNewReturn<C> =
...
+                    _ = &mut close_signal => {
+                        info!(log_close, "received request to begin graceful shutdown");
+                        break;
+                    }
+                }
+            }
 
-        tokio::spawn(graceful)
+            // optional: could use another select on a timeout
+            graceful.shutdown().await
```
ahl, line 604: i.e. to return early even if shutdown() didn't complete? What would the state of the system be in that case?

dropshot/src/server.rs
```diff
@@ -597,9 +643,13 @@ impl<C: ServerContext> InnerHttpsServerStarter<C> {
         });
 
         let make_service = ServerConnectionHandler::new(Arc::clone(&app_state));
-        let server = Server::builder(https_acceptor).serve(make_service);
+        //let server = Server::builder(https_acceptor).serve(make_service);
```
ahl, line 646: we're moving this into InnerHttpsServerStarter::start effectively?

ahl (commented):
This is looking very promising. I'd love to get @sunshowers and @davepacheco to take a look. Can you please include the design document here? Also I would suggest taking a stab at migration content in the changelog as we try to help consumers figure out the path from previous versions. Thank you!

dropshot/src/body.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub(crate) fn into_data_stream(self) -> DataStream {
+        DataStream(self)
+    }
+}
+
+impl Default for Body {
+    fn default() -> Body {
+        Body::empty()
+    }
+}
```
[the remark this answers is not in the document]
adamchalmers: Adding a `Default` impl makes it easier to compose `Body` in other types that impl `Default`. For example, if you have this struct:

```
#[derive(Default)]
struct Bodies {
  inner: HashMap<i32, dropshot::Body>
}
```

You couldn't use the `derive(Default)` without the `impl Default for Body` here. You'd have to write a manual `impl Default` and set `inner` to `HashMap::new()`.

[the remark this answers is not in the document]
ahl: Yeah true: can you envision a use case for that?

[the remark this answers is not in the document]
seanmonstar: It's also used by `Client::get(url)`, where it creates a "default" body that is assumed to be empty. And I believe that was used in some tests or docs.

In hindsight, default doesn't necessarily mean empty.

dropshot/src/handler.rs
seanmonstar: This type alias can likely be completely ripped out, it was one of the first things I did to address the mountain of compiler errors (it was `type ResponseBody = BoxBody<...>` originally).

dropshot/src/server.rs
seanmonstar: It would then be an abrupt shutdown. You might have some connections that still have requests/responses in-flight, but streaming for forever. In many systems, you can't wait for forever. But if you have something that will kill the process completely, then perhaps there's less value.

dropshot/src/server.rs
seanmonstar: This is because the whole `AddrStream` thing went away, since it offered basically nothing over a `TcpStream`. This conversion dance allows binding to happen synchronously, since `tokio::net::TcpListener::bind` is an `async fn` since it allows a string argument that will trigger an async DNS lookup.

dropshot/src/extractor/body.rs
```diff
@@ -282,21 +284,20 @@ impl ExclusiveExtractor for UntypedBody {
 /// raw bytes available to the consumer.
 #[derive(Debug)]
 pub struct StreamingBody {
```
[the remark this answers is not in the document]
seanmonstar: Here's what I can see:

1. `StreamingBody` is part of Dropshot's public API. The types in `http-body-util` are very useful, but it's purposefully an unstable library to allow utilities to change if a better way is learned.
2. The `into_stream()` method seems important (and different even with hyper v0.14). It converts errors into Dropshot's `HttpError` response helper. And it provides "dumping" of body bytes.

dropshot/src/handler.rs
seanmonstar: I think that was just a string replacement that never got switched back to `dropshot::Body` since the compiler didn't notice. Reading the full docs, seems like this method is kept around for compatibility. It's saying "this is what the types used to be, but it's not anymore." :shrug:

dropshot/src/server.rs
seanmonstar: Ah, it's also commented about originally in L625 of this file, where the same thing is done to build a TLS acceptor.

dropshot/src/server.rs
seanmonstar: Yes, I just removed this line now. But hyper got rid of the `Accept` pattern, so the accept loop is done within this file.

dropshot/src/body.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Create a full body from a specific buffer.
+    pub fn full(buf: impl Into<Bytes>) -> Self {
+        let inner = http_body_util::Full::new(buf.into())
+            .map_err(|never| match never {})
+            .boxed();
+        Body { inner }
+    }
+
+    /// Wrap any body as a dropshot Body.
```
hawkw, line 37: very tiny drive by nit: should probably all be `#[must_use]`?
```suggestion
    /// Create an empty body.
    #[must_use]
    pub fn empty() -> Self {
        let inner = http_body_util::Empty::new()
            .map_err(|never| match never {})
            .boxed();
        Body { inner }
    }

    /// Create a full body from a specific buffer.
    #[must_use]
    pub fn full(buf: impl Into<Bytes>) -> Self {
        let inner = http_body_util::Full::new(buf.into())
            .map_err(|never| match never {})
            .boxed();
        Body { inner }
    }

    /// Wrap any body as a dropshot Body.
    #[must_use]
```

dropshot/src/http_util.rs
```diff
@@ -35,9 +36,11 @@ where
...
     // TODO do we need to use saturating_add() here?
     let mut nbytesread: usize = 0;
-    while let Some(maybebuf) = body.data().await {
-        let buf = maybebuf?;
-        nbytesread += buf.len();
+    while let Some(maybefr) = body.frame().await {
+        let fr = maybefr?;
+        if let Ok(buf) = fr.into_data() {
+            nbytesread += buf.len();
+        }
```
hawkw, line 43: AFAICT, this code will eat any `TRAILERS` frame, which the previous implementation wouldn't do. This is actually probably *more* correct given what this code is intended to do.

dropshot/src/http_util.rs
```diff
@@ -35,9 +36,11 @@ where
...
     }
 
     // TODO-correctness why does the is_end_stream() assertion fail?
```
hawkw, line 46: I think that the commented-out assertion here was _probably_ failing because, after discarding all the `DATA` frames, an HTTP/2 body was terminated by a `TRAILERS`, so the stream was not EOF. I wonder if this assertion can be put back, now that the new implementation will also read and discard any `TRAILERS` frame that's sent as well as the `DATA` frames.

dropshot/src/server.rs
```diff
@@ -296,9 +322,13 @@ impl<C: ServerContext> InnerHttpServerStarter<C> {
         });
 
         let make_service = ServerConnectionHandler::new(app_state.clone());
-        let builder = hyper::Server::builder(incoming);
-        let server = builder.serve(make_service);
-        Ok((InnerHttpServerStarter(server), app_state, local_addr))
+        //let builder = hyper::Server::builder(incoming);
+        //let server = builder.serve(make_service);
```
hawkw, line 326: should probably remove this?
```suggestion
```

dropshot/tests/test_tls.rs
```diff
@@ -211,31 +217,35 @@ async fn test_tls_only() {
...
-    let http_client = hyper::Client::builder().build_http();
-    let error = http_client.request(http_request).await.unwrap_err();
-    assert!(error.is_parse());
+    let http_client = hyper_util::client::legacy::Client::builder(
+        hyper_util::rt::TokioExecutor::new(),
+    )
+    .build(hyper_util::client::legacy::connect::HttpConnector::new());
+    let _error = http_client.request(http_request).await.unwrap_err();
+    // cannot check if parse error...
+    //assert!(error.is_parse());
```
hawkw, line 241: was `is_parse()` removed in `hyper` 1.0?

dropshot/src/body.rs
seanmonstar: Here's my take: I don't think `must_use` needs to be on bodies (or streams, or any other async-ish thing beyond `Future`). True, they don't do anything if not used, but most types in Rust are that way. The thing to watch out for is when a future is started and not awaited.

dropshot/Cargo.toml
```diff
@@ -54,7 +55,11 @@ version = "^0.10.2-dev"
 path = "../dropshot_endpoint"
 
 [dependencies.hyper]
-version = "0.14"
+version = "1"
```
sunshowers, line 58: Same.

dropshot/Cargo.toml
```diff
@@ -23,7 +23,8 @@ debug-ignore = "1.0.5"
 form_urlencoded = "1.2.1"
 futures = "0.3.30"
 hostname = "0.4.0"
-http = "0.2.9"
+http = "1"
```
sunshowers, line 26: Is 1 right as the minimum required version? Just double-checking.

dropshot/src/body.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Create an empty body.
+    pub fn empty() -> Self {
+        let inner = http_body_util::Empty::new()
+            .map_err(|never| match never {})
+            .boxed();
+        Body { inner }
+    }
+
+    /// Create a full body from a specific buffer.
+    pub fn full(buf: impl Into<Bytes>) -> Self {
```
sunshowers, line 31: Is `full` a term of art here? It's a little surprising to me as someone who isn't deep into this stuff.

dropshot/src/server.rs
sunshowers: Interesting about Tokio's `TcpListener::bind` being an async fn. Is making a std `TcpListener` and then converting it to Tokio the general recommendation? (I've used a similar pattern for Tokio's processes so it's not a huge surprise to me if it is.)

Could you add a comment here explaining this?

dropshot/src/server.rs
sunshowers: To confirm -- is this the same behavior as the old code -- i.e. did the old code not have a timeout on shutdown?

Can this code be combined with `InnerHttpServerStarter` above? Seems like it's complex enough (especially if we add a second select loop for this) that it's worth factoring out.

dropshot/src/server.rs
```diff
@@ -1019,8 +1056,14 @@ impl<C: ServerContext> ServerConnectionHandler<C> {
...
         ServerConnectionHandler { server }
     }
+
+    fn make_svc(&self, remote_addr: SocketAddr) -> ServerRequestHandler<C> {
+        info!(self.server.log, "accepted connection"; "remote_addr" => %remote_addr);
+        ServerRequestHandler::new(self.server.clone(), remote_addr)
+    }
 }
 
+/* TODO(SEAN)
```
sunshowers, line 1066: To confirm, this is still on your list of things to do for this PR?

dropshot/src/test_util.rs
```diff
@@ -121,8 +121,8 @@ impl ClientTestContext {
         path: &str,
         request_body: Option<RequestBodyType>,
         expected_status: StatusCode,
-    ) -> Result<Response<Body>, HttpErrorResponseBody> {
-        let body: Body = match request_body {
+    ) -> Result<Response<hyper::body::Incoming>, HttpErrorResponseBody> {
```
sunshowers, line 124: Should this be Dropshot's body type?

sunshowers (commented):
Thanks for your hard work on this!

dropshot/src/server.rs
seanmonstar: The current code in the PR is the same behavior as before. There was not a timeout before. I can remove the "optional" comment if it's confusing, I meant it as a possible addition down the line.

dropshot/src/server.rs
seanmonstar: Yea, I can copy over the comment from the other line this pattern is used.

dropshot/src/body.rs
seanmonstar: I think I first saw it years ago in netty, as something like `FullHttpMessage`. Since then, hyper has internally used `Full` as the variant for a complete, non-streaming body. And then when making the types separate, there is now `http_body_util::Full`.

dropshot/Cargo.toml
seanmonstar: Yep, not using anything newer than what was in 1.0.

dropshot/Cargo.toml
seanmonstar: I don't believe I used anything newer, but I can just stick it to 1.3. 🤷

dropshot/src/test_util.rs
seanmonstar: Sure, I can do that.

dropshot/src/extractor/body.rs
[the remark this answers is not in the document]
ahl: Makes sense.

dropshot/src/server.rs
ahl: It's fine to leave it for future generations to ponder.

dropshot/src/body.rs
ahl: @seanmonstar agreed, unless, @hawkw, you want to elaborate.

ahl:
@seanmonstar this is looking great. I would like @davepacheco to take a pass before we merge it, but, @davepacheco, feel free to interpret that with whatever degree of rigor you consider appropriate.

dropshot/src/body.rs
hawkw: :woman_shrugging: my general feeling is that all non-side-effecting constructors should be `#[must_use]` because if you construct something that doesn't have side effects and don't use it, you almost never meant to do that. but, i'm not attached to it.

davepacheco:
Sorry I'm late to this one.  I haven't had a chance to look yet.  If there are breaking changes here, as it sounds like there are, then I'd like to get the changelog update in this PR.  Our general approach is to provide super clear instructions for people to understand how to know if they're affected by a breaking change and exactly what changes to make, as concretely as we can.  [There are some examples here, from release 0.9.0](https://github.com/oxidecomputer/dropshot/blob/main/CHANGELOG.adoc#breaking-changes-2).

@ahl also mentioned a design doc?  That might help review?

davepacheco:
I forgot to add: for breaking changes, I usually like to do a test update of one or more of our important consumers.  We've found a lot of issues by doing that.  Have we converted any consumers?  Ideally it'd be great to convert Omicron's consumers -- is that too much to ask?

ahl:
> @ahl also mentioned a design doc? That might help review?

I followed up with dave privately with the document. We'll want to make sure it's recorded in this PR once we have appropriate ownership and permissions.

davepacheco (commented):
I'm still working through server.rs.

dropshot/src/server.rs
```diff
@@ -257,17 +252,41 @@ impl<C: ServerContext> InnerHttpServerStarter<C> {
...
+            // http/1 settings
+            builder.http1().timer(TokioTimer::new());
+
+            let graceful =
+                hyper_util::server::graceful::GracefulShutdown::new();
 
-        tokio::spawn(graceful)
+            loop {
+                tokio::select! {
+                    Ok((sock, remote_addr)) = self.0.accept() => {
```
davepacheco, line 271: What if `accept()` returns an error?  The [docs](https://docs.rs/tokio/1.38.0/tokio/net/struct.TcpListener.html) aren't super explicit but I gather this could happen any time the underlying accept(2) syscall returns an error, which can include transient errors like ECONNABORTED (on illumos, anyway).  EMFILE might also be considered transient.  (The docs also say "These would terminate the stream if not handled in any way."  I'm not sure what that means.  This struct is not a `Stream` and I don't know how it could know whether you handled the error.)

If this happens, according to the [`tokio::select!`](https://docs.rs/tokio/1.38.0/tokio/macro.select.html) docs:

> Once an <async expression> returns a value, attempt to apply the value to the provided <pattern>, if the pattern matches, evaluate <handler> and return. If the pattern does not match, disable the current branch and for the remainder of the current call to select!. Continue from step 3.

Step 3 waits for another of the futures to complete.  So my interpretation of all of this is: if we encounter a transient error from `accept()`, this program will stop accepting new connections.  This would be very bad.

Besides fixing this, it makes me a little nervous that we're implementing something so delicate and hard-to-test here.  Is there some existing implementation we could use here?  Would it help to turn this into a https://docs.rs/tokio-stream/0.1.15/tokio_stream/wrappers/struct.TcpListenerStream.html?

dropshot/src/server.rs
```diff
@@ -257,17 +252,41 @@ impl<C: ServerContext> InnerHttpServerStarter<C> {
...
+                        let fut = builder.serve_connection_with_upgrades(
+                            TokioIo::new(sock),
+                            self.1.make_svc(remote_addr),
+                        );
+                        let fut = graceful.watch(fut.into_owned());
```
davepacheco, line 276: I'm having a hard time understanding the various futures involved here and how the new `Graceful` mechanism works.  Would it be reasonable to add a comment describing this in more detail?  And/or using more descriptive names?

dropshot/src/server.rs
```diff
@@ -1019,38 +1058,10 @@ impl<C: ServerContext> ServerConnectionHandler<C> {
...
-        // operations as part of receiving requests).  To avoid having to ensure
-        // that conn's lifetime exceeds that of this async operation, we simply
-        // copy the only useful information out of the conn: the SocketAddr.  We
-        // may want to create our own connection type to encapsulate the socket
-        // address and any other per-connection state that we want to keep.
-        let server = Arc::clone(&self.server);
-        let remote_addr = conn.remote_addr();
-        Box::pin(http_connection_handle(server, remote_addr))
+    fn make_svc(&self, remote_addr: SocketAddr) -> ServerRequestHandler<C> {
+        info!(self.server.log, "accepted connection"; "remote_addr" => %remote_addr);
```
davepacheco, line 1063: This raised my eyebrow a bit and I think I figured out what happened.

Before this PR, both the InnerHttpServerStarter and InnerHttpsServerStarter invoked `http_connection_handle()` to log this message and produce the `ServerRequestHandler`.  This was done in the `impl Service<...> for ServerConnectionHandler` (where the type parameter was `&AddrStream` for the HTTP case and `TlsConn` for the TLS case).  This PR removes `impl Service<&AddrStream>`, but not the `impl Service<&TlsConn>`.  However, the `impl Service<&TlsConn>` is unused.  I removed it, which exposed that `http_connection_handle` is also unused now.  But this `make_svc` function I think is identical to what that function was.

So I think what I'd suggest is:
- remove the `impl Service<&TlsConn> for ServerConnectionHandler`
- remove `make_svc()` and have its callers call `http_connection_handle` instead?  Alternatively, rename this function to `make_http_request_handler` (which I find a lot more descriptive).  But the reason I had this as a naked top-level function is so that it would be obvious in the control flow at what point a new connection has been received so that if needed to put other logic there, we could do it there.

dropshot/src/server.rs
```diff
@@ -545,18 +562,43 @@ type InnerHttpsServerStarterNewReturn<C> =
...
-            info!(log_close, "received request to begin graceful shutdown");
-        });
+    ) -> tokio::task::JoinHandle<()> {
+        use hyper_util::rt::{TokioExecutor, TokioIo, TokioTimer};
+        use hyper_util::server::conn::auto;
+
+        tokio::spawn(async move {
+            let mut builder = auto::Builder::new(TokioExecutor::new());
+            // http/1 settings
+            builder.http1().timer(TokioTimer::new());
```
davepacheco, line 575: What's this doing?  Do we not need something similar for HTTP2?

seanmonstar: This tells hyper how to use a timer for any timeouts (like a header read timeout). There isn't currently a similar method for `http2()`.

dropshot/src/body.rs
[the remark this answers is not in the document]
seanmonstar: I can convert all calls to `.get()` to `.request(req_with_explicitly_empty_body)`.

dropshot/src/websocket.rs
```diff
@@ -51,11 +51,14 @@ pub type WebsocketEndpointResult = Result<Response<Body>, HttpError>;
 pub struct WebsocketConnection(WebsocketConnectionRaw);
 
 /// A type that implements [tokio::io::AsyncRead] + [tokio::io::AsyncWrite].
-pub type WebsocketConnectionRaw = hyper::upgrade::Upgraded;
+// Not public, since it uses less-stable hyper-util.
+type WebsocketConnectionRaw = hyper_util::rt::TokioIo<hyper::upgrade::Upgraded>;
```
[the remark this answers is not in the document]
seanmonstar: I don't know enough about all consumers to know if being able to name the type is important. I left it as unnamed type since it means less code, and if people need it named, it's not breaking to add a name to it like you said. 🤷

dropshot/src/body.rs
seanmonstar: I can use a different name that you all like. Is `Body::with_content("hello world")` the preferred?

dropshot/src/websocket.rs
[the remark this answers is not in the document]
davepacheco: I understand it to be a bit of an anti-pattern to have `pub` types without names because it makes it harder for consumers to work with them (e.g., what if you want to store it into a struct, or accept it as an argument to a function)?  If it's just a question of more code, I'd like to do the newtype.  That will also avoid one of the breaking changes here, right?

dropshot/src/server.rs
davepacheco: Out of curiosity, is that because HTTP2 doesn't need similar timeouts for some reason?

dropshot/src/body.rs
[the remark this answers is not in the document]
davepacheco: Ah, that's good to know.  I could go either way on this -- @ahl what do you think?

dropshot/src/server.rs
seanmonstar: Good catch, hyper did handle all those before. But as part of the removal of the `Accept` trait, due to all of it's subtle issues, that was removed as well. Something generic _might_ appear again in hyper-util, and when it does, Dropshot could use it.

For now, I can inline the behavior that used to exist.

dropshot/src/websocket.rs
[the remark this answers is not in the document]
seanmonstar: FWIW, I wouldn't call it an anti-pattern. It's a pretty helpful feature, I use it as a library author specifically in places where the traits matter, but the exact type is not important. Most people just need to "use" the value in a function. If someone (rarely) needs to store it, usually they can box it. If it's very common to store, I do give it a name.

But I've added it back here.

dropshot/src/server.rs
seanmonstar: Maybe. But as of right now, the builder doesn't have such a method.

dropshot/src/server.rs
seanmonstar: I've added an `HttpAcceptor` type here that does what hyper used to do.

dropshot/src/server.rs
seanmonstar: I added some comments.

dropshot/src/websocket.rs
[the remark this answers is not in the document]
davepacheco: Thanks!  Yeah, I can see the appeal.  I just think it's hard to know what a consumer will (or won't) want to do -- I've been on the other end of that where the author assumed I wouldn't want to store a value and I was just stuck.

@jgallagher pointed out that there has been some [discussion](https://github.com/rust-lang/api-guidelines/discussions/60) on official guidance but apparently no decision.

seanmonstar:
I believe I've addressed all the feedback. I added some breaking changes notes to the changelog. Upgrading Omicron is _currently_ out of scope for me.

dropshot/src/server.rs
```diff
@@ -327,6 +332,39 @@ impl<C: ServerContext> InnerHttpServerStarter<C> {
...
+
+                    // This could EMFILE implying resource exhaustion.
+                    // Sleep a little bit and try again.
+                    _ => {
+                        warn!(self.log, "accept error"; "error" => e);
+                        tokio::time::sleep(std::time::Duration::from_millis(
+                            100,
+                        ))
+                        .await;
+                    }
```
davepacheco, line 361: This is probably okay but I'm a little nervous about the error handling.  This is definitely an area where the standards and manual pages are good but operational experience is also relevant, which is why I'm still a little uneasy about implementing this behavior here instead of using a hardened library.  But I guess hyper 0.x was doing this inline before?  Is this the same logic it used?

---

The ECONNABORTED check is definitely right and it's one a lot of software misses so thanks for getting that. ✔️ 

I checked out the [illumos](https://illumos.org/man/3SOCKET/accept) manual page (our production systems), [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/accept.html), and [Linux](https://www.man7.org/linux/man-pages/man2/accept.2.html).  I think most of the errors fall into one of these buckets:

* "shouldn't ever happen and if they did you probably don't want to accept any more": EBADF, EFAULT, ENODEV, ENOTSOCK, EOPNOTSUPP, EPROTO, EINVAL, EAGAIN/EWOULDBLOCK
* "can totally happen and should just be retried": ECONNABORTED, EINTR
* "probably shouldn't happen, but definitely can, and there's probably not a great response but we could retry, and maybe pausing would help": EMFILE, ENOMEM, ENOSR, ENOBUFS, EPERM? (Linux)

I don't see ECONNREFUSED or ECONNRESET as possible errors that can be produced here, but the Linux manual page does say any other error for the _new_ socket can happen here.  That seems weird to me, but okay.

It seems brittle to encode all of these, so I guess the logic here is saying: assume the "can't happen" stuff really can't happen, handle ECONNABORTED[/ECONNREFUSED/ECONNRESET] explicitly as transient errors that don't require pausing, and treat everything else as a transient error, with a pause to avoid getting stuck in too tight a loop if we're wrong about it being transient.  Is that the idea?

dropshot/src/server.rs
```diff
@@ -444,26 +482,14 @@ impl HttpsAcceptor {
...
-                                    // TODO-robustness: Consider handling these
-                                    // more gracefully.
-                                    _ => {
-                                        yield Err(e);
-                                        break;
-                                    }
-                                }
+                                // The HttpAcceptor already handled transient errors.
+                                // At this point, it's a bad one.
+                                yield Err(e);
```
davepacheco, line 491: As far as I can tell, the `HttpAcceptor` can't return `Err`.  Can you make it not return a `Result` and then simplify this code path?

dropshot/src/server.rs
```diff
@@ -327,6 +332,39 @@ impl<C: ServerContext> InnerHttpServerStarter<C> {
     }
 }
 
+/// This accepts connections more gracefully from a TcpListener.
+struct HttpAcceptor {
```
davepacheco, line 336: Now that we're incorporating this pretty tricky functionality into Dropshot, we should probably have some test coverage for it.  Any ideas how to do that?  It'd be really nice if we could exercise some of those error cases but that seems hard.

dropshot/src/server.rs
```diff
@@ -271,7 +274,7 @@ impl<C: ServerContext> InnerHttpServerStarter<C> {
                     Ok((sock, remote_addr)) = self.0.accept() => {
```
davepacheco, line 274: It's still the case in the new code that _if_ the acceptor returns an error, we'll stop trying to call accept.  I guess the idea is that it's only returning persistent errors now?  I think I'd do the thing I suggested elsewhere and have the function _not_ return `Result`.  If we really want to do that, I think we could use a comment here because the issue (and the fact that this is correct) here is really subtle and would be easy to break by accident.

dropshot/src/server.rs
```diff
@@ -327,6 +332,39 @@ impl<C: ServerContext> InnerHttpServerStarter<C> {
     }
 }
 
+/// This accepts connections more gracefully from a TcpListener.
```
davepacheco, line 335:
```suggestion
/// Accepts TCP connections like a `TcpListener`, but ignores transient errors rather than propagating them to the caller
```

CHANGELOG.adoc
```diff
@@ -15,6 +15,10 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.10.1\...HEAD[Full list of commits]
 
+=== Breaking Changes
+
+* https://github.com/oxidecomputer/dropshot/pull/1028[#1028] Upgraded to hyper v1.0. This means types from the `http` must come from v1.x. Dropshot no longer expects requests and responses to use the `hyper::Body` type, but rather `dropshot::Body`. It behaves very similar.
```
davepacheco, line 20: I'm looking for more specificity here, like in the examples I linked to.  But I can take a swing at this part and just push to this PR, if that works?

davepacheco (commented):
Thanks for the updates.

CHANGELOG.adoc
davepacheco: I hope you don't mind that I just went ahead and did that.  (I'm surprised GitHub let me!)

dropshot/src/server.rs
seanmonstar: Correct, this is essentially the same code that hyper 0.14 was doing (minus some TCP keep-alive options stuff, which Dropshot didn't enable): https://github.com/hyperium/hyper/blob/0.14.x/src/server/tcp.rs#L211-L254

I do think it'd be nice to have this in hyper-util, so server frameworks don't all need a copy of this. There's a couple design questions to make it sufficiently generic (listener type, timer type, how to log, etc), such that I don't expect it to be ready to be depended on immediately.

dropshot/src/server.rs
seanmonstar: Yea, I can remove the possibility of errors. I left it in originally here in case you decided there was an error kind that should always bubble up as permanent, but then again, this code can be easily changed back in that case.

dropshot/src/server.rs
seanmonstar: I did not have a way to programmatically trigger those error kinds. Maybe if it was sufficiently generic, such that some `ErroringListener` could be provided. But that won't be available yet.

davepacheco: Okay.  How about the happy path?  What about a test that starts a Dropshot server and makes 100 or 1000 connections to it in a row?  Just thinking out loud about how best to test this automatically.

davepacheco:
I think the last items here are:

- any tests for the HttpAcceptor, if we can think of something useful
- test-converting Omicron, which @ahl is working on

ahl:
@seanmonstar it looks like Cargo.lock needs an upodate

ahl:
@seanmonstar tests are failing

ahl:
@seanmonstar looks like windows tests are hung?

ahl:
Eyyyy! Passing tests! I'm working on merging this into omicron and its vassal repos; on vacation right now, but hope to shake out any kinks in a week or so.

CHANGELOG.adoc
```diff
@@ -37,6 +37,16 @@ generated OpenAPI spec.
 * Trait-based API definitions. See https://docs.rs/dropshot_endpoint/latest/dropshot_endpoint/attr.api_description.html[the documentation] for details.
 * https://github.com/oxidecomputer/dropshot/pull/1049[#1049] Added `HttpResponse::status_code()`
 
+=== Breaking Changes
```
davepacheco, line 40: FYI it looks like this got mismerged.  It belongs in the "Unreleased changes" section I think.

ahl:
WOOT

dropshot/src/server.rs
davepacheco: @seanmonstar I was looking at this code again for other reasons and seeing if I could commonize the HTTP and HTTPS paths a bit more.  When we talked about this above you said there wasn't a similar method for HTTP2, but in the non-TLS path, there seems to be one and we use it:
https://github.com/seanmonstar/dropshot/blob/35c897b26aa0f9241c5b523a84ec009e8999ba97/dropshot/src/server.rs#L274-L275

Is there a reason these paths should be different for TLS vs. non-TLS?

seanmonstar: Huh, I have no idea what I was thinking. I clearly used it elsewhere... I checked blame on the `hyper-util` repo, in case it was a brand new addition after my saying that... nope. I'm at a complete loss.

There's no reason it should be different for TLS vs not.

davepacheco: No worries!  Thanks for the quick reply.  I'll fix this in #1121.

PR #1029, opened 2024-05-30
[7/n] [dropshot_endpoint] add tests for custom and unnamed params

sunshowers:
Noticed that channels weren't using "_dropshot_crate" sometimes, so address
that.

As part of the test, introduce a syn visitor that bans particular identifiers.
We're going to use this for trait-based servers as well.


dropshot_endpoint/src/test_util.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2023 Oxide Computer Company
```
ahl, line 1: 2024

dropshot_endpoint/src/test_util.rs
```diff
@@ -0,0 +1,44 @@
...
+            BanIdentsVisitor { idents: &idents, found: BTreeSet::new() };
+        visitor.visit_file(file);
+
+        if !visitor.found.is_empty() {
+            let found = visitor.found.into_iter().collect::<Vec<_>>();
+            panic!("banned identifiers found in file: {}", found.join(", "));
+        }
+    }
+
+    assert_banned_idents(file, idents);
```
ahl, line 25: what's the point of this inner function?

ahl (approved):
Would it make sense to have a test that fails the `assert_banned_idents` check? (apologies if I missed it)

dropshot_endpoint/src/test_util.rs
sunshowers: Avoiding excessive monomorphization on the IntoIterator -- but it's really not necessary, it's a legacy from my first attempt where I didn't just use a visitor and the function was longer. Will fix.

sunshowers:
> Would it make sense to have a test that fails the assert_banned_idents check? (apologies if I missed it)

Good idea -- no I haven't added this yet.

dropshot_endpoint/src/test_util.rs
sunshowers: Fixed.

sunshowers:
> Good idea -- no I haven't added this yet.

Done.

ahl (approved):

sunshowers:
Also added a test for unnamed params here, since we're going to fix this in an upcoming PR.

sunshowers:
Accidentally broke main 😅 fixing it in this commit because it's pretty trivial.

PR #1033, opened 2024-06-04
better document pagination interfaces

davepacheco:
This makes some small improvements based on the discussion in oxidecomputer/omicron#2449.

CC @wez

### DIFF
--- a/dropshot/src/pagination.rs
+++ b/dropshot/src/pagination.rs
@@ -184,9 +184,11 @@ impl<ItemType> ResultsPage<ItemType> {
 /// `PageSelector` types.
 ///
 /// `ScanParams` describes the set of querystring parameters that your endpoint
-/// accepts for the _first_ request of the scan (typically: filters and sort
-/// options).  This must be deserializable from a querystring.  These parameters
-/// usually don't change across the multiple requests that make up a scan.
+/// accepts for the _first_ request of the scan.  This must be deserializable
+/// from a querystring.  These parameters typically describe filters and sort
+/// options that will not change across the multiple request that make up the
+/// scan.  (Clients won't have to pass the scan parameters for subsequent
+/// requests, though; instead, you generally encode them into `PageSelector`.)
 ///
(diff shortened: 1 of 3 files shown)


david-crespo (approved):
clear improvement!

dropshot/src/pagination.rs
```diff
@@ -182,7 +185,8 @@ impl<ItemType> ResultsPage<ItemType> {
 ///
 /// `ScanParams` describes the set of querystring parameters that your endpoint
 /// accepts for the _first_ request of the scan (typically: filters and sort
-/// options).  This must be deserializable from a querystring.
+/// options).  This must be deserializable from a querystring.  These parameters
+/// usually don't change across the multiple requests that make up a scan.
```
ahl, line 189: I thought query parameters needed to be encoded in the token and weren't supplied past the first invocation. Perhaps I'm confused.

ahl (commented):
lgtm

dropshot/src/pagination.rs
davepacheco: I think we have the same understanding and I'm trying to work out the confusion.

To be clear: for the first request, there's no token and the scan parameters come in via normal query parameters.  I think that's what the existing text about being "deserializable from a querystring" refers to.  For subsequent requests, there's generally _only_ a query parameter for the token, which encodes (among other things) the scan parameters so that they'll be available (and the same) for subsequent requests.

I added this sentence here to mean "the set of things that go into the scan parameters, like the filter options and sort mode, do not change across the scan -- i.e., if the sort mode was 'name ascending' for the first request then it will be 'name ascending' the whole time".  Did you read this as "clients should pass the same values for these query parameters across all requests (i.e., `sort_mode`)" (which is, indeed, wrong)?  i.e., I'm saying "the sort mode doesn't change", not "the `sort_mode` query parameter has the same value".  I'll push a change to clear this up.

wez:
Thank you!

PR #1035, opened 2024-06-10
[8/n] [dropshot-endpoint] add tests for missing attributes

sunshowers:
This is a bit of a basic/elementary test but I realized we didn't have coverage
for it. (And it's actually important for server traits -- specifically, without
extra work, incorrect span information is produced.)


ahl (approved):
more tests > fewer tests

PR #1036, opened 2024-06-11
[10/n] [dropshot_endpoint] move some shared code to common modules

sunshowers:
This is pure code movement without any functional changes.

We're going to reuse most of this code for our new, improved channel
implementation shortly.

Also make a couple small changes to `do_endpoint_inner` and `EndpointParams`.
The analogous code structure is required for channels, and it's nice to
have the two implementations be as close as possible.


ahl (commented):
Since this is ("just") code movement I'm not going to look at it, but I'm happy to if that would be useful. I think the only useful review I could do would be a careful diff of the new and old code and that doesn't seem particularly valuable.

PR #1038, opened 2024-06-12
[13/n] [dropshot_endpoint] support channels in API traits

sunshowers:
Building on the work in the last few PRs, this is relatively straightforward.
The bulk of the PR is simply tests for bad channels -- we now have mirrored
tests for all four combinations of (macro kind, endpoint kind).


sunshowers:
@ahl whenever you're back a review on this would be really appreciated. It is mostly tests with ~500 lines of real code. thanks!

dropshot_endpoint/src/api_trait.rs
```diff
@@ -1182,22 +1251,165 @@ impl<'ast> ApiEndpoint<'ast> {
...
+                    "endpoint `{name_str}` must be of the form \
+                     #[channel {{ protocol = WEBSOCKETS, path = \"/path\", ... }}]"
+                ),
+            ));
+            return None;
+        }
+    };
+
+    // TODO: Switch to from_tokenstream_spanned once
+    // https://github.com/oxidecomputer/serde_tokenstream/pull/194 is available.
```
ahl, line 1275: I think it is, right?

ahl (approved):
I didn't look at this very closely -- I kind of figure the tests speak for themselves. If there are parts that you think require a more fine-grained review, please let me know

dropshot_endpoint/src/api_trait.rs
sunshowers: Yeah, I missed removing this. Will do, thanks!

PR #1039, opened 2024-06-12
[dropshot_endpoint] unify error message style

sunshowers:
Always use messages of the form "endpoint `{name_str}` ...", without a period
at the end. The casing and lack of period at the end matches rustc's style.

One difference from that style is that we now include the name of the endpoint
in error messages. That's particularly important for trait-based Dropshot
servers, where in some cases we might end up losing span info and producing
messages that don't include which endpoint they're talking about. Include the
name of the endpoint unconditionally for a unified style, and so that we don't
have to have unnecessary conditionals. (I personally found having the endpoint
name to be friendlier anyway.)


sunshowers:
(Accidentally landed https://github.com/oxidecomputer/dropshot/pull/1009 on the wrong branch, putting this up again here.)

PR #1050, opened 2024-07-05
[12/n] [dropshot] improve API endpoint registration error messages

sunshowers:
Currently, the bulk builder functions created by the `api_description` macro
don't provide information about which endpoint(s) failed to register. This
leads to a pretty suboptimal devx with large APIs like Nexus.

To address this, use a structured error type, and also carry information about
the operation ID (endpoint name) that was registered. Use this to produce
better messages for both `ApiDescription::register`, and the bulk builder
functions created by the `api_description` macro.


dropshot/src/api_description.rs
```diff
@@ -972,34 +979,74 @@ impl<Context: ServerContext> ApiDescription<Context> {
...
     }
 }
 
+impl std::error::Error for ApiDescriptionBuildErrors {}
+
+/// An error that occurred while registering an individual API endpoint.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ApiDescriptionRegisterError {
+    operation_id: String,
+    message: String,
```
sunshowers, line 1023: Rather than storing an unstructured string, we could model the different kinds of errors as an enum if that makes sense. That would also let tests not have to do the somewhat brittle string matching they have to do today.

sunshowers: We had a meeting yesterday and we decided to not do this for now -- tracking endpoint names (operation IDs) in a structured manner is very useful, but the actual message less so. But we can definitely do this in the future if required.

ahl: I think both statements are true here: enums are generally better **and** in this case, no one is going to unpick these messages programmatically. It is definitely an improvement to have the operations as explicit strings.

ahl (approved):

dropshot/src/api_description.rs
sunshowers: Agreed on both counts.

PR #1059, opened 2024-07-09
[14/n] [dropshot_endpoint] support `tag_config` in API traits

sunshowers:
Based on today's discussion, we're supporting a static `tag_config` argument to
`api_description`. The implementation just creates a `dropshot::TagConfig` and
sets that.


dropshot_endpoint/src/api_trait.rs
```diff
@@ -154,13 +157,61 @@ impl ApiMetadata {
...
+/// A mirror of dropshot's `TagConfig`, used as part of arguments to the
+/// top-level `api_description` macro.
+#[derive(Deserialize, Debug)]
+struct ApiTagConfig {
+    #[serde(default)]
+    allow_other_tags: bool,
+    #[serde(default)]
+    endpoint_tag_policy: ApiEndpointTagPolicy,
+    // tag_definitions is required
+    tag_definitions: HashMap<String, ApiTagDetails>,
```
ahl, line 169: how about `tags` ? I think it would be ok to change `dropshot::TagConfig` if we want them

dropshot_endpoint/src/api_trait.rs
```diff
@@ -154,13 +157,61 @@ impl ApiMetadata {
...
 }
 
+/// A mirror of dropshot's `TagConfig`, used as part of arguments to the
+/// top-level `api_description` macro.
+#[derive(Deserialize, Debug)]
+struct ApiTagConfig {
+    #[serde(default)]
+    allow_other_tags: bool,
+    #[serde(default)]
+    endpoint_tag_policy: ApiEndpointTagPolicy,
```
ahl, line 167: similarly: do we want this to be `policy` ... ? I agree that it's nice to match TagConfig... but this seems potentially verbose

dropshot_endpoint/src/lib.rs
```diff
@@ -127,32 +127,102 @@ pub fn channel(
...
+/// }
+/// # // defining fn main puts the doctest in a module context
+/// # fn main() {}
+/// ```
+///
+/// ### Tag configuration
+///
+/// With Dropshot, individual endpoints can have OpenAPI tags associated with
+/// them. For these tags, trait-wide configuration can be specified using the
+/// `tag_config` argument to `api_description`.
```
ahl, line 162:
```suggestion
/// Endpoints can have tags associated with
/// them that appear in the OpenAPI document. These provide a means of organizing an API. 
Use the
/// `tag_config` argument to `api_description` to specify tag information for the trait.
```

I think this might be clearer? Feel free to ignore or adapt.

dropshot_endpoint/src/lib.rs
```diff
@@ -127,32 +127,102 @@ pub fn channel(
...
+///
+/// `tag_config` is optional. If not specified, the default is to allow all
+/// tags.
```
ahl, line 165:
```suggestion
/// `tag_config` is optional. If not specified, the default is to allow any
/// tag value and any number of tags associated with each endpoint.
```

dropshot_endpoint/src/lib.rs
```diff
@@ -127,32 +127,102 @@ pub fn channel(
...
+///
+/// The shape of `tag_config` is broadly similar to that of
+/// `dropshot::TagConfig`. It has the following fields:
+///
+/// * `tag_definitions`: A map of tag names to information about them.
+///   _Required, but can be empty._
```
ahl, line 171:
```suggestion
/// * `tag_definitions`: A map of tag names with information about them.
///   _Required, but can be empty._
```

dropshot_endpoint/src/lib.rs
```diff
@@ -127,32 +127,102 @@ pub fn channel(
...
+///
+///   The keys are tag names, which are strings. The values consist of:
```
ahl, line 173:
```suggestion
///   The keys are tag names, which are strings. The values are objects that consist of:
```

dropshot_endpoint/src/api_trait.rs
```diff
@@ -154,13 +157,61 @@ impl ApiMetadata {
...
+    // tag_definitions is required
+    tag_definitions: HashMap<String, ApiTagDetails>,
+}
+
+/// A mirror of dropshot's `EndpointTagPolicy`, used as part of arguments to the
+/// top-level `api_description` macro.
+#[derive(Clone, Copy, Default, Deserialize, Debug)]
+#[serde(rename_all = "snake_case")]
```
ahl, line 175: are there other places where we use snake_case in a similar setting?

dropshot_endpoint/src/lib.rs
```diff
@@ -127,32 +127,102 @@ pub fn channel(
...
+///       _Optional._
+///     * `url`: The URL for the external documentation. _Required._
+///
+/// * `allow_other_tags`: Whether to allow tags not explicitly defined in
+///   `tag_definitions`. _Optional, defaults to false. But if `tag_config` as a
+///   whole is not specified, all tags are allowed._
+///
+/// * `endpoint_tag_policy`: Must be one of the following:
+///   * `at_least_one`: Endpoints are required to have at least one tag, or
+///     possibly more.
```
ahl, line 188:
```suggestion
///   * `at_least_one`: Endpoints are required to have at least one tag, and
///     may have more.
```

dropshot_endpoint/src/lib.rs
```diff
@@ -127,32 +127,102 @@ pub fn channel(
...
+///   * `exactly_one`: Endpoints are required to have exactly one tag.
+///   * `any`: Endpoints may have any number of tags.
```
ahl, line 190:
```suggestion
///   * `any`: Endpoints may have any number of tags (and may have no tags at all).
```

dropshot_endpoint/src/lib.rs
```diff
@@ -127,32 +127,102 @@ pub fn channel(
...
+///     }]
+///     async fn put_test(
+///         rqctx: RequestContext<Self::MyContext>,
+///     ) -> Result<HttpResponseUpdatedNoContent, HttpError>;
+/// }
+/// # // defining fn main puts the doctest in a module context
+/// # fn main() {}
+/// ```
+///
+/// ### Tag configuration
```
ahl, line 158: I'm not exactly sure where to put this, but I think we should say something like:

Compliance with tag configuration is evaluated when generating the OpenAPI document output (is that right?). A failure to comply--for example, if `endpoint_tag_policy = `at_least_one` and some endpoint has no associated tags--results in a failure during OpenAPI document generation that identifies the endpoints that are out of compliance.

ahl (approved):

dropshot_endpoint/src/api_trait.rs
sunshowers: > I think it would be ok to change dropshot::TagConfig if we want them

Yeah I was thinking about suggesting that, I think I value consistency in names highly enough that if we're changing to `tags` here (something I think is a good idea) we should also change `dropshot::TagConfig`.

All right, I'll make that change.

sunshowers: Done.

dropshot_endpoint/src/api_trait.rs
sunshowers: Also done.

dropshot_endpoint/src/api_trait.rs
sunshowers: Hmm, I don't think so -- we use snake_case for keys but don't have any other places where enum variants are tagged as values. `snake_case` looks more correct to me here though, do you think otherwise?

dropshot_endpoint/src/lib.rs
sunshowers: Noted this as `(including none at all)`.

dropshot_endpoint/src/lib.rs
sunshowers: Compliance is actually evaluated while registering endpoints, so the failure will happen in the `api_description` and `stub_api_description` functions. Definitely worth noting though.

sunshowers: Added a paragraph.

dropshot_endpoint/src/api_trait.rs
ahl: I guess I don't have an opinion. It seemed odd to me... based on my knowledge of the implementation of serde_tokenstream, not as an end-user. Maybe show it to a couple of unbiased people and see if they have a preference?

sunshowers: Ran a poll on Mastodon which overwhelmingly agreed with you: https://hachyderm.io/@rain/112832705113530863

But one suggestion was to use `EndpointTagPolicy::AtLeastOne`, i.e. the actual value in there. I think that should be possible with `ParseWrapper`, right?

sunshowers: Implemented this and I really like it, but at what cost: https://github.com/oxidecomputer/serde_tokenstream/pull/195

PR #1060, opened 2024-07-12
optionally include additional header values in request log

jclulow:
I would like to be able to log additional headers in some applications, like **X-Forwarded-For**, which is potentially critical for programs that sit behind a reverse proxy.  This change allows the consumer to list additional headers for inclusion in request log records.


davepacheco:
I'm finding myself wishing we had a better solution for the more general thing here (i.e., allowing consumers to record whatever information they would find useful at various points in the server and request lifecycle).  In the meantime, this seems reasonable.  I was a little worried about the new dynamic-keys slog feature but it looks like Omicron already enables that.

PR #1068, opened 2024-07-24
Preserve blank lines in endpoint descriptions

david-crespo:
Currently we turn blank lines in a doc comment into a single `\n`. When we then try to treat that text as markdown (e.g., on the docs site), markdown ignores single blank lines. This is the difference between this

image

and this

image

So here we keep those blank lines as a double `\n\n`.

### DIFF
--- a/dropshot_endpoint/src/doc.rs
+++ b/dropshot_endpoint/src/doc.rs
@@ -53,7 +53,8 @@ impl ExtractedDoc {
                         // Continuation lines and newlines.
                         format!("{}{}", acc, comment)
                     } else if comment.is_empty() {
-                        // Handle fully blank comments as newlines we keep.
+                        // Blank lines get a markdown paragraph break (unless
+                        // acc already ends in '\n' -- see above)
                         format!("{}\n\n", acc)
                     } else {
                         // Default to space-separating comment fragments.
--- a/dropshot_endpoint/src/doc.rs
+++ b/dropshot_endpoint/src/doc.rs
@@ -54,7 +54,7 @@ impl ExtractedDoc {
                         format!("{}{}", acc, comment)
                     } else if comment.is_empty() {
                         // Handle fully blank comments as newlines we keep.
-                        format!("{}\n", acc)
+                        format!("{}\n\n", acc)
                     } else {
                         // Default to space-separating comment fragments.
                         format!("{} {}", acc, comment)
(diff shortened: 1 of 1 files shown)


david-crespo:
Tagging you @sunshowers for fun since we just talked about an unrelated doc comment parsing issue [here](https://github.com/oxidecomputer/omicron/pull/6145#discussion_r1688534939).

david-crespo:
I regenerated the Nexus OpenAPI JSON using this code and confirmed the comment I added in https://github.com/oxidecomputer/omicron/pull/5914 is the only one affected by this change.

dropshot_endpoint/src/doc.rs
```diff
@@ -227,13 +227,21 @@ mod tests {
...
         struct SummaryDescriptionBreak;
         assert_eq!(
             ExtractedDoc::from_attrs(&SummaryDescriptionBreak::schema().attrs),
             ExtractedDoc {
                 summary: Some("Summary".to_string()),
-                description: Some("Text More\nEven More".to_string())
+                description: Some(
+                    "Text More\n\nEven More\n\nAnd another paragraph"
+                        .to_string()
+                )
```
sunshowers, line 244: Huh, I don't quite understand -- were other doc comments not similarly impacted by this? Or have we been working around this for a while?

Also, does Rust collapse down multiple blank lines? Naively I'd have expected six `\n` instances here.

david-crespo: At least in the external API, until I added one yesterday, we had zero endpoints that included a paragraph break other than the one between the first line and the rest.

Regarding the collapsing, I did not know why it was doing that, so I'm glad you asked and made me figure it out. The reason it's skipping subsequent blank lines is that we're hitting the `ends_with('\n')` on line 50, so we're just concatenating `line`, which is `""` in the case of a blank line, so it's a noop.

https://github.com/oxidecomputer/dropshot/blob/7b594d01f47ca783c5d4a25ca2b256602580fe92/dropshot_endpoint/src/doc.rs#L46-L65

While this is correct in outcome, it is very hard to tell that from looking at it, so I'll try tweaking the code.

david-crespo: Comment in b868ba22b85e6fca2b51c8d589b3c5d4712482ab is the best I could do :|

sunshowers: Gotcha -- so we just look at `acc` and do the collapsing there.

sunshowers (approved):
Looks good to me! Though I'm not sure if Adam or Dave also want to have a look at it.

david-crespo:
Considering the effect in Nexus is so limited, I’m just gonna go for it.

PR #1082, opened 2024-08-05
[dropshot-endpoint] with API traits, add _mod to default module name

sunshowers:
One of the recurring issues I ran into with using API traits is that I'd
define, e.g. `NexusInternalApi` in a crate called `nexus-internal-api`. The
default module name would then clash with the crate name, leading to unhelpful
errors.

As a workaround I've been adding `_mod` to the end of support module names, and
I think it makes sense for `_mod` to be the default. It's a little strange at
first, but it makes sense if you think about it and is quite inoffensive.


sunshowers:
Don't think this needs much looking at, so I'm going to go ahead and land this.

PR #1115, opened 2024-09-24
support for specifying version numbers in API endpoints

### LINKED ISSUE #869: server support for multiple API versions
@ahl already started this in #862.  Here are some more detailed notes on the plan we discussed a few weeks ago.

## Motivation

We use Dropshot in Omicron to provide OpenAPI/HTTP APIs between many components.  We want to be able to evolve these APIs in both compatible and incompatible ways.  We also want to do rolling upgrades of these components without downtime.  With JSON of course it's fairly easy to phrase most API changes backwards-compatibly: have the implementation ignore any unknown fields, make any new fields optional, don't change the semantics of existing fields, etc.  These approaches essentially make any new API a superset of the previous one.  One can evolve APIs with Dropshot this way today.  This approach has two big disadvantages: (1) it can be hard to know that you've done this correctly -- that you haven't accidentally made some change that _will_ cause requests from older clients to fail. (2) you lose a lot of the advantages we enjoy today around strong typing on the server side, the spec, and the client side.  In the limit after a lot of changes we might expect most of the API to wind up being a bunch of `Option`s of complex enum types for all variants that previous clients might have sent.

Instead, we've proposed this basic approach:

- each version of an API has an explicit version number or string (this is true already)
- clients always specify which API version they want to use
- servers support multiple API versions and pick which one to use based on what the client asks for (that's what this issue is all about)
- when we build the image used to deploy any component (whether a client or server of a dropshot API) we include metadata about what APIs the component provides _and_ consumes
- the deployment system ensures that when deploying anything, there already exist components that provide whatever APIs (and versions) the component depends on

In this way, you can roll out a breaking change to an API by:

- starting with server and clients at the old version
- rolling out a version of the server that supports both the old and new version
- rolling out a version of the clients that use the new version
- rolling out a version of the server that supports only the new version

This issue is all about the "servers support multiple API versions" item above.  While on some level this can be achieved very easily (essentially copy/paste all the endpoint registrations multiple times), we want an approach following our usual guidelines: it should be easy to do the right thing and hard to accidentally do the wrong thing.  If you want to make a change to an API, breaking or not, it shouldn't require duplicating the whole server and it should be very hard to accidentally break an earlier version of the API.

For those with access, there's more on this in Oxide [RFD 421](https://rfd.shared.oxide.computer/rfd/421).

## Proposal

- Each endpoint can be tagged with a semver range.  See #862 for an example.
- `ApiDescription` becomes `MultiVersionApiDescription` (logically, if not literally).
    - As endpoints are registered, they're organized in the `router` data structure by version.
    - When a request comes in, the server will use the client-requested version (if provided) when routing the request to a handler.
    - The `ApiDescription` will be able to spit out an API spec for _any_ semver.  It does the same thing it does today, just using the version information to determine if a given endpoint (and its types) should be included in the spec.

Today in consumers like Omicron today, we check the current API spec into the repo and use an [expectorate](https://github.com/oxidecomputer/expectorate) test to avoid accidentally changing it (i.e., to give us a chance to review any potentially breaking changes).  We've largely been okay with breaking most of our APIs so far, so we've just allowed this file to evolve as needed.

The proposal for consumers would be:

- Instead of checking in just one spec, there'd be a directory of supported spec versions.  For each one, we'd have Dropshot re-generate the spec for that version (from the current code) and verify that it matches the checked-in spec.  It'd be nice to just say that we'd never expect one of these older spec files to change, but in practice, we might need to allow changes here but limit it to provably-compatible changes (e.g., doc changes).
- When you want to make a breaking change to an API, you:
    - Copy the latest file to a new version.
    - Make whatever changes you want to the dropshot server.  If you're adding a new endpoint, mark it added in your new version.  If you're removing an old endpoint, you just mark it removed in your new version.  If you want to change an input type, you could do this by "removing" the old endpoint that used that type and "adding" one that uses the new type.
    - Run the tests.  The old specs should not change.  The new spec should differ only in the ways you expect.
    - Note that you can do this even if the most recent version was never "released" to customers and we don't care about officially supporting it.  This way, development is no different from release, and the whole mechanism works (and is exercised) in dev/test, too.
- You can remove support for an old version at any time by just removing the old spec file and any endpoints that are no longer used by any version.
    - You can remove support for the _oldest_ version once you know that clients are all upgraded past it in all deployments.
    - You can also choose to remove support for intermediate versions if you know they've never been released to supported systems.

This is a very general approach.  It should be possible to implement any sort of breaking change this way.  We've discussed other approaches that might make it easier to make specific kinds of breaking changes, but they're a lot murkier both in terms of implementing them and whether they'd support the kinds of changes we need.

davepacheco:
Supersedes #862 (and is based on it).  Fixes #869.

Also based on #1127.  This includes a test that uses that option to make sure that operations have the same names in different versions, even when they have different handler function names.


davepacheco:
I think this one is nearly ready to go (but after #1127).  The one change I'm leaning towards is that I think `versions = A..B` ought to _not_ include `B`, for consistency with Rust's range syntax.  I think people will be confused to discover that it actually means the thing that Rust denotes with `A..=B`.

davepacheco:
@jgallagher and @leftwo raised the excellent point that earlier versions of this PR would cause a Dropshot server to happily serve requests for API versions newer than the server was built with.  For now, we can make this the responsibility of the `DynamicVersionPolicy` impl.  Dropshot ships with only one `impl`, `ClientSpecifiesVersionInHeader`.  I've changed this one to require a max version and reject requests newer than that.

Stepping back: my goal for this work is to provide a minimal base for us to go prototype versioning in some real consumers.  I think it's pretty likely that as we do that, we will find changes we want to make to this.  I wouldn't be surprised if we wind up with some common impls of `DynamicVersionPolicy` that we want to fold back into Dropshot.  But I don't want to spend too much energy now trying to guess what those things are -- let's see what shakes out of the consumer impls.

dropshot_endpoint/src/metadata.rs
```diff
@@ -156,6 +162,14 @@ pub(crate) struct ValidatedEndpointMetadata {
     unpublished: bool,
     deprecated: bool,
     content_type: ValidContentType,
+    versions: VersionRange,
+}
+
+fn semver_parts(x: &semver::Version) -> (u64, u64, u64) {
+    // This was validated during validation.
```
ahl, line 169: seems valid

dropshot/src/server.rs
```diff
@@ -194,20 +198,30 @@ impl<C: ServerContext> HttpServerStarter<C> {
             })
             .transpose()?;
         let handler_waitgroup = WaitGroup::new();
+
+        let router = api.into_router();
+        if let VersionPolicy::Unversioned = version_policy {
+            if router.has_versioned_routes() {
+                return Err(BuildError::UnversionedServerHasVersionedRoutes);
```
ahl, line 205: is this the point of `has_versioned_routes` i.e. is it mostly here to produce an error message if the user has done something aberrant?

dropshot/tests/test_versions.rs
```diff
@@ -0,0 +1,521 @@
...
+fn test_versions_openapi() {
+    let api = api();
+
+    for version in ["0.9.0", "1.0.0", "1.1.0", "1.3.1", "1.4.0"] {
+        let semver: semver::Version = version.parse().unwrap();
+        let mut found = Cursor::new(Vec::new());
+        api.openapi("Evolving API", semver).write(&mut found).unwrap();
+        let actual = std::str::from_utf8(found.get_ref()).unwrap();
+        expectorate::assert_contents(
+            &format!("tests/test_openapi_v{}.json", version),
```
ahl, line 333: what do you think about 
```suggestion
            &format!("tests/test_versions_v{}.json", version),
```
or something to help associate json with rs?

dropshot_endpoint/src/metadata.rs
```diff
@@ -239,6 +285,87 @@ impl ValidatedEndpointMetadata {
...
+                        TokenTree::from(earliest.token()),
+                        TokenTree::from(latest.token()),
+                    ]
+                    .into_iter()
+                    .collect();
+                    return Err(syn::Error::new_spanned(
+                        span,
+                        String::from(
+                            "semver range (from ... until ...) has the \
+                             endpoints out of order",
```
ahl, line 357: from version is after until version?

dropshot_endpoint/src/metadata.rs
```diff
@@ -239,6 +285,87 @@ impl ValidatedEndpointMetadata {
...
+                let latest_semver = parse_semver(&latest)?;
+                if latest_semver < earliest_semver {
+                    let span: TokenStream = [
+                        TokenTree::from(earliest.token()),
+                        TokenTree::from(latest.token()),
+                    ]
+                    .into_iter()
+                    .collect();
+                    return Err(syn::Error::new_spanned(
+                        span,
```
ahl, line 354: I think we want 
```suggestion
                        input.span(),
```

Maybe? In particular I'm thinking of this test:

dropshot/tests/fail/bad_version_backwards.stderr
```
error: semver range (from ... until ...) has the endpoints out of order
  --> tests/fail/bad_version_backwards.rs:17:16
   |
17 |     versions = "1.2.3".."1.2.2"
   |    
```

Should we be underlining both?

dropshot_endpoint/src/metadata.rs
```diff
@@ -239,6 +285,87 @@ impl ValidatedEndpointMetadata {
...
+    From(semver::Version),
+    Until(semver::Version),
+    FromUntil(semver::Version, semver::Version),
+}
+
+fn parse_semver(v: &syn::LitStr) -> syn::Result<semver::Version> {
+    v.value()
+        .parse::<semver::Version>()
+        .map_err(|e| {
+            syn::Error::new_spanned(v, format!("expected semver: {}", e))
```
ahl, line 300: what do Semver::parse() errors look like? Are those more useful than something like "{} is not of the form MAJOR.MINOR.PATCH"?

ahl (commented):
very nice

ahl:
Do we definitely want to allow users to specify the version number header (i.e. in `ClientSpecifiesVersionInHeader`)? Or is this flexibility they don't need? If we allow this to be configurable, we'll need a similar sort of configuration when generating clients.

We could document (in OpenAPI) this optional header for all API endpoints; that seems quite annoying. Alternatively we could invent an extension to tell progenitor if there's a version header and--if so--what to call it.

dropshot/tests/test_versions.rs
davepacheco: Done.

dropshot_endpoint/src/metadata.rs
davepacheco: This doesn't quite work:

```
error[E0277]: the trait bound `proc_macro2::Span: ToTokens` is not satisfied
   --> dropshot_endpoint/src/metadata.rs:354:25
    |
353 |                     return Err(syn::Error::new_spanned(
    |                                ----------------------- required by a bound introduced by this call
354 |                         input.span(),
    |                         ^^^^^^^^^^^^ the trait `ToTokens` is not implemented for `proc_macro2::Span`
    |
    = help: the following other types implement trait `ToTokens`:
              &'a T
              &'a mut T
              Abstract
              AndAnd
              AndEq
              AngleBracketedGenericArguments
              Arm
              As
            and 308 others
note: required by a bound in `syn::Error::new_spanned`
   --> /Users/dap/.cargo/registry/src/index.crates.io-6f17d22bba15001f/syn-2.0.79/src/error.rs:189:27
    |
189 |     pub fn new_spanned<T: ToTokens, U: Display>(tokens: T, message: U) -> Self {
    |                           ^^^^^^^^ required by this bound in `Error::new_spanned`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `dropshot_endpoint` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

I poked around a bit but did not immediately see how to turn that span into something that impls `ToTokens` (besides what the code is already doing).

dropshot_endpoint/src/metadata.rs
davepacheco: Right.  Are you suggesting changing the message?  I'm fine with that (although I think it's a little hard to parse, using "from" and "until" as adjectives, especially if you're not already in the headspace of these arguments being called the "from version" and the "until version").

dropshot_endpoint/src/metadata.rs
davepacheco: I think the messages it produces are okay.  Here are a few examples:
https://github.com/oxidecomputer/dropshot/blob/e747169e4f2b834c0a03cd7411e38cf9fad64b38/dropshot_endpoint/src/endpoint.rs#L1070-L1085
https://github.com/oxidecomputer/dropshot/blob/e747169e4f2b834c0a03cd7411e38cf9fad64b38/dropshot_endpoint/src/endpoint.rs#L1094-L1109

davepacheco:
> Do we definitely want to allow users to specify the version number header (i.e. in `ClientSpecifiesVersionInHeader`)? Or is this flexibility they don't need? If we allow this to be configurable, we'll need a similar sort of configuration when generating clients.
> 
> We could document (in OpenAPI) this optional header for all API endpoints; that seems quite annoying. Alternatively we could invent an extension to tell progenitor if there's a version header and--if so--what to call it.

We discussed this offline and concluded:

- We'll pick a name that ought to be fine for all of our own stuff (probably `api-version`).
- Progenitor will use this name by default, but the behavior will be configurable so that you can have it not specify anything or use a different header name.
- Dropshot will make it easy to use the same header name.  I'm being vague about the specifics here because I expect this part of the Dropshot interface to potentially change a bit when we start incorporating it into consumers so I plan to flesh this out when we do that.

dropshot_endpoint/src/metadata.rs
ahl: I found the suggestion that the numbers are out of order potentially misleading. e.g. if I have 1.2.3..1.2.2 and what I actually meant was 1.2.3..1.2.12

ahl (approved):

davepacheco:
I considered leaving this open until I'd done the consumer-side workflow.  But with all the other work going on in Dropshot, I don't think it makes sense to maintain this as a feature branch.  Nor is it worth a compile-time feature to exclude it.  I'm going to just mark it as experimental for now (in documentation) and land it.

davepacheco:
I've run into a few flakes like oxidecomputer/omicron#7051 but I believe they're unrelated.

PR #1127, opened 2024-10-02
allow endpoints and channels to specify the `operation_id`

davepacheco:
By default, we pick the function's name as the OpenAPI `operation_id`.  This allows people to specify their own.  This will become more important after #1115.


ahl (approved):
looks good; good tests.

PR #1137, opened 2024-10-11
[dropshot_endpoint] add allow(dead_code) to output trait

### LINKED ISSUE #1129: spurious dead code warning with trait-based API definition
With this code (a version of the code from #1128 where I've moved the trait definition to the top level):

```rust
#[dropshot::api_description]
trait MyApi {
    type Context;
}
fn main() {
    let description = my_api_mod::stub_api_description().unwrap();
    let spec = description.openapi("Counter Server", "1.0.0");
    println!(
        "{}",
        serde_json::to_string_pretty(&spec.json().unwrap()).unwrap()
    );
}
```

When I run it, it works, but I get this when compiling it:

```
dap@zathras dropshot-operation-id $ cargo version
cargo 1.81.0 (2dbb1af80 2024-08-20)
dap@zathras dropshot-operation-id $ cargo check --example=api-trait-broken
    Checking dropshot v0.12.1-dev (/Users/dap/oxide/dropshot-operation-id/dropshot)
warning: trait `MyApi` is never used
 --> dropshot/examples/api-trait-broken.rs:2:7
  |
2 | trait MyApi {
  |       ^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

warning: `dropshot` (example "api-trait-broken") generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.56s
```

I get that technically the trait itself is unused but I think from a Dropshot user's perspective this warning is spurious and we should silence it if we can.

sunshowers:
In general, we should treat the fact that our macro was invoked on the trait as
enough of a use.

Fixes #1129.

### DIFF
--- a/dropshot_endpoint/src/api_trait.rs
+++ b/dropshot_endpoint/src/api_trait.rs
@@ -365,8 +365,27 @@ impl<'ast> ApiParser<'ast> {
         let mut supertraits = item_trait.supertraits.clone();
         supertraits.push(parse_quote!('static));
 
+        // Also add dead_code if the visibility is not `pub`. (If it is `pub`,
+        // then it is expected to be exported, and so the dead code warning
+        // won't fire.)
+        //
+        // Why check for non-`pub` visibility? Because there is a downside to
+        // allow(dead_code): it also applies to any items defined within the
+        // trait. For example, if a provided method on the trait defines an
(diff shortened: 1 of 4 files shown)


ahl (approved):
looks good!

PR #1164, opened 2024-11-05
[WIP] Support error types other than `dropshot::HttpError`

### LINKED ISSUE #7: openapi: deal with error return types
We may do this in several PRs, but the end state should be that the various error types are well-captured for each endpoint. This will likely require us to think more carefully about the specific return types we want to expose to developers, the appropriate granularity, and the types of remediation we expect developers to take in the case of failures.

There are types of failures where we have more information:
- we could provide detailed information about input validation failures
- we could provide specific guidance about retries in the case of 503 (service unavailable)
- etc.

Ideally we should make use of the `components/responses` field of OpenAPI to minimize the number of times these shared error structures are represented in the OpenAPI document.

### LINKED ISSUE #39: could use customisation of error presentation
One imagines many dropshot servers will be straight REST APIs, where clients are just libraries or commands and uniform JSON error messages make sense.  Some, however, may also serve other artefacts; e.g., a web server that had an API component, but also an interactive HTML component that a browser may access with some pages and images and such.

Presently dropshot will produce a stock uniform JSON error response for, say, endpoints that do not map to a handler.  If the client sends an `Accept` header that suggests they understand or prefer a HTML response, it'd be good to be able to choose somehow to render a reasonable 404 page or event a 302 redirect to the right place for browsers.

### LINKED ISSUE #41: error codes for dropshot-provided errors need work
As it says in `error.rs`, we envision that errors from an API typically have a string code (separate from the HTTP status code):
https://github.com/oxidecomputer/dropshot/blob/308be874160569eb8cd6752c04216a914e650f2b/dropshot/src/error.rs#L52-L69

I've generally assumed that the namespace of these codes needs to be under the control of the Dropshot consumer, since they will likely be defining most of the codes for their own application-specific conditions.  However, Dropshot currently generates a bunch of different errors of its own (e.g., when a request arrives for a path that has no handler associated with it).  What code should it use?

I'm thinking out loud through a bunch of options here:

1. Have Dropshot define a set of codes for itself and let consumers use whatever codes they want, too.  This kind of sucks.  It's hard to expand Dropshot's set later since we might stomp on a code that consumers are already using.  We could use a prefix to carve out part of the namespace, but this is part of the consumer's public API -- that's ugly.  Plus, it's hard for Dropshot to make judgments about what error cases a consumer's client wants to distinguish.
2. Let the consumer define the entire namespace and allow consumers to define the codes that Dropshot will use for its own errors.  Maybe the caller provides a function that maps HTTP status codes to Strings (or something that we can turn into a String).  Or maybe Dropshot defines an enum for the conditions it needs to generate codes for and consumers provide a function that maps from that.  We could even provide a default implementation that would hopefully be suitable for most consumers.  This gives us most of the benefits of (1) but without the expansion problem -- if we expand it, if the consumer uses a `match`, they'll get a compile error that forces them to decide how to map the new condition.
3. Have Dropshot own the entire namespace: create a set of codes like `ObjectNotFound`, `BadArgs`, etc. and require that consumers use these.  This might actually be nice because many consumers will wind up using a lot of the same codes, but it seems like a non-starter that consumers can't extend this.

The behavior today is that the code is optional and Dropshot generally provides `None`.  That was basically a hack to be able to make forward progress -- it's pretty crappy for consumers and their users.

See also #39.

### LINKED ISSUE #801: Add ability to set error response headers
I currently have response headers working for CORS, after looking at how Oxide does so in `cio`: https://github.com/bencherdev/bencher/blob/main/services/api/src/util/headers.rs
I think this may solve: https://github.com/oxidecomputer/dropshot/issues/57

But there is no way to set _error_ response headers.

The headers for a valid request look like this (happy CORS):
```
access-control-allow-headers | Content-Type
access-control-allow-methods | GET
access-control-allow-origin | *
content-length | 255
content-type | application/json
date | Sun, 15 Oct 2023 23:43:19 GMT
x-request-id | 46e28335-f418-4731-ae94-a73588908f63
```

But the headers for an error request look like this (unhappy CORS):
```
content-length | 133
content-type | application/json
date | Sun, 15 Oct 2023 23:43:33 GMT
x-request-id | 0c32030a-a999-4b42-9a0f-653696564e26
```

Internally, it looks like `dropshot` uses this `HttpError` `into_response` method: https://docs.rs/dropshot/latest/src/dropshot/error.rs.html#237
So it would be great to be able to add additional headers to `HttpError` or just have it also use the headers set for a valid request.

hawkw:
This still needs some of documentation and API polish, but it works!

Fixes #7
Fixes #39 
Fixes #41
Fixes #801


dropshot/src/server.rs
```diff
@@ -73,6 +72,37 @@ pub struct DropshotState<C: ServerContext> {
     /// Worker for the handler_waitgroup associated with this server, allowing
     /// graceful shutdown to wait for all handlers to complete.
     pub(crate) handler_waitgroup_worker: DebugIgnore<waitgroup::Worker>,
+    /// Determines how to construct responses for server errors.
+    pub(crate) error_handler: Box<
+        dyn Fn(ErrorContext<'_, C>, ServerError) -> Response<Body>
+            + Send
+            + Sync,
+    >,
```
hawkw, line 80: I'm not sure if this is really the nicest interface for the "unhandled dropshot-internal error" API --- open to suggestions here!

hawkw: Another option I considered was having the user provide a _type_ when constructing the server, and bounding it with `IntoErrorResponse + From<ServerError>`, and having the `From` impl for that type construct the error. That way, we could also be aware of the schema for the global catchall error...but, it's trickier to provide request metadata that way if it's just a `From` impl. Potentially, we could instead have this handler return a type implementing `IntoErrorResponse`, but we'd then have to erase that type unless we wanted the server to be generic over  it..

dropshot/src/error.rs
```diff
@@ -50,6 +50,89 @@ use serde::Serialize;
...
+pub use crate::extractor::ExtractorError;
+pub use crate::extractor::MultipartBodyError;
+pub use crate::extractor::QueryError;
+pub use crate::extractor::StreamingBodyError;
+pub use crate::extractor::TypedBodyError;
+pub use crate::http_util::PathError;
+pub use crate::router::RouterError;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ServerError {
```
hawkw, line 62: This is the top-level representation of "all errors generated from within dropshot" which the user can construct their own error responses from using the server-level error handler. Alternatively, if the error handler is not overridden, we will just convert into the stock `HttpError` for you by default.

I also considered calling this `DropshotError` --- open to being convinced either way.

hawkw: Something else I considered was having a `fn suggested_status_code()` or something on this and the nested child errors, which would return the same status code we use when converting them to `HttpError`s. This would be in order to help users construct their own error response types without having to match every possible variant of these errors. You could then use a combination of the `suggested_status_code()` and `to_string()` methods to construct a user error type, and potentially match specific variants and handle them differently.

But, this API can be added in a separate PR.

dropshot/src/error.rs
```diff
@@ -50,6 +50,89 @@ use serde::Serialize;
...
+/// This trait is implemented by [`HttpError`], and may alternatively be
+/// implemented by user-defined error types.
+pub trait IntoErrorResponse: JsonSchema {
+    /// Returns the HTTP status code that should be returned along with this
+    /// error.
+    fn into_error_response(
+        &self,
+        request_id: &str,
+    ) -> http::Response<crate::Body>;
+}
```
hawkw, line 120: This is the trait that must be implemented by error types returned by endpoint functions. Initially, they were just going to be `T: Serialize + JsonSchema + AsStatusCode` (where `AsStatusCode` is a trait where the user can indicate the desired status code for the error, naturally), but I realized that we might also want to provide the request ID when turning user error types into error responses.

I'm wondering about unifying this interface with that of the server's global handler for dropshot-internal errors, but it's tricky: we need to erase the types of the errors returned by individual endpoint functions, in order to allow them to return different error types without forcing the entire server to be generic over every possible error, but we'd like the global error handler function to be able to match on the dropshot-internal errors. I thought about having this function take take the `ErrorContext` type used by the global error handler thingy, but that's currently generic over the `ServerContext`, and would therefore make this trait no longer object-safe, defeating the purpose a bit.

dropshot/src/handler.rs
```diff
@@ -330,26 +335,31 @@ macro_rules! impl_HttpHandlerFunc_for_func_with_params {
...
+        type Error = ErrorType;
         async fn handle_request(
             &self,
             rqctx: RequestContext<Context>,
             _param_tuple: ($($T,)*)
         ) -> HttpHandlerResult
         {
             let response: ResponseType =
-                (self)(rqctx, $(_param_tuple.$i,)*).await?;
+                (self)(rqctx, $(_param_tuple.$i,)*).await.map_err(|e| HandlerError::Handler(Box::new(e)))?;
```
hawkw, line 362: It's necessary to erase the user's error type here, so that handler functions with arbitrary error return values can coexist in the same API.

We could alternatively do that by just going ahead and serializing the error response here. However, I didn't do that, because I thought it would be nicer to be able to give back a thing that also implements `fmt::Display`, for when we log that the request completes with an error. Thus, the `HandlerError` enum.

hawkw (commented):
Leaving some notes for reviewers.

dropshot/src/api_description.rs
```diff
@@ -914,11 +931,15 @@ impl<Context: ServerContext> ApiDescription<Context> {
...
                     openapiv3::ReferenceOr::Item(response),
                 );
+                let ApiEndpointErrorMetadata { ref name, ref schema } =
+                    endpoint.error_type;
+                let err_ref = openapiv3::ReferenceOr::Reference {
+                    reference: format!("#/components/responses/{name}"),
+                };
+                errors
+                    .entry(endpoint.error_type.name.clone())
+                    .or_insert_with(|| schema(&mut generator));
```
hawkw, line 941: We stuff the schemas for each endpoint's error type into a big map, so that we can put the schemas in `components.responses` in order to allow endpoints that return the same Rust error type to also return the same type in the OpenAPI document.

Unfortunately, though, I think just keying this by the name of the schema is probably not sufficient, because I believe `schemars`' `derive(JsonSchema)` just uses the type's name, _not_ its fully qualified name. Which, I think, means that if you have an endpoint that returns a `foo::bar::Error` and another endpoint that returns a `baz::quux::Error`, and these are different Rust types, their schemas will clobber each other. The key for the map should probably be the fully-qualified type name including the module path, and we might need to massage that into something we can use as the reference in the OpenAPI doc.

I think this needs to be fixed before this PR is mergeable; I'm leaving this comment to get it written down while I remember.

hawkw: Went and implemented this in e9728118742f6891bf0c5f427fc93bcd471c8003 --- it's a bit hairy because I wanted to generate reasonably nice unambiguous names for the error responses. It turns out that `schemars` will internally disambiguate _schema names_ by appending a number to them, which I had missed, but it's still necessary to do our own disambiguation as well if we want to reference schemas in `components.schemas` from `components.responses`, since we still have to generate the references to them ourselves. 

It would certainly be nicer if there was a way to get `schemars` to use my prettier unambiguous names for the schemas themselves rather than `Error1`, `Error2`,  but I think this is good enough:

```json
{
  "openapi": "3.0.3",
  "info": {
    "title": "Custom Error Example",
    "version": "1.0"
  },
  "paths": {
    "/pdflatex/{filename}": {
      "get": {
        "operationId": "get_pdflatex",
        "parameters": [
          {
            "in": "path",
            "name": "filename",
            "required": true,
            "schema": {
              "type": "string"
            }
          }
        ],
        "responses": {
          "201": {
            "description": "successful creation",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/Pdf"
                }
              }
            }
          },
          "4XX": {
            "$ref": "#/components/responses/LatexError"
          },
          "5XX": {
            "$ref": "#/components/responses/LatexError"
          }
        }
      }
    },
    "/ye-flask": {
      "get": {
        "summary": "Gets ye flask.",
        "operationId": "get_ye_flask",
        "responses": {
          "200": {
            "description": "successful operation",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/Flask"
                }
              }
            }
          },
          "4XX": {
            "$ref": "#/components/responses/FlaskError"
          },
          "5XX": {
            "$ref": "#/components/responses/FlaskError"
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Error": {
        "description": "Errors returned by pdflatex.\n\nGood luck figuring out what these mean!",
        "oneOf": [
          {
            "description": "An hbox is overfull.",
            "type": "object",
            "properties": {
              "OverfullHbox": {
                "type": "object",
                "properties": {
                  "badness": {
                    "description": "The amount of badness in the overfull hbox.",
                    "type": "integer",
                    "format": "uint",
                    "minimum": 0
                  }
                },
                "required": [
                  "badness"
                ]
              }
            },
            "required": [
              "OverfullHbox"
            ],
            "additionalProperties": false
          },
          {
            "description": "Like an overfull hbox, except the opposite of that.",
            "type": "object",
            "properties": {
              "UnderfullHbox": {
                "type": "object",
                "properties": {
                  "badness": {
                    "description": "This one also has badness.",
                    "type": "integer",
                    "format": "uint",
                    "minimum": 0
                  }
                },
                "required": [
                  "badness"
                ]
              }
            },
            "required": [
              "UnderfullHbox"
            ],
            "additionalProperties": false
          }
        ]
      },
      "Error2": {
        "type": "object",
        "properties": {
          "reason": {
            "type": "string"
          }
        },
        "required": [
          "reason"
        ]
      },
      "Flask": {
        "type": "object",
        "properties": {
          "name": {
            "type": "string"
          }
        },
        "required": [
          "name"
        ]
      },
      "Pdf": {
        "type": "object",
        "properties": {
          "filename": {
            "type": "string"
          }
        },
        "required": [
          "filename"
        ]
      }
    },
    "responses": {
      "LatexError": {
        "description": "Error",
        "content": {
          "application/json": {
            "schema": {
              "$ref": "#/components/schemas/Error"
            }
          }
        }
      },
      "FlaskError": {
        "description": "Error",
        "content": {
          "application/json": {
            "schema": {
              "$ref": "#/components/schemas/Error2"
            }
          }
        }
      }
    }
  }
}
```

dropshot/src/server.rs
hawkw: See also comments in https://github.com/oxidecomputer/dropshot/pull/1164#discussion_r1830222468

hawkw:
I think I'm gonna go sketch out the "recommended headers" thing I suggested in https://github.com/oxidecomputer/dropshot/pull/1164/commits/f54bca937ca46d8fc18cc7786c66dc9146e63619 as well, while I'm working on the dropshot error types.

hawkw:
Another question is whether we ought to generate the "global error" that's synthesized by the error handler from Dropshot errors in the OpenAPI document for each endpoint, since --- implicitly --- the endpoint can also return that error in the face of e.g. extractor errors. We could potentially generate a `oneOf` in the OpenAPI document that indicates that the error response can either be the endpoint-specific error *or* the global error type. But this would require getting rid of the error handler function and instead having the user say "this is the _type_ I want to return, and it can be constructed from a `ServerError`".

dropshot/src/lib.rs
```diff
@@ -742,7 +742,7 @@ mod dtrace;
 mod api_description;
 mod body;
 mod config;
-mod error;
+pub mod error;
```
davepacheco, line 745: Are you thinking users would use this as `dropshot::error::Something`?  We could organize stuff that way but it seems different from what we've done with everything else in Dropshot.

dropshot/src/extractor/mod.rs
```diff
@@ -23,6 +26,70 @@ pub use path::Path;
...
+
+use crate::error::HttpError;
+pub use crate::http_util::PathError;
+
+/// Errors returned by extractors.
+///
+/// Because extractors can be combined (a tuple of types which are extractors is
+/// itself an extractor), all extractors must return the same error type. Thus,
+/// this is an enum of all the possible errors that can be returned by any
+/// extractor.
```
davepacheco, line 42: I'm not sure about this approach here.  These are just the extractors that Dropshot provides.  But consumers can define their own, too.

Also, it seems like a lot of the actual error cases probably overlap (e.g., you can have "invalid input" in a typed body, path param, query param, etc. and it's not clear a consumer should have to handle these all separately).

nit: it seems like this belongs in dropshot/src/extractor/common.rs (or a new file) -- mod.rs is just a thing that includes other files and "common" is where we put the common stuff here.

dropshot/examples/colliding-error-type-names.rs
```diff
@@ -0,0 +1,165 @@
...
+                    write!(f, "underfull hbox, badness {badness}")
+                }
+            }
+        }
+    }
+
+    #[endpoint {
+    method = GET,
+    path = "/pdflatex/{filename}",
+}]
```
davepacheco, line 68: nit: formatting is weird here

dropshot/examples/colliding-error-type-names.rs
```diff
@@ -0,0 +1,165 @@
...
+use dropshot::ConfigLoggingLevel;
+use dropshot::RequestContext;
+use dropshot::ServerBuilder;
+
+mod latex {
+    use super::*;
+
+    /// Errors returned by pdflatex.
+    ///
+    /// Good luck figuring out what these mean!
```
davepacheco, line 18: This might just be me but when reading this I found myself confused about what any of this stuff is before I realized that these are just opaque tokens you might know if you'd fought `pdflatex` recently.  Take it or leave it but it might be clearer if the example came from the space that dropshot consumers are likely to know already.

dropshot/examples/colliding-error-type-names.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    impl dropshot::error::IntoErrorResponse for Error {
+        fn into_error_response(
+            &self,
+            ctx: dropshot::error::ErrorContext<'_>,
+        ) -> http::Response<dropshot::Body> {
+            http::Response::builder()
+                .status(http::StatusCode::BAD_REQUEST)
+                .header("x-request-id", ctx.request_id)
```
davepacheco, line 42: I see what you were saying -- yeah it really seems like we want Dropshot to do this rather than expect everybody will remember to do this stuff.

dropshot/examples/colliding-error-type-names.rs
```diff
@@ -0,0 +1,165 @@
...
+    #[endpoint {
+        method = GET,
+        path = "/ye-flask",
+    }]
+    pub(crate) async fn get_ye_flask(
+        _rqctx: RequestContext<()>,
+    ) -> Result<dropshot::HttpResponseOk<Flask>, Error> {
+        Err(Error { reason: "can't get ye flask!".to_string() })
+    }
+}
```
davepacheco, line 139: What is this example adding?

edit: I think I realized this one is trying to show what happens if you use the same error type name in multiple places?  I would have thought it'd just be a test case and not an example.  Is that something we'd expect a user to wonder how to do and consult this example for?

dropshot/examples/colliding-error-type-names.rs
```diff
@@ -0,0 +1,165 @@
+// Copyright 2024 Oxide Computer Companyy
+
+//! Example of an API using a user-defined error type to serialize error
+//! representations differently from [`dropshot::HttpError`].
```
davepacheco, line 4: This seems maybe not right but copy/pasted from the other example?

dropshot/src/api_description.rs
```diff
@@ -652,6 +672,53 @@ impl<Context: ServerContext> ApiDescription<Context> {
...
+        // This means that if one API endpoint returns a `foo::Error`, and
+        // another one returns a `bar::Error`, and we use the schema name
+        // according to schemars as the key for referencing the schema, both
+        // endpoints will return the same schema in the OpenAPI document, but
+        // the actual Rust types may serialize according to totally different
+        // schemas. Therefore, we must ensure that if two error return types
+        // have the same Rust type name, we instead refer to their schemas by
+        // the fully-qualified module path, rather than just the type name.
+        // However, this is ugly, so we would prefer to only do it when it's
+        // necessary to disambiguate stuff.
```
davepacheco, line 694: Doesn't this apply to regular types, too?  I think you run into the same thing if you have two ordinary return types (or types used in types) whose schema name is the same (often, the Rust type basename is the same, but they're in different modules).  Are errors different here?

dropshot/src/server.rs
```diff
@@ -73,6 +73,40 @@ pub struct DropshotState<C: ServerContext> {
...
+    // easily simplify it by just stuffing it in a type alias --- we'd have to
+    // add our own trait that's implemented for `T: Fn(ErrorContext<'_, C>,
+    // ServerError) -> Response<Body> + Send + Sync`, which has its own downsides...
+    #[allow(clippy::type_complexity)]
+    pub(crate) error_handler: Box<
+        dyn Fn(ErrorContext<'_>, ServerError) -> Response<Body> + Send + Sync,
+    >,
+}
+
+impl<C: ServerContext + std::fmt::Debug> std::fmt::Debug for DropshotState<C> {
```
davepacheco, line 87: nit: could use `DebugIgnore` if we're just trying to ignore the new field?

dropshot/src/server.rs
```diff
@@ -73,6 +73,40 @@ pub struct DropshotState<C: ServerContext> {
     /// Worker for the handler_waitgroup associated with this server, allowing
     /// graceful shutdown to wait for all handlers to complete.
     pub(crate) handler_waitgroup_worker: DebugIgnore<waitgroup::Worker>,
+    /// Determines how to construct responses for server errors.
+    // Since most of the complex type here is the `dyn Fn` trait, we can't
+    // easily simplify it by just stuffing it in a type alias --- we'd have to
+    // add our own trait that's implemented for `T: Fn(ErrorContext<'_, C>,
```
davepacheco, line 79: Out of curiosity, why not just make this a trait object?  That seems like a clean way to let consumers provide what they want here.

davepacheco (commented):
Thanks for taking this on.  I think this is a great direction but as you said there are some important details still to work out.  I have some specific comments below.  I'm interested in @ahl's take on the OpenAPI bits.

My biggest questions are around what this looks like for the end user, which I kind of lost track of in all the low-level details here.  It'd help to write up the end-user documentation (which, for better or worse, I think the rest of Dropshot puts into the crate-level docs) and the CHANGELOG entry.  I think it's also worth spelling out somewhere (even if it's just in a PR comment) how this looks in the OpenAPI doc and maybe generated clients (if it's not obvious).

This is definitely one of those changes where we'll want to do a test-upgrade of Omicron and make sure it does what we think.

When this is closer to ready, I'll be curious if it will address @jclulow's goals, too.

This probably sounds like a lot of work.  These aren't all blockers.  But this is a pretty substantive change to both the Rust API and the resulting specs and it's important to get right.  On the plus side, once we do, this is going to be a *really* nice improvement to a longstanding pain point!

dropshot/examples/colliding-error-type-names.rs
hawkw: Sure, I'm happy to change the example, or change the docs to make it clearer that I just picked random error strings from my memory, as the errors themselves were supposed to be context-free. I'd be happy to contrive an example that's actually doing some kind of (basic) validation, or something; I started by sketching out examples mostly as a way to ensure that code using the new APIs compiled, and I think it's not a bad idea to go back and make them more useful for actually learning how to use the APIs.

davepacheco:
Two more thoughts:

- It's worth noting that the changes to the Extractor traits are breaking changes for consumers.  That may be unavoidable, given the goals.
- Conceivably we could separate some of this into smaller PRs that we could land right away.  I'm thinking of the parts that use structured errors for the Router and some other request handling paths.  I initially thought that would make the interesting bits of this PR easier to focus on but as I write this I think most of the structured errors are in areas that are tricky (the extractors) or related to this PR (the handler errors) so there may not be _that_ much that could be separated here.

dropshot/src/error.rs
```diff
@@ -50,6 +50,123 @@ use serde::Serialize;
...
 
+pub use crate::extractor::ExtractorError;
+pub use crate::extractor::MultipartBodyError;
+pub use crate::extractor::QueryError;
+pub use crate::extractor::StreamingBodyError;
+pub use crate::extractor::TypedBodyError;
+pub use crate::http_util::PathError;
+pub use crate::router::RouterError;
+
+#[derive(Debug, thiserror::Error)]
```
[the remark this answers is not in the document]
hawkw: Yup, definitely agree. I haven't written docs for most of the new stuff yet because I wanted to make sure the API direction was broadly right before spending a bunch of time documenting it, but I'm sorry if I made that hard to review!

dropshot/src/server.rs
hawkw: What do you mean? The current approach is a trait object --- it's a `Box<dyn Fn(ErrorContext<'_>, ServerError) -> Response<Body> + Send + Sync>`. The comment is regarding how `clippy` dislikes the type _of the boxed trait object_; we could give the type a shorter name by introducing a trait with a blanket impl (`impl<T> ErrorHandler for T where T: Fn(...) -> ...)`) and changing the type to `Box<dyn ErrorHandler + Send + Sync>`, but I didn't love the idea of introducing a trait with a blanket impl just to make the clippy warning go away.

dropshot/src/server.rs
hawkw: good call, didn't realize there was a dependency on that

dropshot/examples/colliding-error-type-names.rs
hawkw: This was really intended as a test case. As far as I could tell, there doesn't seem to be much infrastructure for testing that generated OpenAPI docs are well-formed, but I may have overlooked it. I wrote this as an example just to test this edge case but was hoping to turn it into a proper test before merging this.

dropshot/src/server.rs
davepacheco: Sorry, what I meant was defining a user-facing trait called `ErrorResponseSerializer` or something like that with one method which has the same signature as this function and then this would just be `Box<dyn ErrorResponseSerializer>`.  I wouldn't do it for the clippy warning but because it seems a little cleaner.  (I don't feel strongly about it.)

dropshot/examples/colliding-error-type-names.rs
davepacheco: Makes sense but yeah there are definitely expectorate tests for generated OpenAPI docs (e.g., `test_openapi.rs`).

dropshot/src/server.rs
hawkw: Yeah, that was what I was saying I didn't do in the comment. Broadly, my preference has been to _avoid_ such "trait aliases" (traits that are blanket implemented for all `T: SomeOtherTrait`) in public APIs, since I've seen folks get confused a lot by APIs that expect some trait that you can't actually implement and can only implement by implementing another trait. But, I'm open to being convinced otherwise, if you think that it's better!

dropshot/src/lib.rs
hawkw: I'm fine with changing this to be not-`pub` and reexporting everything at the root level!

dropshot/src/extractor/mod.rs
hawkw: > I'm not sure about this approach here. These are just the extractors that Dropshot provides. But consumers can define their own, too.

So, one thought is that we could provide an `Other(Box<dyn Error + Send + 'static>)` variant for user-implemented extractors? I hadn't realized that the extractor traits were meant to be publicly implemented.

> Also, it seems like a lot of the actual error cases probably overlap (e.g., you can have "invalid input" in a typed body, path param, query param, etc. and it's not clear a consumer should have to handle these all separately).

So, I also wondered about changing these to be more about the semantic error than the specific extractor that generated it. This approach is partially because some of the extractor errors represent things that have recommended HTTP semantics (e.g. a missing or invalid `Content-Type` header is supposed to return a 415 Unsupported Media Type status). But, I think we can just special-case those as part of a single `ExtractorError` that's shared across all extractors. I'll rework this a bit!

dropshot/src/server.rs
davepacheco: I don't mean a trait alias with a blanket impl for functions with this signature.  I mean that the consumer would use their own `struct MySerializer` and explicitly `impl ErrorResponseSerializer for MySerializer` and impl the trait's function and then they'd pass that struct into the builder that constructs the Dropshot server.

dropshot/src/extractor/mod.rs
davepacheco: > > I'm not sure about this approach here. These are just the extractors that Dropshot provides. But consumers can define their own, too.
> 
> So, one thought is that we could provide an `Other(Box<dyn Error + Send + 'static>)` variant for user-implemented extractors? I hadn't realized that the extractor traits were meant to be publicly implemented.

That seems like it would technically work but feels like second-classing them a bit.  The two things I mentioned combined made me think that organizing these variants by the type of extractor isn't great.  That makes me want to go down the other path you mentioned, organizing this more semantically.  But did you go down that path and find it problematic, too?

hawkw: Not exactly, I think organizing the error type variants semantically is fine. What I meant is just that I think it's important that _some_ extractor errors have first class representations (e.g. being enum variants that are easy to match on) because they have specific recommended response status codes/headers. This includes content-type errors and the websocket errors. Basically everything else can probably be grouped into a generic 400 Bad Request + message, or a generic 500 Internal Server Error.

dropshot/src/server.rs
hawkw: Oh, I see what you mean. Hm, yeah, that could also work.

dropshot/src/extractor/mod.rs
hawkw: I think the real question becomes: "beyond those errors which we know ought to have specific response statuses etc, what categories of errors might users want to programmatically match on?". Pretty much every error that _we_ return that isn't content type related is "some kind of parse error", so we could have an `InvalidInput` variant or something, and perhaps there should be an internal error variant for "the extractor is broken".  I'm on the fence as to whether those variants ought to be represented by `Box<dyn Error>` or by dropshot's `HttpError`...

davepacheco: For the Dropshot-provided errors, I think the two use cases are:

- programmatic consumers will treat them either like generic network-like errors (similar to a 500) or an unhandleable error that shouldn't happen (like a 400).  Their real goal from this work is being able to get a typed error for the errors returned by handlers, not the Dropshot ones.
- the other main consumer is people who want to serve a custom web page to a browser (@jclulow wants this and I think @david-crespo has too).  I'm not sure if they need much programmatic information about what went wrong.  (They probably _do_ want, like, the URL that was requested, though -- they might want to redirect to something that depends on it.)

Maybe the way to go is to start pretty generic (similar to `HttpError`: status code, internal message, external message) and extend it based on what people then ask for.

hawkw: Yeah, that makes sense; I'm inclined to potentially still represent it as an enum internally because I'd like some of the error variants to also be able to provide a "recommended headers" interface lazily, but I think it seems reasonable to wrap an enum behind an opaque type that just says "here's my recommended status code, internal/external message, and a method that adds headers associated with the error to a `HeaderMap`.

PR #1173, opened 2024-11-12
[3/n] rename request_body_max_bytes to default_request_body_max_bytes

sunshowers:
In upcoming work (#1171), we're going to add optional per-endpoint size limits.
To make it clear that per-endpoint limits always override whatever's specified
in the configuration, rename this field.

(Without the rename, it can be confusing to determine what happens if the
per-endpoint limit is smaller than the configuration limit.)

Also add a deserialize-time check to fail if `request_body_max_bytes` is
specified, guiding users to the new name.

Depends on:

* #1170
* #1172


dropshot/src/config.rs
```diff
@@ -113,9 +116,108 @@ impl Default for ConfigDropshot {
...
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(default)]
+struct DeserializedConfigDropshot {
+    bind_address: SocketAddr,
+    default_request_body_max_bytes: usize,
+    // Previous name for default_request_body_max_bytes. Present only to guide
+    // users to the new name.
+    #[serde(
+        deserialize_with = "deserialize_invalid_request_body_max_bytes",
+        skip_serializing_if = "Option::is_none"
```
ahl, line 135: I think this could be simply `skip_serializing` in that it has to be None

dropshot/src/config.rs
```diff
@@ -113,9 +116,108 @@ impl Default for ConfigDropshot {
...
+    fn default() -> Self {
+        ConfigDropshot::default().into()
+    }
+}
+
+/// A marker type to indicate that the configuration is invalid.
+///
+/// This type can never be constructed, which means that for any valid config,
+/// `Option<InvalidConfig>` is always none.
+#[derive(Clone, Debug, PartialEq, Serialize)]
```
ahl, line 175: do we need Serialize?

dropshot/src/config.rs
```diff
@@ -113,9 +116,108 @@ impl Default for ConfigDropshot {
...
+pub enum InvalidConfig {}
+
+fn deserialize_invalid_request_body_max_bytes<'de, D>(
```
ahl, line 178: do we prefer this to `impl Serialize for InvalidConfig`?

README.adoc
```diff
@@ -45,11 +45,13 @@ include:
 |No
 |Specifies that the server should bind to the given IP address and TCP port.  In general, servers can bind to more than one IP address and port, but this is not (yet?) supported. Defaults to "127.0.0.1:0".
 
-|`request_body_max_bytes`
+|`default_request_body_max_bytes`
 |`4096`
 |No
 |Specifies the maximum number of bytes allowed in a request body. Larger requests will receive a 400 error. Defaults to 1024.
 
+Can be overridden per-endpoint via the `request_body_max_bytes` parameter to `#[endpoint { ... }]`.
```
ahl, line 53: future tense ;-)

(just joking; no need to change this)

dropshot/src/config.rs
```diff
@@ -43,12 +43,15 @@ pub type RawTlsConfig = rustls::ServerConfig;
 /// }
 /// ```
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
-#[serde(default)]
+#[serde(
+    from = "DeserializedConfigDropshot",
+    into = "DeserializedConfigDropshot"
```
ahl, line 48: what's the point of the into?

dropshot/src/config.rs
```diff
@@ -113,9 +116,108 @@ impl Default for ConfigDropshot {
...
     }
 }
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(default)]
+struct DeserializedConfigDropshot {
+    bind_address: SocketAddr,
+    default_request_body_max_bytes: usize,
+    // Previous name for default_request_body_max_bytes. Present only to guide
+    // users to the new name.
```
ahl, line 132: worth noting the version when this change was made? Might save us time in the future if/when we choose to EOL this migration guidance.

dropshot/src/config.rs
```diff
@@ -113,9 +116,108 @@ impl Default for ConfigDropshot {
...
             bind_address: "127.0.0.1:0".parse().unwrap(),
-            request_body_max_bytes: 1024,
+            default_request_body_max_bytes: 1024,
             default_handler_task_mode: HandlerTaskMode::Detached,
             log_headers: Default::default(),
         }
     }
 }
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
```
ahl, line 126: does this need to be Serialize?

ahl (commented):
looks good; some questions

dropshot/src/config.rs
sunshowers: Ah I didn't know `skip_serializing` was a thing! Makes sense.

dropshot/src/config.rs
sunshowers: We don't, good catch -- removed.

dropshot/src/config.rs
sunshowers: This pattern allows a custom error message to be passed in -- `InvalidConfig` is a general marker value that shouldn't necessarily be tied with the specific error message.

I'll add a comment to this note.

sunshowers: Oh also this returns an `Option<InvalidConfig>` which wouldn't be possible to do this way.

README.adoc
sunshowers: Haha, I'm actually going to move this to #1171 :)

dropshot/src/config.rs
sunshowers: It is possible to derive `Serialize` on `ConfigDropshot` directly, and it would work today, but in case there's more divergence between the in-memory implementation and the serializable implementation over time, that may no longer hold true. I've seen this be a recurring source of bugs in other places, and making the types that directly interact with serde be the same for serialization and deserialization has always been helpful.

dropshot/src/config.rs
sunshowers: This is coupled with the `into` comment.

dropshot/src/config.rs
sunshowers: Done.

dropshot/src/config.rs
ahl: > Oh also this returns an `Option<InvalidConfig>` which wouldn't be possible to do via a type implementation.

that's true... but I think `impl Serialize for InvalidConfig` would probably suffice since the implementation of Serialize for Option will call it.

sunshowers: Hmm, `InvalidConfig` doesn't implement `Serialize` at all (the `skip_serializing` makes that moot). This code only does deserialization.

sunshowers:
Going to go ahead and land this per chat message -- happy to address lingering comments in a followup.

PR #1181, opened 2024-11-20
[dropshot] improve doc organization

sunshowers:
In the review for #1171 I realized that we were specifying the endpoint
attribute params in two places, neither of which had a complete list. Put it
all in one place: on documentation for the proc macro.

Also move the documentation to `dropshot/src/lib.rs` to allow doctests to work.


ahl (approved):
nice

PR #1193, opened 2024-12-04
Add a `HeaderMap` to `HttpError`

hawkw:
There are certain response headers which the HTTP standard specifies
that servers should send along with particular error responses or status
codes. For example, when returning a 405 Method Not Allowed status, 
[§ 15.5.0 RFC9110] states that:

> The origin server MUST generate an Allow header field in a 405
> response containing a list of the target resource's currently
> supported methods.

Currently, Dropshot does not do this, so we are technically not
compliant with RFC9110.

Similarly, some of Dropshot's extractors return error responses for
which we ought to emit headers. For example, the `TypedBody` extractor
will return a 415 Unsupported Media Type status, for which [§ 15.5.16
RFC9110] recommends returning an `Accept` or `Accept-Encoding` header
as appropriate:

> If the problem was caused by an unsupported content coding, the
> Accept-Encoding response header field (Section 12.5.3) ought to be
> used  to indicate which (if any) content codings would have been
> accepted in the request.
>
> On the other hand, if the cause was an unsupported media type, the
> Accept response header field (Section 12.5.1) can be used to indicate
> which media types would have been accepted in the request.

Note that, unlike the `Allow` header for 405 responses, this is just a
suggestion --- the RFC doesn't include normative language stating that
we MUST do this. But it's nice if we do.

Errors emitted internally by Dropshot are represented by the `HttpError`
type. Currently, there is no way for this type to provide headers which
should be added to responses generated from an `HttpError`. Therefore,
this branch does so by adding a `http::HeaderMap` to `HttpError`. When
the `HttpError` is converted into a response, any headers in the
`HeaderMap` are added to the response. Adding this field is a breaking
change, so I've added it to the changelog.

Because the `http::HeaderMap` type is fairly large even when empty, the
`HeaderMap` is stored in an `Option<Box<HeaderMap>>` to reduce the size
of `HttpError`s, especially in the common case where there is no header
map to add. Unlike containers such as `Vec` and `HashMap`, an empty
`http::HeaderMap` contains a fairly large number of fields, rather than
being a single pointer to a heap allocation containing the actual data.
Thus, we store an `Option<Box<HeaderMap>>` here, so that the empty case
is just a null pointer. Not doing this results in a clippy warning
on...every function returning a `Result<_, HttpError>`, which is
unfortunate because even if we ignored that lint in Dropshot, it would
still trigger for consumers of Dropshot's APIs.

As an initial proof of concept, I've added code to dropshot's router
module to add `Allow` headers with the list of allowed methods when
returning a 405 Method Not Allowed error. There are other places where
we should be returning headers in our error responses, but this felt
like one of the more important ones, since the RFC says we MUST do it.
We can add headers to other errors separately; my priority was to make
the breaking parts of the change first.

For example, note that we now return `Allow` headers in responses to
requests with invalid methods:

```console
eliza@noctis ~/Code/dropshot $ cargo run --example basic &
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/examples/basic`
Dec 04 18:51:47.826 INFO listening, local_addr: 127.0.0.1:41751

eliza@noctis ~/Code/dropshot $ curl -v http://localhost:41751/counter -X DELETE
* Host localhost:41751 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:41751...
* connect to ::1 port 41751 from ::1 port 53992 failed: Connection refused
*   Trying 127.0.0.1:41751...
* Connected to localhost (127.0.0.1) port 41751
* using HTTP/1.x
> DELETE /counter HTTP/1.1
> Host: localhost:41751
> User-Agent: curl/8.11.0
> Accept: */*
>
* Request completely sent off
Dec 04 18:59:14.913 INFO accepted connection, remote_addr: 127.0.0.1:35084, local_addr: 127.0.0.1:41751
Dec 04 18:59:14.917 INFO request completed, error_message_external: Method Not Allowed, error_message_internal: Method Not Allowed, latency_us: 834, response_code: 405, uri: /counter, method: DELETE, req_id: e1d14d57-ca67-49cc-84b0-d1dc1e0e5793, remote_addr: 127.0.0.1:35084, local_addr: 127.0.0.1:41751
< HTTP/1.1 405 Method Not Allowed
< allow: GET
< allow: PUT
< content-type: application/json
< x-request-id: e1d14d57-ca67-49cc-84b0-d1dc1e0e5793
< content-length: 93
< date: Wed, 04 Dec 2024 18:59:14 GMT
<
{
  "request_id": "e1d14d57-ca67-49cc-84b0-d1dc1e0e5793",
  "message": "Method Not Allowed"
* Connection #0 to host localhost left intact
}%

```

[§ 15.5.0 RFC9110]: https://httpwg.org/specs/rfc9110.html#status.405
[§ 15.5.16 RFC9110]: https://httpwg.org/specs/rfc9110.html#status.415

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -88,7 +88,7 @@ error.
 * `HttpError` now contains an `Option<Box<http::HeaderMap>>` of headers to add
 to error responses constructed for the error.
 +
-Code which constructs `HttpError` literals must now initialize this field.
+Code that constructs `HttpError` literals must now initialize this field.
 
 === Other notable changes
 
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -85,6 +85,37 @@ Additionally, `ErrorStatusCode` provides an `as_client_error` method that
 returns a `ClientErrorStatusCode` if the status code is a client error, or an
 error.
 
+* `HttpError` now contains an `Option<Box<http::HeaderMap>>` of headers to add
+to error responses constructed for the error.
++
+Code which constructs `HttpError` literals must now initialize this field. For
+example, this will no longer compile:
++
+```rust
+let err = dropshot::HttpError {
+    status_code: ErrorStatusCode::INTERNAL_SERVER_ERROR,
+    error_code: None
+    internal_message: "something bad happened".to_string(),
+    external_message: "Internal Error".to_string(),
+};
+```
++
+Instead, write:
++
+```rust
+let err = dropshot::HttpError {
+    status_code: ErrorStatusCode::INTERNAL_SERVER_ERROR,
+    error_code: None
+    internal_message: "something bad happened".to_string(),
+    external_message: "Internal Error".to_string(),
+    headers: None, // <-- Add this
+};
+```
++
+Similarly, code which destructures `HttpError`s, such as in a `match` or `let`
+expression, must now either destructure the `headers` field or add `..` to
+ignore it.
+
 === Other notable changes
 
 * Endpoint handler functions may now return any error type that implements the
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -88,33 +88,7 @@ error.
 * `HttpError` now contains an `Option<Box<http::HeaderMap>>` of headers to add
 to error responses constructed for the error.
 +
-Code which constructs `HttpError` literals must now initialize this field. For
-example, this will no longer compile:
-+
-```rust
-let err = dropshot::HttpError {
-    status_code: ErrorStatusCode::INTERNAL_SERVER_ERROR,
-    error_code: None
-    internal_message: "something bad happened".to_string(),
-    external_message: "Internal Error".to_string(),
-};
-```
-+
-Instead, write:
-+
-```rust
-let err = dropshot::HttpError {
-    status_code: ErrorStatusCode::INTERNAL_SERVER_ERROR,
-    error_code: None
-    internal_message: "something bad happened".to_string(),
-    external_message: "Internal Error".to_string(),
-    headers: None, // <-- Add this
-};
-```
-+
-Similarly, code which destructures `HttpError`s, such as in a `match` or `let`
-expression, must now either destructure the `headers` field or add `..` to
-ignore it.
+Code which constructs `HttpError` literals must now initialize this field.
 
 === Other notable changes
 
--- a/dropshot/src/error.rs
+++ b/dropshot/src/error.rs
@@ -118,7 +118,7 @@ pub struct HttpError {
     pub external_message: String,
     /// Error message recorded in the log for this error
     pub internal_message: String,
-    /// Headers which should be added to error responses generated from this
+    /// Headers that will be added to responses generated from this
     /// error.
     ///
     /// The [`http::HeaderMap`] is boxed to reduce the size of `HttpError`s
@@ -399,7 +399,7 @@ impl HttpError {
     ) -> hyper::Response<crate::Body> {
         let mut builder = hyper::Response::builder();
 
-        // Set the builder's initial `HeaderMap` to the headers recommended by
+        // Set the builder's initial `HeaderMap` to the headers specified by
         // the error, if any exist.  Replacing the initial `HeaderMap` is more
         // efficient than inserting each header from `self.headers` into the
         // builder one-by-one, as we can just reuse the existing header map.
--- a/dropshot/src/error.rs
+++ b/dropshot/src/error.rs
@@ -120,6 +120,14 @@ pub struct HttpError {
     pub internal_message: String,
     /// Headers which should be added to error responses generated from this
     /// error.
+    ///
+    /// The [`http::HeaderMap`] is boxed to reduce the size of `HttpError`s
+    /// which don't contain a `HeaderMap` (the common case).  Unlike containers
+    /// such as `Vec` and `HashMap`, an empty [`http::HeaderMap`] contains a
+    /// fairly large number of fields, rather than being a pointer to a heap
+    /// allocation containing the actual data.  Thus, we store an
+    /// `Option<Box<HeaderMap>>` here, so that the empty case is just a null
+    /// pointer.
     pub headers: Option<Box<http::HeaderMap>>,
 }
 
--- a/dropshot/src/error.rs
+++ b/dropshot/src/error.rs
@@ -308,9 +308,16 @@ impl HttpError {
         }
     }
 
-    /// Mutably borrow the `http::HeaderMap`
+    /// Mutably borrow the `http::HeaderMap` associated with this error, if one
+    /// exists.
     pub fn headers_mut(&mut self) -> Option<&mut http::HeaderMap> {
         self.headers.as_deref_mut()
+    }
+
+    /// Borrow the `http::HeaderMap` associated with this error, if one
+    /// exists.
+    pub fn headers_ref(&self) -> Option<&http::HeaderMap> {
+        self.headers.as_deref()
     }
 
     /// Adds a header to the [`http::HeaderMap`] of headers to add to responses
@@ -324,7 +331,7 @@ impl HttpError {
     ///   header name and value, respectively.
     /// - [`Err`]`(`[`http::Error`]`)` if the header name or value is invalid,
     ///   or the `HeaderMap` is full.
-    pub fn set_header<K, V>(
+    pub fn add_header<K, V>(
         &mut self,
         name: K,
         value: V,
--- a/dropshot/src/error.rs
+++ b/dropshot/src/error.rs
@@ -305,16 +305,10 @@ impl HttpError {
         }
     }
 
-    /// Mutably borrow the `http::HeaderMap` associated with this error, if one
-    /// exists.
-    pub fn headers_mut(&mut self) -> Option<&mut http::HeaderMap> {
-        self.headers.as_deref_mut()
-    }
-
-    /// Borrow the `http::HeaderMap` associated with this error, if one
-    /// exists.
-    pub fn headers_ref(&self) -> Option<&http::HeaderMap> {
-        self.headers.as_deref()
+    /// Mutably borrow the `http::HeaderMap` associated with this error. If
+    /// there is no header map for this error, this method creates one.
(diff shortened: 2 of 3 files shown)


dropshot/src/error.rs
```diff
@@ -118,6 +118,17 @@ pub struct HttpError {
...
+    /// error.
+    ///
+    /// The [`http::HeaderMap`] is boxed to reduce the size of `HttpError`s
+    /// which don't contain a `HeaderMap` (the common case).  Unlike containers
+    /// such as `Vec` and `HashMap`, an empty [`http::HeaderMap`] contains a
+    /// fairly large number of fields, rather than being a pointer to a heap
+    /// allocation containing the actual data.  Thus, we store an
+    /// `Option<Box<HeaderMap>>` here, so that the empty case is just a null
+    /// pointer.
+    pub headers: Option<Box<http::HeaderMap>>,
```
hawkw, line 131: Boxing this resolves a Clippy lint about the error type being too large. I figured we shouldn't just disable that lint, since it would presumably also trigger on user code that returns an `HttpError`, and it seems sad to have an API that results in lints in consumers if they don't go out of their way to disable them.

I also considered making the field private, and requiring it be accessed via methods, since the `Option<Box<...>>` makes it a bit awkward. However, that would prevent consumers of the API from manually constructing `HttpError`s by initializing the other fields, which I've seen a lot of in Omicron, so I figured it was better to just make it public.

dropshot/src/error.rs
```diff
@@ -290,15 +304,118 @@ impl HttpError {
...
+    /// generated from this error, taking the error by value.
+    ///
+    /// If this error does not already have a header map (`self.header_map` is
+    /// `None`), this method creates one.
+    ///
+    /// Unlike [`HttpError::set_header`], this method takes `self` by value,
+    /// allowing it to be chained to form an expression that returns an
+    /// `HttpError`. However, because this takes `self` by value, returning an
+    /// error for an invalid header name or value will discard the `HttpError`.
+    /// To avoid this, use [`HttpError::set_header`] instead.
```
hawkw, line 365: This is a bit awkward. Another option could be to make the method take `HeaderName` and `HeaderValue` only, instead of types that can be converted into them; then, it could be infallible, instead of potentially eating the `HttpError`.

I did it this way because it's similar to the `http::Response::builder()` API, and allows users to just pass strings as the header name/value, rather than having to convert them in user code. But, returning a `Result` and maybe throwing away the error is also unfortunate. I could be convinced to change this --- what do y'all think?

CHANGELOG.adoc
```diff
@@ -85,6 +85,37 @@ Additionally, `ErrorStatusCode` provides an `as_client_error` method that
...
++
+```rust
+let err = dropshot::HttpError {
+    status_code: ErrorStatusCode::INTERNAL_SERVER_ERROR,
+    error_code: None
+    internal_message: "something bad happened".to_string(),
+    external_message: "Internal Error".to_string(),
+    headers: None, // <-- Add this
+};
+```
```
ahl, line 113:
```suggestion
Code that constructs `HttpError` literals must now initialize this field.
```

CHANGELOG.adoc
```diff
@@ -85,6 +85,37 @@ Additionally, `ErrorStatusCode` provides an `as_client_error` method that
...
++
+Similarly, code which destructures `HttpError`s, such as in a `match` or `let`
+expression, must now either destructure the `headers` field or add `..` to
+ignore it.
```
ahl, line 117: del

dropshot/src/error.rs
```diff
@@ -118,6 +118,17 @@ pub struct HttpError {
     pub external_message: String,
     /// Error message recorded in the log for this error
     pub internal_message: String,
+    /// Headers which should be added to error responses generated from this
```
ahl, line 121:
```suggestion
    /// Headers that will be added to responses generated from this
```

dropshot/src/error.rs
```diff
@@ -118,6 +118,17 @@ pub struct HttpError {
...
+    /// error.
+    ///
+    /// The [`http::HeaderMap`] is boxed to reduce the size of `HttpError`s
+    /// which don't contain a `HeaderMap` (the common case).  Unlike containers
+    /// such as `Vec` and `HashMap`, an empty [`http::HeaderMap`] contains a
+    /// fairly large number of fields, rather than being a pointer to a heap
+    /// allocation containing the actual data.  Thus, we store an
+    /// `Option<Box<HeaderMap>>` here, so that the empty case is just a null
+    /// pointer.
```
ahl, line 130: I would suggest replacing all of this with a `// Boxed in obeisance to clippy.`

dropshot/src/error.rs
```diff
@@ -290,15 +304,118 @@ impl HttpError {
...
+    /// exists.
+    pub fn headers_mut(&mut self) -> Option<&mut http::HeaderMap> {
+        self.headers.as_deref_mut()
+    }
+
+    /// Borrow the `http::HeaderMap` associated with this error, if one
+    /// exists.
+    pub fn headers_ref(&self) -> Option<&http::HeaderMap> {
+        self.headers.as_deref()
+    }
```
ahl, line 321: is this needed? I see that we don't have it for `HttpResponseHeaders`

dropshot/src/error.rs
```diff
@@ -290,15 +304,118 @@ impl HttpError {
...
+
     /// Generates an HTTP response for the given `HttpError`, using `request_id`
     /// for the response's request id.
     pub fn into_response(
         self,
         request_id: &str,
     ) -> hyper::Response<crate::Body> {
+        let mut builder = hyper::Response::builder();
+
+        // Set the builder's initial `HeaderMap` to the headers recommended by
```
ahl, line 402:
```suggestion
        // Set the builder's initial `HeaderMap` to the headers specified by
```

dropshot/src/router.rs
```diff
@@ -532,10 +532,21 @@ impl<Context: ServerContext> HttpRouter<Context> {
...
             find_handler_matching_version(handlers, version).is_some()
         }) {
-            Err(HttpError::for_client_error_with_status(
+            let mut err = HttpError::for_client_error_with_status(
                 None,
                 ClientErrorStatusCode::METHOD_NOT_ALLOWED,
-            ))
+            );
+
+            // Add `Allow` headers for the methods that *are* acceptable for
```
ahl, line 540: cite the RFC?

dropshot/src/error.rs
hawkw: That's fair. I felt that including the justification for this in the API docs (rather than in an implementation comment) was worthwhile, since this _is_ exposed to consumers of the API, and they might wonder why this isn't just a `HeaderMap` when constructing/destructuring `HttpError`s. But, I'm happy to be convinced otherwise!

CHANGELOG.adoc
hawkw: Hm, @davepacheco had previously expressed a preference for including examples of breaking changes whenever possible...though I agree that this one is pretty trivial. Happy to remove it.

CHANGELOG.adoc
hawkw: IMO, this is worth mentioning, because otherwise, the language in the previous sentence suggests that this change _only_ applies to code that constructs a `HttpError` literal. We could simplify it to something like:

> Code which constructs a `HttpError` literal or destructures `HttpError`s in a pattern must be changed to add this field.

What do you think?

dropshot/src/router.rs
hawkw: good call, done in https://github.com/oxidecomputer/dropshot/pull/1193/commits/82d4f91c05cfa111952f08099011684416d72939!

dropshot/src/error.rs
hawkw: It's not strictly necessary, but it felt like due to the `Option<Box<...>>` situation, it was a bit nicer than making the user figure out `as_deref()` on their own. I'm fine with removing it.

ahl: In general: I think less is more. Perhaps at least we should have these two mechanisms (i.e. HttpResponseHeaders) work similarly.

dropshot/src/error.rs
ahl: I guess my perspective is that for the 99% of folks who don't care, it's a distraction. For the 1% who do, it's self-evident.

CHANGELOG.adoc
ahl: All this feels like the expected consequences of adding a new field to a structure. I don't expect folks will proactively fix things based on the changelog and I also don't think folks will go looking at the changelog to figure out how to address these compilation errors.

dropshot/src/error.rs
hawkw: Fair enough!

dropshot/src/error.rs
hawkw: Yeah, that's fair. I've removed `headers_ref()` in d69f141d3dbd210cfb91a92f5a8cf9a4e55562a1.

I've also changed `headers_mut()` to return `&mut HeaderMap`, creating one if it's `None`, so that it can have the same signature as `HttpResponseHeaders::headers_mut()`. I figured that most of the time, if you're looking to mutate a headermap, it's because you want there to be headers.  Users who care about the `Option` can just touch the `headers` field directly.

CHANGELOG.adoc
hawkw: Removed it in a34b66c1a2c2293a23cfa4ac3f3bd54effffe053.

CHANGELOG.adoc
```diff
@@ -88,33 +88,7 @@ error.
...
-    internal_message: "something bad happened".to_string(),
-    external_message: "Internal Error".to_string(),
-    headers: None, // <-- Add this
-};
-```
-+
-Similarly, code which destructures `HttpError`s, such as in a `match` or `let`
-expression, must now either destructure the `headers` field or add `..` to
-ignore it.
+Code which constructs `HttpError` literals must now initialize this field.
```
ahl, line 91:
```suggestion
Code that constructs `HttpError` literals must now initialize this field.
```

ahl (approved):

PR #1194, opened 2024-12-06
handler.rs: fix typo

nshalman:
Looks like a copy-paste that wasn't then edited afterwards. Noticed it while messing around with

https://github.com/nshalman/dropshot/commit/34c441b8e394873af0a2b33217405e637b5a84ae

This is just a bug report with a patch. Feel free to re-implement in a different commit elsewhere, etc.


davepacheco (commented):
Thanks!

I think this is right but want to double-check with @hawkw because I remember there being some case(s) where the external message _is_ the internal message.

hawkw (approved):
whoops!

PR #1201, opened 2024-12-12
[DRAFT - DO NOT MERGE] Basic OpenTelemetry integration

nshalman:
# OpenTelemetry for Dropshot

This is now version 2. It has leaned into the use of `tokio-tracing` and `opentelemetry-tracing`.
The `tracing` feature funnels tracing logs out to `slog` for unified logging.
The `otel-tracing` feature enables shipping traces via OTLP.
They can be used together or individually.

Checklist of things that are needed (note that much of it is currently incomplete):
- [x] Documentation
- [ ] Tests
- [ ] All desired fields properly traced
- [x] Examples including how to do propagation

# Known Issues

~~- If a handler panics, currently we get a 499 error span. I don't really like that.~~ See #1359 

# dtrace.conf(24) References

Slides: 
https://docs.google.com/presentation/d/1AW_ugVrmkXDTROgPbEt7bCLrLhFo8-z13PoJAdfPO78/

Raw video link:
https://www.youtube.com/watch?v=KQN1t2qlhaw&t=11760s

# Screenshot

image

### DIFF
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -287,6 +287,7 @@ where
 /// This type is not exported to Dropshot consumers; it is purely an internal
 /// implementation detail of the interface between `HttpHandlerFunc` and the
 /// server.
+#[derive(Debug)]
 pub enum HandlerError {
     /// An error returned by a fallible handler function itself.
     ///
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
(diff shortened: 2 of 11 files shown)


dropshot/src/handler.rs
```diff
@@ -86,6 +86,8 @@ pub struct RequestContext<Context: ServerContext> {
     pub log: Logger,
     /// basic request information (method, URI, etc.)
     pub request: RequestInfo,
+    #[cfg(feature = "otel-tracing")]
+    pub span_context: opentelemetry::trace::SpanContext,
```
nshalman, line 90: This is gross and I need something better.

dropshot/src/handler.rs
```diff
@@ -377,6 +379,19 @@ where
...
+            f,
+            "{}",
+            match self {
+                Self::Handler { ref message, .. } => message,
+                Self::Dropshot(ref e) => &e.external_message,
+            }
+        )
+    }
+}
+
```
nshalman, line 394: I think this is for tokio-tracing and could potentially be dropped or separated into a dedicated commit. Really all the tokio-tracing bits should be a separate commit now that I think about it.

nshalman:
As I've been struggling to make this cleaner I think I have had a realization. I want to be able to trace futures which means I need to use `tracing-futures`  which means that I have an explicit opinion on the https://github.com/open-telemetry/opentelemetry-rust/issues/1571 debate. I want to just use `tracing` and `tracing-opentelemetry`  rather than doing quite so much low-level otel stuff.
This PR needs a refactor to reflect that, but that should also clean it up tremendously.

ahl:
We can re-open if work resumes

PR #1208, opened 2024-12-21
Fix argument inconsistency in example documentation

bahamat:
The "bare minimum" example has (among other things) this:

```
    let server = ServerBuilder::new(api, Arc::new(()), log)
        .start()
        .map_err(|error| format!("failed to start server: {}", error))?;
```

Then later the examples for registering APIs (both as a free function and as an API trait) don't wrap the argument in `Arc<>`.

This leads new users into a trap where these two (three) disconnected examples *seem* like they should work, but lead to a compile error once they add a handler.

Discussing with @davepacheco, it seems the consensus is that `Arc` is not particularly recommended here anymore.

This change will help new users walk right into a working example, rather than encountering an error.

### DIFF
--- a/dropshot/src/lib.rs
+++ b/dropshot/src/lib.rs
@@ -1,4 +1,5 @@
 // Copyright 2024 Oxide Computer Company
+// Copyright 2024 Brianna Bennett
 
 //! Dropshot is a general-purpose crate for exposing REST APIs from a Rust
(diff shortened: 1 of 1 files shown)


ahl:
I noticed you added your copyright; this seems like what would qualify as a *de minimis* change; if you feel otherwise let us know. Thanks!

PR #1246, opened 2025-01-31
versioning specifiers should allow identifiers

davepacheco:
The motivation here is that if you're working on a project where, say:

- the latest version of your API on "main" is 1.0.0
- you branch from "main" to create 1.1.0
- somebody else lands a change on "main" that adds 1.1.0 (or any higher version)
- when you merge with "main", you need to change your new version number to be 1.2.0 (or something higher than whatever's there)

It would be painful and error-prone if you had to go find all the places where you added "1.1.0" to an `endpoint` macro and change it to "1.2.0".  It wouldn't even work to find/replace 1.1.0 -> 1.2.0 since that will catch both the ones you added and the ones that were added to "main".

The idea here is that the project instead defines symbolic constants, so it would look like this:

- the latest version of your API on "main" is VERSION_INITIAL, which happens to be `1.0.0`
- you branch from "main" to create `const VERSION_MY_FEATURE: semver::Version`, which is 1.1.0
- somebody else lands a change on "main" that adds `VERSION_OTHER_FEATURE`, which is 1.1.0 (or any higher version)
- when you merge with "main", you need to change `VERSION_MY_FEATURE` to 1.2.0, but you don't need to change this value anywhere else (i.e., in the endpoint macro usage)

---

The change to `dropshot/src/lib.rs` (docs) explains the updated syntax and the change to `dropshot/examples/versioning.rs` shows it in action.


dropshot_endpoint/src/metadata.rs
```diff
@@ -314,9 +328,72 @@ impl ValidatedEndpointMetadata {
...
+
+impl syn::parse::Parse for VersionSpecifier {
+    fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
+        let lookahead = input.lookahead1();
+        if lookahead.peek(syn::LitStr) {
+            let s = input.parse::<syn::LitStr>()?;
+            let v = parse_semver(&s)?;
+            Ok(VersionSpecifier::Literal(v))
+        } else {
+            Ok(VersionSpecifier::Identifier(input.parse::<syn::Path>()?))
```
ahl, line 394: Good call to have a Path rather than an Ident. In the future it would seem reasonable to have an Expr, but agreed that there's no need now

dropshot_endpoint/src/metadata.rs
```diff
@@ -314,9 +328,72 @@ impl ValidatedEndpointMetadata {
...
+                Ok(VersionRange::All)
+            } else {
+                let latest = input.parse::<VersionSpecifier>()?;
+                Ok(VersionRange::Until(latest))
+            }
+        } else {
+            let earliest = input.parse::<VersionSpecifier>()?;
+            let dotdot = input.parse::<syn::Token![..]>()?;
+            let lookahead = input.lookahead1();
+            if lookahead.peek(syn::LitStr) || lookahead.peek(syn::Ident) {
```
ahl, line 351: Do you need to peek or can you if let Ok(..) parse?

ahl (approved):

dropshot_endpoint/src/metadata.rs
davepacheco: Good question!  I almost wrote that but I think you could ask the same thing about the existing code.  I inferred maybe a failed `parse()` might eat some tokens or otherwise leave the stream in the wrong state so I just went and did the lookahead here, even though this involves knowing what the `VersionSpecifier` accepts.  (We're also using `syn::Ident` instead of `syn::Path` because for reasons I don't understand `syn::Ident` has no data but `syn::Path` does so I wasn't sure how to construct that.)  Sorry this all feels janky but the jank at least feels contained (and correct).

PR #1269, opened 2025-02-27
server: Log HTTP response code as u16

wfchandler:
Currently HTTP response codes are generally logged as strings, which is a bit unintuitive as these are integers.

Log the response code as a numeric value.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -15,6 +15,8 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.16.0\...HEAD[Full list of commits]
 
+* https://github.com/oxidecomputer/dropshot/pull/1269[#1269] HTTP response codes are now logged as numeric values, rather than strings.
+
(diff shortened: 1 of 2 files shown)


davepacheco:
Thanks for this.  Interestingly, I see that of the four cases you changed:

- three were using `as_str()`, which is just a string value of the same thing so that seems like no downside to changing as you've done here
- one was using the `%` syntax, which uses `Display`, which would produce not, say, `"404"` but `"404 Not Found"`:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3cd99f30c5c70b39aa3c562db6a9c742

It's kind of nice to get the reason string in this last case but we mostly weren't getting that anyway (this was kind of an edge case) so this seems like no reason not to make this change.

Could you update the changelog to mention this?  Then we can land this.

wfchandler:
Thanks @davepacheco, I've updated the changelog.

davepacheco (approved):

PR #1325, opened 2025-05-15
Add a Header extractor for header parameters

ahl:
I think `from_map` could be updated to be more elegant (it's basically the first Rust I ever wrote), but I think this will be fine in the meantime (and probably much longer).

### DIFF
--- a/dropshot/src/extractor/header.rs
+++ b/dropshot/src/extractor/header.rs
@@ -1,0 +1,67 @@
+// Copyright 2025 Oxide Computer Company
+
+//! Header extractor
+
+use std::collections::BTreeMap;
+
+use async_trait::async_trait;
+use schemars::JsonSchema;
+use serde::de::DeserializeOwned;
+
+use crate::{
+    from_map::from_map, ApiEndpointBodyContentType,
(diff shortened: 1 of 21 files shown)


dropshot/src/extractor/header.rs
```diff
@@ -0,0 +1,67 @@
...
+
+use crate::{
+    from_map::from_map, ApiEndpointBodyContentType,
+    ApiEndpointParameterLocation, HttpError, RequestContext, RequestInfo,
+    ServerContext,
+};
+
+use super::{metadata::get_metadata, ExtractorMetadata, SharedExtractor};
+
+pub struct Header<HeaderType: DeserializeOwned + JsonSchema + Send + Sync> {
```
smklein, line 19: This is part of the public-facing API; it might be worth documenting (or including a small example)

smklein (approved):
Makes sense to me, and the example is readable! Happy to use this.

dropshot/src/extractor/header.rs
ahl: Good point; thanks

PR #1328, opened 2025-05-16
Header parameters may not be null/nullable

ahl:


david-crespo (approved):
seems good to me

ahl:
Want to see if this works for @smklein before merging (and cutting *another* release)

smklein:
I updated Cargo.toml to use this branch in my PR (https://github.com/oxidecomputer/omicron/pull/8183), then regenerated openapi specs.

This resulted in the following diff:

```
diff --git a/openapi/sled-agent.json b/openapi/sled-agent.json
index 2798c2846..1156332a1 100644
--- a/openapi/sled-agent.json
+++ b/openapi/sled-agent.json
@@ -1128,7 +1128,6 @@
             "name": "range",
             "description": "A request to access a portion of the resource, such as:\n\n``` bytes=0-499 ```\n\nhttps://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Range",
             "schema": {
-              "nullable": true,
               "type": "string"
             }
           },
```

(I'm sorta assuming that `"required": false` is the default here)

Anyway, this compiles, I can built the sled agent client, and I'm seeing an `Option<&str>` as a new argument, for this header.

TL;DR: So far, appears working-as-intended

ahl:
> (I'm sorta assuming that `"required": false` is the default here)

Correct per the [OpenAPI docs](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields-10) for `required`:
> Determines whether this parameter is mandatory. If the [parameter location](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#parameterIn) is "path", this property is REQUIRED and its value MUST be true. Otherwise, the property MAY be included and its default value is false.

smklein (approved):
This PR is working for me in https://github.com/oxidecomputer/omicron/pull/8183 ; I'm depending on it there to make range requests work

PR #1329, opened 2025-05-21
support optional values in the HttpResponseHeaders inner type

### LINKED ISSUE #1312: Allow Option type in HttpResponseHeaders
The `to_map` function in dropshot does not support Option types right now.

This means a type like this is invalid:

```
pub struct GlobalHeaders {
    #[serde(skip_serializing_if = "Option::is_none", rename = "Access-Control-Allow-Origin")]
    pub access_control_allow_origin: Option<String>,

    /// Access-Control-Allow-Credentials header.
    #[serde(skip_serializing_if = "Option::is_none", rename = "Access-Control-Allow-Credentials")]
    pub access_control_allow_credentials: Option<String>,

    #[serde(skip_serializing_if = "Option::is_none", rename = "Access-Control-Allow-Methods")]
    pub access_control_allow_methods: Option<String>,

    #[serde(skip_serializing_if = "Option::is_none", rename = "Access-Control-Allow-Headers")]
    pub access_control_allow_headers: Option<String>,
}

This test fails:
```
    #[test]
    fn test_global_headers_serialization() {
        let headers = super::GlobalHeaders {
            access_control_allow_origin: Some("http://example.com".to_string()),
            access_control_allow_credentials: Some("true".to_string()),
            access_control_allow_methods: Some(super::ACCESS_CONTROL_ALLOWED_METHODS.to_string()),
            access_control_allow_headers: Some(super::ACCESS_CONTROL_ALLOWED_HEADERS.to_string()),
        };

        HttpResponseHeaders::new(HttpResponseOk(""), headers).to_result().expect("headers should serialize");
    }
```

### PR DESCRIPTION
**Pull request by ahl**
FIxes #1312

### DIFF
--- a/dropshot/src/to_map.rs
+++ b/dropshot/src/to_map.rs
@@ -189,10 +189,8 @@ impl SerializeStruct for MapSerializeStruct {
     where
         T: Serialize + ?Sized,
     {
-        let mut serializer = StringSerializer;
-        let value = value.serialize(&mut serializer)?;
-        self.output.insert(key.to_string(), value);
-        Ok(())
+        let mut serializer = ValueSerializer { key, output: &mut self.output };
+        value.serialize(&mut serializer)
     }
 
     fn end(self) -> Result<Self::Ok, Self::Error> {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: dropshot/src/to_map.rs [hunk 1]
```diff
@@ -203,9 +201,12 @@ impl SerializeStruct for MapSerializeStruct {
 /// A trivial `Serializer` used to extract a `String`. One could imagine
```
aapoalas, line 201: **nitpick:** The comment seems to not necessarily match the implementation anymore.

aapoalas (approved):
Comment nitpicking aside LGTM, for what it is worth.

dropshot/src/to_map.rs
ahl: I think it still does: we don't permit non-string values.

PR #1346, opened 2025-06-11
Add Serialize/Deserialize for ErrorStatusCode

hawkw:
PR #1180 added `ErrorStatusCode` and `ClientErrorStatusCode` types that represent HTTP status codes that are validated to be in the 400-599 and 400-499 ranges, respectively. These are itnended to be returned by user-defined error types in their `HttpResponseError` implementations.

User-defined errors are typically implemented as a type that implements `Serialize` and `JsonSchema` along with `HttpResponseError`, which in turn requires that the type have a `From<dropshot::HttpError>` conversion, used in the case of extractor and dropshot-internal errors. In order for the user-defined error's `HttpResponseError` implementation to have the same status as the `dropshot::HttpError` from which it was constructed, it must contain that status code within the error type so that it may return it. However, `ErrorStatusCode` does not implement `Serialize`, `Deserialize`, or `JsonSchema`, so holding it in the user-defined error type breaks deriving those traits.

One solution is to use `#[serde(skip)]` for the `ErrorStatusCode` field, which I showed in the [`custom-error.rs` example][1]. When the user-defined error type is only used in the server, this is sufficient, and it will simply omit the status when serializing. However, this prevents the user-defined error type from implementing `Deserialize`, since the status code is never serialized. This means that generated client code cannot use the same type for the error response value as the server code, which is desirable in some cases (e.g. to use the same `fmt::Display` implementation on both sides, etc). Also, in some cases, it may be useful to include a status code in the serialized body, such as when embedding an HTTP error returned by an external service which may not actually be the same as the response's status code (e.g. I might return a 500 or a 503 when a request to an upstream service returns a 4xx error).

Therefore, this commit adds `Serialize`, `Deserialize`, and `JsonSchema` implementations for the `ErrorStatusCode` and `ClientErrorStatusCode` types. These implementations serialize and deserialize these types as a `u16`, and we generate a JSON Schema that represents them as integers with appropriate minimum and maximum value validation.

[1]:
    https://github.com/oxidecomputer/dropshot/blob/c028c6751771d89457c44286e26b465ed14ef25f/dropshot/examples/custom-error.rs#L63-L69

### DIFF
--- a/dropshot/src/error_status_code.rs
+++ b/dropshot/src/error_status_code.rs
@@ -4,6 +4,7 @@
 //! Newtypes around [`http::StatusCode`] that are limited to status ranges
 //! representing errors.
 
+use serde::{Deserialize, Serialize};
 use std::fmt;
 
 /// An HTTP 4xx (client error) or 5xx (server error) status code.
@@ -20,7 +21,10 @@ use std::fmt;
 /// fallible conversion from an [`http::StatusCode`].
 ///
 /// [iana]: https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
-#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(
+    Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize,
+)]
+#[serde(try_from = "u16", into = "u16")]
 pub struct ErrorStatusCode(http::StatusCode);
 
 // Generate constants for a `http::StatusCode` wrapper type that re-export the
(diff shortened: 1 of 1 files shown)


ahl (approved):
Looks good. What do you see as the tradeoff with the custom json schemas? What you've done is definitely more precise, but I'm not sure if that's also going to make it harder for e.g. generators to work with. I could see it either way, but just wanted to offer that schema precision isn't always the most convenient (i.e. if you had not considered "just" making it an integer). Thanks for doing this!

ahl:
I think this seemed like a clear win and then less clearly a win. I'm closing this without bias; and we may revisit in the future.

PR #1352, opened 2025-06-16
repro for x-rust-type param not being present

sunshowers:


sunshowers:
Incorporated in #1356.

PR #1356, opened 2025-06-21
various improvements to JSON schema processing

ahl:
@sunshowers noticed that we aren't including schemas in OpenAPI output that are only referenced by our custom `x-rust-type` extension. We can address this by massively simplifying how we process schemas. This is some very very old code in dropshot, and I'm not quite sure why it felt like the right approach, but we had been generating them schema, and then scraping out references using a visitor.... but the references were already present in the structures used to generate the schema. It's a bit perplexing why we took this extra step--perhaps there's some other shoe that will drop, but we've certainly developed many more tests since then, and none of them point to any reason why this excursion might have been necessary.

We also noticed that `schemars` allows for modeling a empty `enum` value e.g. `{ "enum": [] }` and `openapiv3` does not; `schemars` will also emit such a schema when the `JsonSchema` derive macro is applied to a never-type enum (e.g. `enum Foo {}). To handle this while preserving extensions we emit the following schema which **is** representable by `openapiv3` structures: `{ "not": {} }`.

And for funsies, we also handle the `false` schema using the same construction.

### DIFF
--- a/dropshot/src/schema_util.rs
+++ b/dropshot/src/schema_util.rs
@@ -255,50 +255,6 @@ pub(crate) fn schema_extensions(
     }
 }
 
-/// Used to visit all schemas and collect all dependencies.
-pub(crate) struct ReferenceVisitor<'a> {
-    generator: &'a schemars::gen::SchemaGenerator,
-    dependencies: indexmap::IndexMap<String, schemars::schema::Schema>,
-}
-
-impl<'a> ReferenceVisitor<'a> {
-    pub fn new(generator: &'a schemars::gen::SchemaGenerator) -> Self {
-        Self { generator, dependencies: indexmap::IndexMap::new() }
-    }
-
-    pub fn dependencies(
-        self,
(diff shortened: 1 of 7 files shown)


dropshot/src/schema_util.rs
```diff
@@ -379,14 +335,26 @@ pub(crate) fn j2oas_schema(
         // when consumers use a type such as serde_json::Value.
         schemars::schema::Schema::Bool(true) => {
             openapiv3::ReferenceOr::Item(openapiv3::Schema {
-                schema_data: openapiv3::SchemaData::default(),
-                schema_kind: openapiv3::SchemaKind::Any(
-                    openapiv3::AnySchema::default(),
-                ),
+                schema_data: Default::default(),
+                schema_kind: openapiv3::SchemaKind::Any(Default::default()),
```
sunshowers, line 339: tiny nit: worth extracting the maximally permissive schema into a function? we use it three times

sunshowers (approved):
Thank you!

PR #1359, opened 2025-06-25
Fix panic handling in detached mode to return 500 instead of 499

nshalman:
Handler panics in HandlerTaskMode::Detached were incorrectly being reported as 499 (Client Closed Request) instead of 500 (Internal Server Error) in telemetry and traces.

This occurred because the panic handling code called panic::resume_unwind(), which caused the request handling task to panic and trigger the disconnect guard, leading to incorrect status code assignment.

Changes:
- Extract panic message from JoinError and convert to HttpError
- Return 500 Internal Server Error instead of propagating panic
- Add integration test to verify panic handling returns proper 500 status
- Preserve existing behavior for HandlerTaskMode::CancelOnDisconnect

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -15,6 +15,8 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.17.1\...HEAD[Full list of commits]
 
+* A handler panic that unwinds through the request task is now reported as a panic ("request handling panicked", with no status code) instead of being misreported as a client disconnection with the non-standard 499 status code.  Panic propagation itself is unchanged.
+
 == 0.17.1 (released 2026-06-02)
 
 https://github.com/oxidecomputer/dropshot/compare/v0.17.0\...v0.17.1[Full list of commits]
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -16,6 +16,7 @@
 https://github.com/oxidecomputer/dropshot/compare/v0.17.1\...HEAD[Full list of commits]
 
 * A handler panic that unwinds through the request task is now reported as a panic ("request handling panicked", with no status code) instead of being misreported as a client disconnection with the non-standard 499 status code.  Panic propagation itself is unchanged.
+* Added `ServerBuilder::catch_handler_panics()`, an opt-in for catching panics in endpoint handler functions (and only there) and converting them into 500 responses.  The default remains to let panics propagate.
 
 == 0.17.1 (released 2026-06-02)
 
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -377,6 +377,19 @@ where
     }
 }
 
+/// Extracts a human-readable message from a handler panic payload, for
+/// logging and for the internal message of a 500 response when running with
+/// [`crate::ServerBuilder::catch_handler_panics()`].
+fn panic_message(panic: Box<dyn std::any::Any + Send>) -> String {
+    match panic.downcast::<&'static str>() {
+        Ok(s) => s.to_string(),
+        Err(panic) => match panic.downcast::<String>() {
+            Ok(s) => *s,
+            Err(_) => "handler panicked".to_string(),
+        },
+    }
+}
+
 /// An error type that can be converted into an HTTP response.
 ///
 /// The error types returned by handlers must implement this trait, so that a
@@ -661,7 +674,41 @@ macro_rules! impl_HttpHandlerFunc_for_func_with_params {
             _param_tuple: ($($T,)*),
         ) -> Result<Response<Body>, HandlerError>
         {
-            let response: ResponseType = (self)(rqctx, $(_param_tuple.$i,)*).await?;
+            // Read these before `rqctx` is moved into the handler's future.
+            let catch_panics = rqctx.server.config.catch_handler_panics;
+            let panic_log = catch_panics.then(|| rqctx.log.clone());
+            let handler_future = (self)(rqctx, $(_param_tuple.$i,)*);
+            let result = if catch_panics {
+                // This `catch_unwind` wraps exactly the consumer's endpoint
+                // handler: extraction has already happened and the response
+                // conversion happens below, so a caught panic is known to
+                // have come from consumer code, not Dropshot's.  (This also
+                // means the response could not have begun to be sent yet.)
+                // The future is no less unwind-safe for being polled inside
+                // `catch_unwind`: the panic could otherwise only escape this
+                // frame by unwinding through the caller and executor.
+                match futures::FutureExt::catch_unwind(
+                    std::panic::AssertUnwindSafe(handler_future),
+                )
+                .await
+                {
+                    Ok(result) => result,
+                    Err(panic) => {
+                        let message = panic_message(panic);
+                        slog::error!(
+                            panic_log.unwrap(),
+                            "handler panicked; returning 500 error";
+                            "panic_message" => &message,
+                        );
+                        return Err(HandlerError::from(
+                            HttpError::for_internal_error(message),
+                        ));
+                    }
+                }
+            } else {
+                handler_future.await
+            };
+            let response: ResponseType = result?;
             response.to_result().map_err(|error| {
                 // If turning the endpoint's response into a
                 // `http::Response<Body>` failed, try to convert the `HttpError` into
--- a/dropshot/src/server.rs
+++ b/dropshot/src/server.rs
@@ -805,6 +805,18 @@ async fn http_request_handle_wrap<C: ServerContext>(
     let on_disconnect = guard((), |_| {
         let latency_us = start_time.elapsed().as_micros();
 
+        // Besides cancellation, this guard also runs if a panic unwinds out
+        // of the request handler (see `http_request_handle`), dropping the
+        // request future.  That's not a client disconnection, so report the
+        // panic as such -- and report no status code at all, much as a
+        // process terminated by a signal has no exit code.
+        if std::thread::panicking() {
+            error!(request_log, "request handling panicked";
+                "latency_us" => latency_us,
+            );
+            return;
+        }
+
         warn!(request_log, "request handling cancelled (client disconnected)";
             "latency_us" => latency_us,
         );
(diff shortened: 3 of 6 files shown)


nshalman:
> I don't love the idea of treating a panic as a handleable error, though. In the "programmer error" / "operational error" parlance, our general disposition among most of the Dropshot consumers I'm familiar with is to treat panics as programmer errors that reflect something deeply wrong, whose blast radius could include the whole process. I know this isn't quite fair, but I think of a panic handler in our programs as analogous to a SIGSEGV handler in a C program that reports the error and lets the program continue. That'd be pretty dangerous -- often _much_ worse than crashing. Folks have made the case that Rust provides enough safety that the blast radius of a panic might well be more contained than "the whole process", and I could see that. But in practice, I think it's been quite rare for us to see panics in request handlers, which both validates the idea that these really shouldn't happen but also means we don't have a lot of experience to say that the blast radius is likely to be pretty contained.

Agreed! So the remaining question on my mind then is, why, when a handler panics in detached mode does dropshot not find a way to panic the entire server? It seems to me that either if a panic does not crash the server, then the user deserves a 500 error, or, otherwise the entire process should abort and dump core. Currently it seems like dropshot does neither. But maybe I am not using it quite right?

davepacheco:
Your question has caused me to re-think a lot of this from first principles (probably too far).  Sorry in advance this is a little rambly.

> Agreed! So the remaining question on my mind then is, why, when a handler panics in detached mode does dropshot not find a way to panic the entire server? It seems to me that either if a panic does not crash the server, then the user deserves a 500 error, or, otherwise the entire process should abort and dump core. Currently it seems like dropshot does neither. But maybe I am not using it quite right?

Dropshot doesn't do this because it can't know what the consumer wants.  I assume it's the consumer's responsibility to set things up so that panics are handled the way that they want, especially if their goal is to abort on panic since that's easy to do globally.  For our own stuff, we set `panic = abort` in release builds (which, incidentally, I think means that it doesn't matter what Dropshot does -- the program will call `abort(3C)` directly from the panic subsystem in Rust).

I believe Dropshot propagates this particular panic because it seemed like the thing that would let consumers decide how they wanted to handle it.  In retrospect, that approach works fine for ordinary synchronous dependencies, but it's a lot less clear what should be done in dependencies like hyper or Dropshot that spawn background threads or tasks that _themselves_ could panic.  These crates should probably translate that panic into an error that they can propagate to the consumer (and, ideally, force the consumer to have a handler for such errors so that they can't go silently unnoticed -- my big fear about this sort of thing is that some critical background thread/task simply stops running with little more than a note in the log and the program limps along in a permanent state of implicit, non-fatal failure).  There's probably a gap in Dropshot here.

But even if we had that, we'd still want to clean up the state of the request, which means figuring out what to report for the result and how to respond to the request.  499 is clearly wrong since this is not any kind of client error.  I still don't love 500 for a few reasons:

- it implies a clean failure with no consequences for other Dropshot requests, and we can't really know that if we panicked inside Dropshot
- it implies that the client got the 500, or at least that it didn't get something _else_.  But it's possible we sent even a 200 header to the client and panicked after that.  It'd be _really_ confusing to see a 500 on the server for the same request that the client saw some other code for.

It's always possible that if Dropshot panics at the wrong time, _nothing_ gets reported for the request and the client might see an abruptly terminated TCP connection.  I think that's what will happen if Dropshot panics before setting up the scope guard or from the code that runs in the guard.  So I think one option is: if Dropshot panics, it just doesn't run that guard at all.  Another option is that we report "no status".  When an HTTP request completes, it could have an ordinary HTTP status, or it may never complete (if we panicked at a bad time), or it completes with an indicator that the handler panicked.  I'd view this as analogous to the question of "what's the exit code of a Unix process that was terminated by SIGSEGV?"  The answer is it doesn't have one -- exit codes are only for processes not terminated by signals.

That'd be fine, but unless you have a compelling reason to handle these panics (and are fine with some panics not being handleable), I'd be inclined to just have the scope guard not run at all if the task panicked (if that's even possible...).  But I'm interested to know more about your use case to see if that makes sense!

steveklabnik:
> I don't love the idea of treating a panic as a handleable error, though. In the "programmer error" / "operational error" parlance, our general disposition among most of the Dropshot consumers I'm familiar with is to treat panics as programmer errors that reflect something deeply wrong, whose blast radius could include the whole process. I

For whatever it's worth, "a web server which has a handler that panics doesn't take down the whole server" was one of the direct motivations for being able to catch panics in the first place, as it's a common user expectation.

davepacheco:
> > I don't love the idea of treating a panic as a handleable error, though. In the "programmer error" / "operational error" parlance, our general disposition among most of the Dropshot consumers I'm familiar with is to treat panics as programmer errors that reflect something deeply wrong, whose blast radius could include the whole process. I
> 
> For whatever it's worth, "a web server which has a handler that panics doesn't take down the whole server" was one of the direct motivations for being able to catch panics in the first place, as it's a common user expectation.

I'm not surprised, as I've seen this approach used in other environments.  Joyent wound up deploying it in Node.js (via an `uncaughtException` handler).  It led to a number of serious outages requiring operator intervention that would have been inconsequential (and automatically resolved) had the process just crashed instead of trying to stay up.  We learned from those incidents that this pattern creates considerable risk that the server remains up when it's not actually healthy, when crashing and restarting could have restored service immediately.  Rust is undoubtedly in a better place to handle this pattern than Node.js, but the pattern remains fraught for the same reasons, and Rust is also in a much better position to avoid panicking so much in the first place.  That said, I don't care about convincing Dropshot consumers about any of this.  My goal is to make sure that _we_ avoid that problem and ideally avoid leading new users towards it who might otherwise choose differently.

nshalman:
> For our own stuff, we set panic = abort in release builds (which, incidentally, I think means that it doesn't matter what Dropshot does -- the program will call abort(3C) directly from the panic subsystem in Rust).

I assume this is https://doc.rust-lang.org/book/ch09-01-unrecoverable-errors-with-panic.html
```
[profile.release]
panic = 'abort'
```

This is helpful context for me to experiment with. Thank you!

An additional point that I think might be interesting is whether we can detect the difference between a panic originating in the handler vs a panic originating elsewhere in dropshot itself. As long as we can detect the difference, I think it's reasonable to give the user control over what happens when a handler panics while a dropshot panic should continue to do whatever it does.

Alternately it's possible that the mechanism offered should simply be panic = abort.

So I do think that perhaps this change is salvageable if I can prove it has no real impact when panic = abort and thus creates no behavior change for existing Oxide software. But per Steve's comment, if the server isn't going to crash and dropshot has detected that a panic specifically came from handler code and not its own code, dropshot should be allowed to return the 500 error to the user and log it with some appropriately critical error message.

Oh! That was the other thing. Is it really possible for a handler to return a status code of 200 to the user and then proceed to panic? I *think* the handler has to successfully return a value for dropshot to send a response to the user. I agree that a generic panic within dropshot in general could potentially trip over returning the 200 to the user and then crashing, but at that point you're already pretty screwed because 200 made it out.

I'm tempted to add a tiny bit of philosophical note about panic=abort somewhere in the README so that those who follow the same philosophy will know what to put into their dropshot software.

davepacheco:
> > For our own stuff, we set panic = abort in release builds (which, incidentally, I think means that it doesn't matter what Dropshot does -- the program will call abort(3C) directly from the panic subsystem in Rust).
> 
> I assume this is https://doc.rust-lang.org/book/ch09-01-unrecoverable-errors-with-panic.html
> 
> ```
> [profile.release]
> panic = 'abort'
> ```

Yup.

> An additional point that I think might be interesting is whether we can detect the difference between a panic originating in the handler vs a panic originating elsewhere in dropshot itself. As long as we can detect the difference, I think it's reasonable to give the user control over what happens when a handler panics while a dropshot panic should continue to do whatever it does.

I would phrase this as a new feature request, which is: allow a Dropshot server to treat [consumer] handler panics as 500s [without propagating them].  If we want to go down this road, I would move the panic check as close to the point where we invoke the user handler as possible so that we know that if we caught a panic, it was their code, not Dropshot's.  Then either turn it into the appropriate `HttpError` immediately or propagate a type that conveys which case we were in so that this code and do that.

> Oh! That was the other thing. Is it really possible for a handler to return a status code of 200 to the user and then proceed to panic? I _think_ the handler has to successfully return a value for dropshot to send a response to the user.

Agreed -- it'd have to be a panic after the handler function returns.

> I agree that a generic panic within dropshot in general could potentially trip over returning the 200 to the user and then crashing, but at that point you're already pretty screwed because 200 made it out.

Well, there's different degrees of "screwed".  If the program crashes, then:

- The client can (and should) detect this because they won't get a well-formed HTTP response.  They'll get say a non-zero `content-length` but then EOF before they've read all the bytes.
- Instrumentation will reflect a request started but never finished (this can be detected, though I don't know if existing tools do this well).

These are ordinary operational errors that can happen and are handleable.  The thing I was worried about is _reporting_ that request as a 500, and now the instrumentation is wrong about what actually happened.

> I'm tempted to add a tiny bit of philosophical note about panic=abort somewhere in the README so that those who follow the same philosophy will know what to put into their dropshot software.

That's okay but I don't think this is really Dropshot-specific.

nshalman:
A fresh version of this. Please don't review it just yet. I need to review it myself a couple of times as it's LLM generated and I haven't had a chance to put it through its paces. I am trying to modernize my tracing work and this was in that branch as well.

nshalman:
After reviewing the code and thinking a bit more. I am closing this one out. IF someone really wants to resurrect this idea, https://github.com/oxidecomputer/dropshot/commit/1bedf828be1b512c3b829b7f18c3fcfea8288ceb is probably a decent approach. So while using panic helped me write incredibly concise test code, since no one else has come into this PR asking it to land, I don't think there is enough demand for dropshot to handle panics to merit adding this code path. The https://github.com/oxidecomputer/dropshot/commit/615affb384aa4e377d182f06a4c90c3d3f6ae263 change is worthwhile and I will PR it separately in a fresh PR after I clean it up to my satisfaction.

I have a new approach to the otel functionality which I hope to PR as well.

PR #1382, opened 2025-07-30
Allow up to 5 extractors

### LINKED ISSUE #1372: Allow one more extractor for endpoint
I need a fourth extractor in my endpoint, which currently does not seem to be possible.

My endpoint is defined like this
```rust
#[endpoint {
    method = POST,
    path = "/upload/{id}/chunk",
}]
pub async fn upload_chunk(
    rqctx: RequestContext<AppContext>,
    auth_headers: Header<AuthHeaders>,
    query: Query<UploadChunkQuery>,
    path: DsPath<UploadPath>,
    body: UntypedBody,
) -> Result<HttpResponseUpdatedNoContent, HttpError>
```
which gives the following error
```
118 |   pub async fn upload_chunk(
    |                ^^^^^^^^^^^^ the trait `dropshot::handler::HttpHandlerFunc<_, _, _>` is not implemented for fn item `fn(RequestContext<AppContext>, Header<AuthHeaders>, Query<...>, ..., ...) -> ... {upload_chunk}`
```

This seems to be due to the fact that only up to 3 extractors are supported, as documented here:
https://github.com/oxidecomputer/dropshot/blob/83f78e78735a98578b7601c25ef20f428283d667/dropshot/src/lib.rs#L344-L347

I am wondering what the reason for this is and whether you would be willing to accept a PR that increases this limit to 4. Or is there another way to have more extractors that I overlooked?

The following changes on a local version of dropshot make my above code compile, but this may have some side-effects that I don't fully understand:
```diff
--- a/dropshot/src/extractor/common.rs
+++ b/dropshot/src/extractor/common.rs
@@ -207,3 +207,4 @@ macro_rules! impl_rqextractor_for_tuple {
 // extractors and exactly one exclusive extractor.
 impl_rqextractor_for_tuple!(S1);
 impl_rqextractor_for_tuple!(S1, S2);
+impl_rqextractor_for_tuple!(S1, S2, S3);
diff --git a/dropshot/src/handler.rs b/dropshot/src/handler.rs
index de589a4..9625e43 100644
--- a/dropshot/src/handler.rs
+++ b/dropshot/src/handler.rs
@@ -680,6 +680,7 @@ impl_HttpHandlerFunc_for_func_with_params!();
 impl_HttpHandlerFunc_for_func_with_params!((0, T0));
 impl_HttpHandlerFunc_for_func_with_params!((0, T1), (1, T2));
 impl_HttpHandlerFunc_for_func_with_params!((0, T1), (1, T2), (2, T3));
+impl_HttpHandlerFunc_for_func_with_params!((0, T1), (1, T2), (2, T3), (3, T4));
```

JuliDi:
Fixes #1372 by adding the impl for two more (now up to 4) `SharedExtractor`s.
Also adds tests for handlers with up to 5 extractors in total.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -15,6 +15,8 @@
 
 https://github.com/oxidecomputer/dropshot/compare/v0.16.2\...HEAD[Full list of commits]
 
+* https://github.com/oxidecomputer/dropshot/pull/1382[#1382] Allow 2 more `SharedExtractor`s to be used with the `endpoint` macro
+
 == 0.16.2 (released 2025-05-21)
 
 https://github.com/oxidecomputer/dropshot/compare/v0.16.1\...v0.16.2[Full list of commits]
--- a/dropshot/src/extractor/common.rs
+++ b/dropshot/src/extractor/common.rs
@@ -207,3 +207,4 @@ macro_rules! impl_rqextractor_for_tuple {
 // extractors and exactly one exclusive extractor.
 impl_rqextractor_for_tuple!(S1);
 impl_rqextractor_for_tuple!(S1, S2);
+impl_rqextractor_for_tuple!(S1, S2, S3);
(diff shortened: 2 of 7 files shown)


davepacheco:
Thanks!

Issue #1247, opened 2025-01-31
Panic when generating an API for a struct with an enum `serde(flatten)`ed into a struct.


blinsay:
Following up on #1244, I've been trying to use a struct with a `serde(flatten)` field as a return type, and I'm getting a panic when generating an openapi description. The panic seems to only happen when calling `write` on a generated stub description - I can serve traffic fine!

Here's a repro:

```rust
use dropshot::HttpError;
use dropshot::HttpResponseOk;
use dropshot::RequestContext;
use schemars::JsonSchema;
use serde::Deserialize;
use serde::Serialize;

fn main() {
    let description = example_api_mod::stub_api_description().unwrap();
    let openapi = description.openapi("example", "0.1.0".parse().unwrap());

    openapi.write(&mut std::io::stdout().lock()).unwrap();
}

#[derive(Serialize, Deserialize, JsonSchema)]
struct CoolStruct {
    #[serde(flatten)]
    cool_enum: CoolEnum,

    another_thing: u16,
}

#[derive(Serialize, Deserialize, JsonSchema)]
#[serde(tag = "type")]
enum CoolEnum {
    Foo(Foo),
    Bar(Bar),
}

#[derive(Serialize, Deserialize, JsonSchema)]
struct Foo {
    thing_one: String,
}

#[derive(Serialize, Deserialize, JsonSchema)]
struct Bar {
    thing_one: String,
    thing_two: String,
}

#[dropshot::api_description]
trait ExampleApi {
    type Context;

    #[endpoint {
        method = GET,
        path = "/hello",
    }]
    async fn hello(
        ctx: RequestContext<Self::Context>,
    ) -> Result<HttpResponseOk<CoolStruct>, HttpError>;
}
```
 <details> 

<summary>Here's the full panic</summary>

```
thread 'main' panicked at /Users/benl/.cargo/registry/src/index.crates.io-6f17d22bba15001f/dropshot-0.15.1/src/schema_util.rs:466:31:
a schema can't have both a type and subschemas:
{
  "type": "object",
  "oneOf": [
    {
      "type": "object",
      "required": [
        "thing_one",
        "type"
      ],
      "properties": {
        "thing_one": {
          "type": "string"
        },
        "type": {
          "type": "string",
          "enum": [
            "Foo"
          ]
        }
      }
    },
    {
      "type": "object",
      "required": [
        "thing_one",
        "thing_two",
        "type"
      ],
      "properties": {
        "thing_one": {
[...]

PR #1388, opened 2025-08-05
[docs] update extractor docs: fix a typo, increase verbosity

### LINKED ISSUE #1244: Trying to use an enum as a `Query` parameter causes a panic
I've got an enum as part of my API and tried to use it as a `Query` type, and immediately got a panic from dropshot while generating the API. I expected this to work, since it looks like `PaginationParams` and `WhichParams` are enums and there's no warnings in the docs.

Here's a minimal repro: 

```rust
use dropshot::HttpError;
use dropshot::HttpResponseOk;
use dropshot::Query;
use dropshot::RequestContext;
use schemars::JsonSchema;
use serde::Deserialize;

fn main() {
    let description = example_api_mod::stub_api_description().unwrap();
    let openapi = description.openapi("example", "0.1.0".parse().unwrap());

    openapi.write(&mut std::io::stdout().lock()).unwrap();
}

#[derive(Deserialize, JsonSchema)]
#[serde(tag = "type")]
enum QueryParamEnum {
    Foo(String),
    Bar { n: usize },
}

#[dropshot::api_description]
trait ExampleApi {
    type Context;

    #[endpoint {
        method = GET,
        path = "/hello",
    }]
    async fn do_stuff(
        ctx: RequestContext<Self::Context>,
        query: Query<QueryParamEnum>,
    ) -> Result<HttpResponseOk<()>, HttpError>;
}
```

The panic dumps the schema for `QueryParamEnum` and suggests wrapping it in a struct:

```
thread 'main' panicked at /Users/benl/.cargo/registry/src/index.crates.io-6f17d22bba15001f/dropshot-0.15.1/src/schema_util.rs:35:13:
while generating schema for QueryParamEnum (parameters): invalid type: {
  "title": "QueryParamEnum",
  "oneOf": [
    {
      "type": [
        "object",
        "string"
...
}
(hint: this appears to be an enum, which needs to be wrapped in a struct)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

If I try to wrap everything in a struct like the error message suggests I do, I still get a panic.

```rust
#[derive(Deserialize, JsonSchema)]
struct QueryParam {
    inner: QueryParamEnum,
}

// snip

    #[endpoint {
        method = GET,
        path = "/hello",
    }]
    async fn do_stuff(
        ctx: RequestContext<Self::Context>,
        query: Query<QueryParam>,
    ) -> Result<HttpResponseOk<()>, HttpError>;

```
```
thread 'main' panicked at src/main.rs:9:63:
called `Result::unwrap()` on an `Err` value: ApiDescriptionBuildErrors { errors: [ApiDescriptionRegisterError { operation_id: "do_stuff", message: "for endpoint do_stuff the parameter 'inner' must have a scalar type" }] }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

joelsa:
This fixes a copy-paste typo in the docs for https://docs.rs/dropshot/latest/dropshot/struct.Header.html and adds some information about how the types for extractors should look like.

Like the user in #1244, I've run into the same issue that enums need to be wrapped in an external struct and also need to be flattened with the serde_flatten attribute. I believe this PR makes it more clear and thus fixes #1244, as the actual bug (#1247) was fixed in #1330 and only a documentation issue remained.

### DIFF
--- a/dropshot/src/lib.rs
+++ b/dropshot/src/lib.rs
@@ -343,7 +343,7 @@
 //!
 //! Generally, the type parameter for the `Query`, `Header`, and `Path` extractors
 //! should be Rust structs. The struct's field _names_ correspond to the keys in
-//! the thing being extracted (i.e., they correspond to qquery parameter names
+//! the thing being extracted (i.e., they correspond to query parameter names
 //! for `Query`, header names for `Header`, and path component names for `Path`).
 //! The struct's field _values_ should generally be Rust primitives, strings,
 //! or enums containing no data.
--- a/dropshot/src/lib.rs
+++ b/dropshot/src/lib.rs
@@ -323,13 +323,14 @@
 //!
 //! * [`Query`]`<Q>` extracts parameters from a query string, deserializing them
 //!   into an instance of type `Q`. `Q` must implement `serde::Deserialize` and
-//!   `schemars::JsonSchema`.
+//!   `schemars::JsonSchema`. See below for additional restrictions.
 //! * [`Path`]`<P>` extracts parameters from HTTP path, deserializing them into
 //!   an instance of type `P`. `P` must implement `serde::Deserialize` and
-//!   `schemars::JsonSchema`.
+//!   `schemars::JsonSchema`. See below for additional restrictions.
 //! * [`Header`]`<H>` extracts parameters from HTTP headers, deserializing
 //!   them into an instance of type `H`. `H` must implement
-//!   `serde::Deserialize` and `schemars::JsonSchema`.
+//!   `serde::Deserialize` and `schemars::JsonSchema`. See below for additional
+//!   restrictions.
 //! * [`TypedBody`]`<J>` extracts content from the request body by parsing the
 //!   body as JSON (or form/url-encoded) and deserializing it into an instance
 //!   of type `J`. `J` must implement `serde::Deserialize` and `schemars::JsonSchema`.
@@ -339,6 +340,15 @@
 //! * [`RawRequest`] provides access to the underlying [`hyper::Request`].  The
 //!   hope is that this would generally not be needed.  It can be useful to
 //!   implement functionality not provided by Dropshot.
+//!
+//! Generally, the type parameter for the `Query`, `Header`, and `Path` extractors
+//! should be Rust structs. The struct's field _names_ correspond to the keys in
+//! the thing being extracted (i.e., they correspond to qquery parameter names
+//! for `Query`, header names for `Header`, and path component names for `Path`).
+//! The struct's field _values_ should generally be Rust primitives, strings,
+//! or enums containing no data.
+//! There is no facility for automatically parsing the values _again_ (e.g.,
+//! as JSON), which means nested values or enums with data cannot be supported.
 //!
 //! `Query` and `Path` impl `SharedExtractor`.  `TypedBody`, `UntypedBody`,
 //! `StreamingBody`, and `RawRequest` impl `ExclusiveExtractor`.  Your function
--- a/dropshot/src/extractor/body.rs
+++ b/dropshot/src/extractor/body.rs
@@ -30,9 +30,11 @@ use std::fmt::Debug;
 // type.  Only JSON is currently supported.
 
 /// `TypedBody<BodyType>` is an extractor used to deserialize an instance of
-/// `BodyType` from an HTTP request body.  `BodyType` is any structure of yours
-/// that implements `serde::Deserialize`.  See this module's documentation for
-/// more information.
+/// `BodyType` from an HTTP request body.  `BodyType` may be any struct of yours
+/// that implements [serde::Deserialize] and [schemars::JsonSchema], where
+/// primitives and enums have to be wrapped in an outer struct and enums need
+/// to be flattened using the `#[serde(flatten)]` attribute.  See this module's
+/// documentation formore information.
 #[derive(Debug)]
 pub struct TypedBody<BodyType: JsonSchema + DeserializeOwned + Send + Sync> {
     inner: BodyType,
--- a/dropshot/src/extractor/body.rs
+++ b/dropshot/src/extractor/body.rs
@@ -28,13 +28,10 @@ use std::fmt::Debug;
 
 // TypedBody: body extractor for formats that can be deserialized to a specific
 // type.  Only JSON is currently supported.
-
 /// `TypedBody<BodyType>` is an extractor used to deserialize an instance of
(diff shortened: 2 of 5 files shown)


davepacheco (changes requested):
Thanks for the improvement!  There seems to remain some confusion here.

The text you added was:

> [`...Type` may be any struct of yours that implements `serde::Deserialize` and `schemars::JsonSchema`] where primitives and enums have to be wrapped in an outer struct and enums need to be flattened using the `#[serde(flatten)]` attribute

It's definitely not true that primitives have to be wrapped for all these extractors.  I just tested this change to `dropshot/examples/basic.rs`:

```diff
dap@ivanova dropshot $ git diff
diff --git a/dropshot/examples/basic.rs b/dropshot/examples/basic.rs
index 6dc4a81..e508503 100644
--- a/dropshot/examples/basic.rs
+++ b/dropshot/examples/basic.rs
@@ -96,18 +96,18 @@ async fn example_api_get_counter(
 }]
 async fn example_api_put_counter(
     rqctx: RequestContext<ExampleContext>,
-    update: TypedBody<CounterValue>,
+    update: TypedBody<u64>,
 ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
     let api_context = rqctx.context();
     let updated_value = update.into_inner();
 
-    if updated_value.counter == 10 {
+    if updated_value == 10 {
         Err(HttpError::for_bad_request(
             Some(String::from("BadInput")),
-            format!("do not like the number {}", updated_value.counter),
+            format!("do not like the number {}", updated_value),
         ))
     } else {
-        api_context.counter.store(updated_value.counter, Ordering::SeqCst);
+        api_context.counter.store(updated_value, Ordering::SeqCst);
         Ok(HttpResponseUpdatedNoContent())
     }
 }

```

and it works as expected.  I expect enums would work as types for `TypedBody` as well.

Enums even work as query parameter values, provided they have no data.  This one also worked as expected:

```diff
diff --git a/dropshot/examples/basic.rs b/dropshot/examples/basic.rs
index 6dc4a81..71ab92f 100644
--- a/dropshot/examples/basic.rs
+++ b/dropshot/examples/basic.rs
@@ -88,6 +88,17 @@ async fn example_api_get_counter(
     }))
 }
 
+#[derive(Deserialize, JsonSchema)]
+struct MyCounter {
+    value: MyCounterValue,
+}
+
+#[derive(Deserialize, JsonSchema)]
+enum MyCounterValue {
+    Big,
+    Small,
+}
+
 /// Update the current value of the counter.  Note that the special value of 10
 /// is not allowed (just to demonstrate how to generate an error).
 #[endpoint {
@@ -96,18 +107,21 @@ async fn example_api_get_counter(
 }]
 async fn example_api_put_counter(
     rqctx: RequestContext<ExampleContext>,
-    update: TypedBody<CounterValue>,
+    update: dropshot::Query<MyCounter>,
 ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
     let api_context = rqctx.context();
-    let updated_value = update.into_inner();
+    let updated_value = match update.into_inner().value {
+        MyCounterValue::Big => 100,
+        MyCounterValue::Small => 10,
+    };
 
-    if updated_value.counter == 10 {
+    if updated_value == 10 {
         Err(HttpError::for_bad_request(
             Some(String::from("BadInput")),
-            format!("do not like the number {}", updated_value.counter),
+            format!("do not like the number {}", updated_value),
         ))
     } else {
-        api_context.counter.store(updated_value.counter, Ordering::SeqCst);
+        api_context.counter.store(updated_value, Ordering::SeqCst);
         Ok(HttpResponseUpdatedNoContent())
     }
 }

```

What I think is true is that for the `Query`, `Header`, and `Path` extractors, the type parameter on the extractor must describe keys and values, since query parameters, headers, and path parameters are by definition key-value pairs.  In practice, we basically expect people to use a `struct` here.  (Primitives are definitely out because what would the key be?  I bet you could use an `enum` if you made sure that its `JsonSchema` at the top level was an `object`, rather than a `oneOf`, but I think you'd have to impl `JsonSchema` yourself because even an adjacently-tagged enum has a `oneOf` at the top level.)

I believe it's also true that the values must be flat, not nested.  This comes from the fact that the underlying thing (query parameter value, header value, component in a URL path) is a string and there is no obvious way to parse that as a collection.  (You could relate this to the fact that `FromStr` isn't impl'd for `Vec<T>`.  There are many ways to parse a `Vec<T>` from a string, but none so obvious and universal that Rust would do it for you.  Neither does Dropshot.)

Tangentially, long ago we discussed having these type parameters impl some new trait that's more specific than `Deserialize`.  That could make these restrictions evident at compile-time.  This would be nice!

---

What does any of this mean for this PR?  For `TypedBody`, I would just make the small change to mention `JsonSchema`.  The other restrictions don't apply to it.

For the other three, I would probably clarify this in the top-level crate docs instead of copying it into each one.  There are already bullets like:

```
Query<Q> extracts parameters from a query string, deserializing them into an instance of type Q. Q must implement serde::Deserialize and schemars::JsonSchema.
```

I'd add to these bullets: "See below for additional restrictions."

Then below that, add a new paragraph saying something like:

> Generally, the type parameter for the `Query`, `Header`, and `Path` extractors should be Rust structs.  The struct's field names correspond to the keys in the thing being extracted (i.e., they correspond to query parameter names for `Query`, header names for `Header`, and path component names for `Path`).  The struct's field _values_ should generally be Rust primitives, strings, or enums containing no data.  There is no facility for automatically parsing the values _again_ (e.g., as JSON), which means nested values or enums with data cannot be supported.

Then we could emphasize the sentence in each extractor's docs saying to look at the top-level crate docs.

joelsa:
Thanks for this write-up @davepacheco. This really clarifies what should work and what shouldn't and it's already good documentation in itself. I'll update the PR and try to reflect that in the doc comments.

dropshot/src/lib.rs
```diff
@@ -340,6 +341,15 @@
 //!   hope is that this would generally not be needed.  It can be useful to
 //!   implement functionality not provided by Dropshot.
 //!
+//! Generally, the type parameter for the `Query`, `Header`, and `Path` extractors
+//! should be Rust structs. The struct's field _names_ correspond to the keys in
+//! the thing being extracted (i.e., they correspond to qquery parameter names
```
davepacheco, line 346:
```suggestion
//! the thing being extracted (i.e., they correspond to query parameter names
```

davepacheco (commented):
Looks good, aside from the one typo!

davepacheco (approved):
Fixed the typo

Issue #1373, opened 2025-07-18
Header names with capital letters are not deserialized


JuliDi:
Headers with (renamed) names that include capital letters fail to deserialize.

For example:

```rust
#[derive(Serialize, Deserialize, JsonSchema)]
pub struct MyHeaders {
    header_a: String,
    #[serde(rename = "X-Header-B")]
    header_b: String,
}

#[endpoint {
    method = GET,
    path = "/test",
}]
pub async fn get_test(
    rqctx: RequestContext<AppContext>,
    header_params: Header<MyHeaders>,
) -> Result<HttpResponseOk<()>, HttpError> {
    info!("rqctx headers: {:?}", rqctx.request.headers());
    let headers = header_params.into_inner();
    info!("Headers: {:?}, {:?}", headers.header_a, headers.header_b);

    Ok(HttpResponseOk(()))
}
```

Using the following request

<img width="595" height="404" alt="Image" src="https://github.com/user-attachments/assets/f20447ca-0de4-427f-9076-70bc87a9de35" />

causes the following error:
```
thread 'tokio-runtime-worker' panicked at /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/dropshot-0.16.2/src/extractor/header.rs:53:44:
called `Result::unwrap()` on an `Err` value: "missing field `X-Header-B`"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Jul 18 08:31:05.526 ERRO handler panicked; propogating panic, uri: /test, method: GET, req_id: 921b5a49-668e-4a6a-9e48-284ff37f91cb, remote_addr: 127.0.0.1:60506, local_addr: 127.0.0.1:8020
```

Removing the `#[serde(rename = "X-Header-B")]` reveals in the logs:
```
rqctx headers: {"accept": "application/json, text/plain, */*", "user-agent": "bruno-runtime/2.7.0", "header_a": "header_a value", "x-header-b": "header_b value", "header_b": "header_b value", "request-start-time": "1752827600143", "accept-encoding": "gzip, compress, deflate, br", "host": "localhost:8020", "connection": "keep-alive"}
```

So the error seems to be caused by "X-Header-B" being renamed to "x-header-b" in the request context. Since HTTP headers are not case-sensitive, I think deserialization should generally not be case sensitive either.

If someone could tell me where to best introduce this change, I am happy to make a PR for this.

[...]

PR #1403, opened 2025-08-21
header marshaling should be case-insensitive and more robust

### LINKED ISSUE #1380: dropshot panics when it fails to extract headers
Details in #1373.  `from_map()` returned an error for what seems like a pretty ordinary reason (a required field was missing, but this could as well have been a type error or something else).  Then dropshot panicked.  I expect it should return a 400-level error similar to getting bad JSON in the request body.

@ahl can you take a look at this?

ahl:
Fixes #1380, #1373.

The `Header` extractor (which I wrote) was very poor. In particular it didn't consider the case of required header (or of deserialization failing for *any* reason). This PR intends to address the issues of robustness and add proper handling for the case-insensitivity of headers. We do so while preserving desired casing for the OpenAPI output.

An alternative approach would have been to **require** lower-case naming for headers with the ability to use `schemars` attributes to direct the OpenAPI output (e.g. `#[serde(rename = "x-my-header")] #[schemars(rename = "X-My-Header")]`). But this seemed confusing and error-prone.

### DIFF
--- a/dropshot/src/extractor/header.rs
+++ b/dropshot/src/extractor/header.rs
@@ -9,7 +9,7 @@ use schemars::JsonSchema;
 use serde::de::DeserializeOwned;
 
 use crate::{
-    from_map::from_map, ApiEndpointBodyContentType,
+    from_map::from_map_insensitive, ApiEndpointBodyContentType,
     ApiEndpointParameterLocation, HttpError, RequestContext, RequestInfo,
     ServerContext,
 };
@@ -17,11 +17,20 @@ use crate::{
 use super::{metadata::get_metadata, ExtractorMetadata, SharedExtractor};
 
 /// `Header<HeaderType>` is an extractor used to deserialize an instance of
-/// `HeaderType` from an HTTP request's header values. `PathType` may be any
+/// `HeaderType` from an HTTP request's header values. `HeaderType` may be any
 /// structure that implements [serde::Deserialize] and [schemars::JsonSchema].
 /// While headers are accessible through [RequestInfo::headers], using this
 /// extractor in an entrypoint causes header inputs to be documented in
 /// OpenAPI output. See the crate documentation for more information.
+///
+/// Note that (unlike the [`Query`] and [`Path`] extractors) headers are case-
+/// insensitive. You may rename fields with mixed casing (e.g. by using
+/// #[serde(rename = "X-Header-Foo")]) and that casing will appear in the
+/// OpenAPI document output. Case-insensitive name conflicts may lead to
+/// unexpected behavior, and should be avoided. For example, only one of the
+/// conflicting fields may be deserialized, and therefore deserialization may
+/// fail if any conflicting field is required (i.e. not an `Option<T>` type)
+#[derive(Debug)]
 pub struct Header<HeaderType: DeserializeOwned + JsonSchema + Send + Sync> {
     inner: HeaderType,
 }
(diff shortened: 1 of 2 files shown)


dropshot/src/extractor/header.rs
```diff
@@ -50,8 +59,14 @@ where
         .map_err(|message: http::header::ToStrError| {
             HttpError::for_bad_request(None, message.to_string())
         })?;
-    let x: HeaderType = from_map(&headers).unwrap();
-    Ok(Header { inner: x })
+    println!("headers: {headers:?}");
```
smklein, line 62: Leftover println - remove?

dropshot/src/extractor/header.rs
```diff
@@ -9,19 +9,28 @@ use schemars::JsonSchema;
...
 /// structure that implements [serde::Deserialize] and [schemars::JsonSchema].
 /// While headers are accessible through [RequestInfo::headers], using this
 /// extractor in an entrypoint causes header inputs to be documented in
 /// OpenAPI output. See the crate documentation for more information.
+///
+/// Note that (unlike the [`Query`] and [`Path`] extractors) headers are case-
+/// insensitive. You may rename fields with mixed casing (e.g. by using
+/// #[serde(rename = "X-Header-Foo")]) and that casing will appear in the
+/// OpenAPI document output. Case-insensitive name conflicts may lead to
+/// unexpected behavior, and should be avoided. For example, only one of the
```
smklein, line 30: Trying to make sense of "Case-insensitive name conflicts may lead to unexpected behavior, and should be avoided".

Who is this advice for? Why not just say "Since HTTP headers are case-insensitive, we automatically treat headers as lowercase. Upper-case characters are still allowed for OpenAPI documents, but they are converted to lowercase when deserializing and dealing with extraction conflicts".

dropshot/src/extractor/header.rs
```diff
@@ -71,3 +86,55 @@ where
...
+        let request =
+            hyper::Request::builder().uri("http://localhost").body(()).unwrap();
+        let info = RequestInfo::new(&request, addr);
+
+        let parsed = http_request_load_header::<TestHeaders>(&info);
+        assert!(parsed.is_err());
+
+        let request = hyper::Request::builder()
+            .header("header_a", "header_a value")
+            .header("X-Header-B", "header_b value")
```
smklein, line 121: If we used `x-hEaDeR-b` here as the key, would this value still make it to the server?

dropshot/src/from_map.rs
```diff
@@ -46,22 +59,68 @@ impl MapValue for String {
     }
 }
 
+// To handle headers, which are case-insensitive, we use a trait to adapt
+// deserialization of structs to be case-insensitive.
+
+trait Casing {
+    fn rename(fields: &'static [&'static str]) -> BTreeMap<String, String>;
```
smklein, line 66: This could use some docs?

(e.g., what are the output "Key/Values" in the BTreeMap here?)

dropshot/src/from_map.rs
```diff
@@ -46,22 +59,68 @@ impl MapValue for String {
...
+impl Casing for CaseSensitive {
+    fn rename(_fields: &'static [&'static str]) -> BTreeMap<String, String> {
+        Default::default()
+    }
+}
+impl Casing for CaseInsensitive {
+    fn rename(fields: &'static [&'static str]) -> BTreeMap<String, String> {
+        fields
+            .iter()
+            .map(|field| (field.to_lowercase(), field.to_string()))
```
smklein, line 81: Is the lifetime too much of a pain to return the "value" here as a `&'static str`? seems unnecessary to perform the string allocation (via `field.to_string()`) on the case-unchanged string here

smklein (approved):

PR #1413, opened 2025-08-29
accept body parameters of the form application/subtype+json

### LINKED ISSUE #1383: dropshot should support other application/json aliases
While working on SCIM v2 support within nexus, @jmpesp and I ran into an issue with the request bodies coming from identity providers such as Okta. It turns out SCIM v2 is using a content-type of `application/scim+json` as defined in [RFC 7644 Section 8.1](https://datatracker.ietf.org/doc/html/rfc7644#section-8.1).

Dropshot currently chokes on this because the built-in extractors for the request body have a whitelist of mime types defined here:
https://github.com/oxidecomputer/dropshot/blob/45dbc19af783e9dab8b6928c44f1e7b190501818/dropshot/src/api_description.rs#L332-L341

Which is later checked here:
https://github.com/oxidecomputer/dropshot/blob/45dbc19af783e9dab8b6928c44f1e7b190501818/dropshot/src/extractor/body.rs#L153-L155

There are a few paths forward discussed on the control plane huddle:
- Implement a custom extractor for your json like alias
- Simply add the `scim+json` type to the defined enum
- Maintain a list of json mime type aliases and expose them via the same enum
- Another approach not yet realized...

The returned body out of dropshot can be easily be worked around at this time as the Response type lets you set header value yourself.

ahl:
To support SCIM support, dropshot needs to be able to accept requests whose `content-type` header is specifically `application/scim+json`. Taking this apart, this MIME type means "the SCIM protocol, serialized as JSON". One can think of it as strictly a subset of JSON i.e. all messages are in the SCIM protocol, but all are also valid JSON.

We have several options for how to address this. It would seem reasonable and robust to allow an `ApiDescription` (i.e. a collection of endpoints) for which we say "the input and response bodies are all SCIM protocol so application/scim+json is applicable" (and that might be inclusive or exclusive of application/json i.e. the `scim+` part might be optional). We could include this in the OpenAPI output to document the content type for body parameters and responses. This idea, however, requires something along the lines of what's imagined in PR #922 and hasn't been completed.

Instead, this PR implements a more expedient approach: treat `content-type: application/X+json` as `application/json`. This isn't wrong *per se* as the former is a subset of the latter, but it doesn't allow a consumer to impose a distinction between the two.

Closes #1383

### DIFF
--- a/dropshot/src/api_description.rs
+++ b/dropshot/src/api_description.rs
@@ -353,9 +353,7 @@ impl ApiEndpointBodyContentType {
 /// Split the mime type in to the type, subtype, and optional suffix
 /// components.
 fn mime_split(mime_type: &str) -> Option<(&str, &str, Option<&str>)> {
-    let mut parts = mime_type.splitn(2, '/');
-    let type_ = parts.next()?;
-    let rest = parts.next()?;
+    let (type_, rest) = mime_type.split_once('/')?;
     let mut sub_parts = rest.splitn(2, '+');
     let subtype = sub_parts.next()?;
     let suffix = sub_parts.next();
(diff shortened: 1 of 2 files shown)


sunshowers (approved):
Thanks!

PR #1439, opened 2025-09-28
migrate to Rust 2024

sunshowers:
Also bump the MSRV to Rust 1.85, and add CI for that.


sunshowers:
This would need an update to the GHA required config, since job names have changed.

ahl:
Obviated by #1605 and #1608

PR #1448, opened 2025-10-02
Streaming gzip response compression

### LINKED ISSUE #221: Gzip response bodies
Substantially reduce response sizes by gzipping response body and adding `content-encoding: gzip` to headers. Probably simplest to make this configurable server-wide.

### Extra features to punt on

- Respect `accept-encoding: identity` request header by not compressing even if compression feature is turned on server-wide

david-crespo:
Closes #221

 * Compression is opt-in: `compression` config field defaults to `CompressionConfig::None` to avoid behavior changes on upgrade
* Compression depends on `Accept-Encoding` request header (with RFC 9110 quality factor support) and the MIME type matching a list of compressible types (JSON, ndjson, text/*, XML, JavaScript, and RFC 6839 structured suffixes like +json/+xml). SSE (text/event-stream) is excluded to avoid adding latency.
* `Vary: Accept-Encoding` header is added to all compressible responses, even when compression isn't applied, for correct cache behavior
* Don't bother compressing responses where we know the size and it's less than 512 bytes
* `NoCompression` extension lets an endpoint opt out of compression

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -76,13 +76,12 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.4.32"
+version = "0.4.41"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a89bce6054c720275ac2432fbba080a66a2106a44a1b804553930ca6909f4e0"
+checksum = "d0f9ee0f6e02ffd7ad5816e9464499fba7b3effd01123b515c41d1697c43dad1"
 dependencies = [
  "compression-codecs",
  "compression-core",
- "futures-core",
  "pin-project-lite",
  "tokio",
 ]
@@ -237,9 +236,9 @@ dependencies = [
 
 [[package]]
 name = "compression-codecs"
-version = "0.4.31"
+version = "0.4.37"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef8a506ec4b81c460798f572caead636d57d3d7e940f998160f52bd254bf2d23"
+checksum = "eb7b51a7d9c967fc26773061ba86150f19c50c0d65c887cb1fbe295fd16619b7"
 dependencies = [
  "compression-core",
  "flate2",
@@ -248,9 +247,9 @@ dependencies = [
 
 [[package]]
 name = "compression-core"
-version = "0.4.29"
+version = "0.4.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e47641d3deaf41fb1538ac1f54735925e275eaf3bf4d55c81b137fba797e5cbb"
+checksum = "75984efb6ed102a0d42db99afb6c1948f0380d1d91808d5529916e6c08b49d8d"
 
 [[package]]
 name = "concurrent-queue"
@@ -587,9 +586,9 @@ checksum = "e8c02a5121d4ea3eb16a80748c74f5549a5665e4c21333c6098f283870fbdea6"
 
 [[package]]
 name = "flate2"
-version = "1.1.4"
+version = "1.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dc5a4e564e38c699f2880d3fda590bedc2e69f3f84cd48b457bd892ce61d0aa9"
+checksum = "843fba2746e448b37e26a819579957415c8cef339bf08564fe8b7ddbd959573c"
 dependencies = [
  "crc32fast",
  "miniz_oxide",
@@ -1001,7 +1000,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.0",
+ "socket2 0.5.9",
  "system-configuration",
  "tokio",
  "tower-layer",


dropshot/tests/integration-tests/gzip.rs
```diff
@@ -0,0 +1,674 @@
...
+    let mut decoder = flate2::read::GzDecoder::new(compressed_data);
+    let mut decompressed = Vec::new();
+    decoder
+        .read_to_end(&mut decompressed)
+        .expect("Failed to decompress gzip data");
+    decompressed
+}
+
+#[tokio::test]
+async fn test_gzip_compression_with_accept_encoding() {
```
david-crespo, line 218: This is the primary happy path test for compression.

david-crespo:
I was confused about what is supposed to happen if the client can't handle a streaming response, but it seems all HTTP/1.1 clients must be able to handle chunked transfer encoding:

> A recipient MUST be able to parse the chunked transfer coding ([Section 4.1](https://datatracker.ietf.org/doc/html/rfc7230#section-4.1)) because it plays a crucial role in framing messages when the payload body size is not known in advance.

https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.1

Full answer from GPT-5 through the Codex CLI below. Pretty helpful. The line "HTTP/1.1 clients that advertise `Accept-Encoding: gzip` are already required to understand chunked responses." is technically true but confusingly phrased because it makes it sound like the requirement is somehow tied to header when it is not.

---

Dropshot only decides to gzip when `should_compress_response` returns true (`dropshot/src/compression.rs:100`), which already filters out cases that inherently need fixed-size bodies: no body statuses, `HEAD`, partial ranges, missing `Content-Type`, `SSE`, or content smaller than 512 bytes. Once we do compress, `apply_gzip_compression` wraps the body in a streaming Gzip encoder and explicitly strips `Content-Length`/`Accept-Ranges` (`dropshot/src/compression.rs:191` and `dropshot/src/compression.rs:238`). Hyper will therefore deliver the response with chunked transfer coding on HTTP/1.1, even if the original body was a single buffer. That's the trade-off for not buffering the full compressed entity up front—we never know the final size early enough to set `Content-Length`.

Why this doesn't break non-streaming clients: HTTP/1.1 clients that advertise `Accept-Encoding: gzip` are already required to understand chunked responses. Clients that cannot handle streaming should omit `Accept-Encoding` (they'll stay on the uncompressed path with a fixed length) or explicitly disable compression via the `NoCompression` extension. Our streaming endpoints continue to stream but remain uncompressed because they typically lack a compressible `Content-Type`, which `should_compress_response` uses to short-circuit (`dropshot/tests/integration-tests/gzip.rs:344`). Tests cover all combinations: buffered responses keep a fixed length (`dropshot/tests/integration-tests/streaming.rs:55`), streaming endpoints stay chunked (`dropshot/tests/integration-tests/streaming.rs:91`), and streaming + gzip gets rejected (`dropshot/tests/integration-tests/gzip.rs:366`).

Matrix (server view vs client capability):

- Server sends fixed-length body → clients that can't stream are fine; clients that can stream are also fine.
- Server sends chunked/streaming body (including gzip) → clients that can stream are fine; clients that can't must not have asked for compression, so they fall back to the first row.

If you have a population of `Accept-Encoding: gzip` clients that truly cannot process chunked bodies, the short-term mitigation is to strip or downgrade that header in front of Dropshot; longer term you could add an alternate compression path that buffers and sets `Content-Length`, but that's intentionally avoided in the current design.

ahl:
Thanks for the clarification. I see in the tests how you're validating the content-length header, etc. 

Should compression be configurable in the server settings? i.e. would it make sense to e.g. have compression=on be the default but allow one to make a server that never compresses responses?

dropshot/tests/integration-tests/streaming.rs
```diff
@@ -12,7 +12,7 @@ use crate::common;
 
 extern crate slog;
 
-fn api() -> ApiDescription<usize> {
+pub fn api() -> ApiDescription<usize> {
```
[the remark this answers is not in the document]
david-crespo: Took this out, instead implemented dedicated streaming endpoints in the gzip tests file.

dropshot/src/config.rs
```diff
@@ -63,6 +63,10 @@ pub struct ConfigDropshot {
     /// is made to deal with headers that appear multiple times in a single
     /// request.
     pub log_headers: Vec<String>,
+    /// Whether to enable gzip compression for responses when response contents
+    /// allow it and clients ask for it through the Accept-Encoding header.
+    /// Defaults to true.
+    pub compression: bool,
```
ahl: Are there more than 2 options when it comes to compression? I.e. might there be additional configuration in the future; rather than a bool might we want an enum? i.e. avoid further breaking changes

dropshot/src/server.rs
```diff
@@ -981,6 +990,33 @@ async fn http_request_handle<C: ServerContext>(
...
+        // Add Vary: Accept-Encoding header for all compressible content
+        // types. This needs to be there even if the response ends up not being
+        // compressed because it tells caches how to handle the possibility that
+        // responses could be compressed or not.
+        //
+        // This tells caches (like browsers and CDNs) that the response content
+        // depends on the value of the Accept-Encoding header. Without this, a
+        // cache might mistakenly serve a compressed response to a client that
+        // cannot decompress it, or serve an uncompressed response to a client
+        // that could have benefited from compression.
```
ahl:
```suggestion
        // Add Vary: Accept-Encoding header for all compressible content
        // types. This needs to be there even if the response ends up not being
        // compressed because it tells caches (like browsers and CDNs) how to handle the possibility that
        // responses could be compressed or not
        // depending on the value of the Accept-Encoding header. Without this, a
        // cache might mistakenly serve a compressed response to a client that
        // cannot decompress it, or serve an uncompressed response to a client
        // that could have benefited from compression.
```

Felt redundant. Not sure mine is better; do what you will here.

david-crespo:
From the commit message on 9be53957097b03a506750813b7da9730ef02ed90:

> Gemini's theory: When a test case requests the  /large-response
endpoint but compression is disabled (either by config, lack of header,
or rejection), the server attempts to write the full ~5KB JSON body to
the socket. On illumos, the default TCP socket buffer size is likely
smaller than on macOS/Linux, causing the server's write operation to
block because the test client never reads the data to drain the buffer.
When  teardown()  is called, the server hangs trying to finish the write
during graceful shutdown, eventually timing out.
> 
> The fix is to consume the response body in these tests.

The fix did work, so it seems directionally right even if the TCP socket buffer size isn't the precise reason. Not sure whether it's worth chasing down the reason further.

dropshot/src/compression.rs
```diff
@@ -0,0 +1,858 @@
...
+use hyper::body::{Body as HttpBodyTrait, Frame};
+use tokio_util::io::{ReaderStream, StreamReader};
+
+/// Marker type for disabling compression on a response.
+/// Insert this into response extensions to prevent compression:
+/// ```ignore
+/// response.extensions_mut().insert(NoCompression);
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct NoCompression;
```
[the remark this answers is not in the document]
ahl: meh up to you

dropshot/src/config.rs
david-crespo: Very good point. I checked what tower-http do and they have a lot more than a bool going on:

```rs
pub enum CompressionLevel {
    Fastest,
    Best,
    #[default]
    Default,
    Precise(i32),
}

pub(crate) struct AcceptEncoding {
    pub(crate) gzip: bool,
    pub(crate) deflate: bool,
    pub(crate) br: bool,
    pub(crate) zstd: bool,
}

pub struct CompressionLayer<P = DefaultPredicate> {
    accept: AcceptEncoding,
    predicate: P,
    quality: CompressionLevel,
}
```

That is overkill for us but it gives a sense of the way we might complicate this. The most likely one to me would be that we add brotli. I'll figure it out.

https://github.com/tower-rs/tower-http/blob/3bf1ba7b7893b57264dfe663165a2bc57a40d2c4/tower-http/src/compression/layer.rs#L7-L18

https://github.com/tower-rs/tower-http/blob/3bf1ba7b7893b57264dfe663165a2bc57a40d2c4/tower-http/src/compression_utils.rs#L425-L445

https://github.com/tower-rs/tower-http/blob/3bf1ba7b7893b57264dfe663165a2bc57a40d2c4/tower-http/src/compression_utils.rs#L18-L23

david-crespo:
As far as I know it is completely done and mergeable (minus the lockfile conflict, which I'll fix). I'll have the latest robots give it another review just in case.

Issue #1375, opened 2025-07-21
Migration to schemars 1.0


joelsa:
Schemars 1.0 has been out for about two months now and adds some interesting features, like support for [garde](https://github.com/jprochazk/garde) as an alternative to validator.

Per #1336, migration wasn't desired at the time, but I'm wondering if that's still the case now that schemars has continued to mature.

I took a quick look and think the migration might be doable in roughly a day's work, though I haven't dug into all the details yet. One important aspect is that 1.0 does make heavy use of Cow<'static, str> though.

Is there interest in migrating to 1.0 in the near-future at all? Happy to investigate further and work on it if so.

davepacheco:
CC @ahl

adamchalmers:
Just noting that schemars 1.0 adds some nice new kinds of expressivity. For example, it can handle `#[serde(untagged)]` on an enum variant, which lets you take a field that was previously, say, i32, and generalize it to an enum like 
```
#[serde(tag = "type")
MyEnum {
  String(String),
  #[serde(untagged)
  Number(i32),
}
```

Anyway. Just another vote for supporting schemars 1.0 for that reason.

PR #1449, opened 2025-10-04
Document blocker for upgrading schemars to v1.0

halvko:
Tried doing a project with dropshot and was surprised it was on an old version of schemars. I thought I would give back by upgrading. After spending some time updating schema constructions and checks to the new format, I encountered that we use a `SchemaSettings` setting which is no longer viable in the new version of `schemars`. I hope you are willing to add a comment like the one in this PR to upstream to avoid another person going down the same rabbit hole.

### DIFF
--- a/dropshot/Cargo.toml
+++ b/dropshot/Cargo.toml
@@ -87,10 +87,8 @@ version = "1.18.1"
 features = [ "serde", "v4" ]
 
 [dependencies.schemars]
-# While a 1.0 version is out, there is no way to generate an OpenAPIv3 spec without adding nullable
-# for all of the fields - see https://docs.rs/schemars/1.0.4/src/schemars/generate.rs.html#126 for
-# how SchemaSettings::option_nullable is handled now. As the other transformers are not public
-# creating our own version would take quite a bit of source code copy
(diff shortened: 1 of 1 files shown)


dropshot/Cargo.toml
```diff
@@ -87,6 +87,10 @@ version = "1.18.1"
 features = [ "serde", "v4" ]
 
 [dependencies.schemars]
+# While a 1.0 version is out, there is no way to generate an OpenAPIv3 spec without adding nullable
+# for all of the fields - see https://docs.rs/schemars/1.0.4/src/schemars/generate.rs.html#126 for
+# how SchemaSettings::option_nullable is handled now. As the other transformers are not public
+# creating our own version would take quite a bit of source code copy
```
davepacheco, line 93:
```suggestion
# While newer versions of schemars are available, they are not trivial for us to update to.
# See oxidecomputer/dropshot#1375 for more information.
```

davepacheco (commented):
Thanks for trying to help and sorry it didn't pan out.

halvko:
#1375 <- link such this issue will also be obvious for whoever may or may not attempt doing this at some point

PR #1470, opened 2025-11-05
Update tokio-rustls

maxammann:
The tokio-rustls version is pretty old and pulls in `ring` by default.


maxammann:
@halvko would be sweet if you can take a look at this one :) 

also this one might be done? https://github.com/oxidecomputer/dropshot/issues/1375

maxammann:
ah wait this is just documentation

maxammann:
@ahl woudl be sweet to have that in :)

davepacheco:
Note this updates both `tokio-rustls` _and_ `rustls` and it looks like the `rustls` includes a bunch of breaking changes: https://github.com/rustls/rustls/releases/tag/v%2F0.23.0

In particular I'm not sure if the default change to `aws-lc-rs` is a problem for us.  @ahl @sunshowers or @iliana does one of you know what the state of this is?  I'm not sure why Dropshot's been behind for so long.

sunshowers:
I'm not sure about the details but I vaguely recall @ahl knowing about this.

iliana:
From what I understand the ecosystem is generally shifting (has shifted?) from ring to aws-lc-rs. Several months ago we made some illumos build fixes upstream so that we could consume it and things have been smooth sailing since.

iliana:
We're several aws-lc-sys revisions behind on Omicron but I just checked to make sure the current one still builds on Helios, and it does.

maxammann:
At Zoo we switched to aws-lc-sys. `reqwests` also supports it through build flags wince 0.13.

We also have AWS dependencies which kind of force it down your throat. 

Apart from that the ecosystem has already switched away from ring. The upgrade also still allows you to go with ring and I'd say that is recommended until you switch over most of your projects.

ahl:
I think this is now obsolete -- sorry for the delay!!

PR #1475, opened 2025-11-13
add ClientSpecifiesVersionInHeader::on_missing

sunshowers:
Allow substituting a version if the header is missing. We're going to use this policy for the Nexus external API.

### DIFF
--- a/dropshot/src/versioning.rs
+++ b/dropshot/src/versioning.rs
@@ -165,6 +165,10 @@ impl ClientSpecifiesVersionInHeader {
     /// version (for backwards compatibility with older clients), or the newest
     /// supported version (in case clients are generally kept up-to-date but not
     /// all clients send the header).
+    ///
+    /// Using this function is not recommended if you control all clients—in
+    /// that case, arrange for clients to send the header instead. In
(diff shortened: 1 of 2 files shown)


dropshot/src/versioning.rs
```diff
@@ -148,7 +153,21 @@ impl ClientSpecifiesVersionInHeader {
         name: HeaderName,
         max_version: Version,
     ) -> ClientSpecifiesVersionInHeader {
-        ClientSpecifiesVersionInHeader { name, max_version }
+        ClientSpecifiesVersionInHeader { name, max_version, on_missing: None }
+    }
+
+    /// If the header is missing, use the provided version instead.
```
davepacheco: Can we update this comment to dissuade people from using this if they control the clients (and maybe explicitly say not for Oxide internal services?)

davepacheco (commented):
Looks good aside from the one comment request.  Can you update the CHANGELOG as well?

dropshot/src/versioning.rs
sunshowers: Done, and bolded.

PR #1555, opened 2026-03-02
Link example code in documentation

schneems:
- Before: Reference to `examples/basic.rs` is not clickable when viewed at https://docs.rs/dropshot/latest/dropshot/
- After: It is clickable


ahl:
Thanks for this contribution @schneems !

PR #1592, opened 2026-04-22
Update schemars to v1

joe-saronic:
This is an initial stab at the update. A bunch of stuff needed to change, and based on https://github.com/oxidecomputer/dropshot/issues/1375, a number of additional tools will need to be touched.

Clippy, format, test all passed locally.

Please feel free to use this branch as you see fit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -448,7 +448,7 @@ dependencies = [
 
 [[package]]
 name = "dropshot"
-version = "0.17.0"
+version = "0.18.0"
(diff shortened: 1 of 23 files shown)


ahl:
Migrating to schemars 1.0 is going to be an impactful and breaking change for all consumers. It will be a big pain in the neck for Oxide's constellation of repos in particular. As noted in #1375, if we were to allow for schemars 1.0, we'd want to maintain support for 0.8.

PR #1599, opened 2026-04-27
replace hyper-staticfile with minimal alternatives

ahl:
hyper-staticfile hasn't been updated in a while and was pulling in a few old transitive crates. I had claude replace its use with some on-hand alternatives, and it looks no worse and no more complicated.


ahl:
Validated the example like this:

```console
$ cargo run --example file_server
    Finished [`dev` profile [unoptimized + debuginfo]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.13s
     Running `target/debug/examples/file_server`
Apr 27 13:31:12.283 INFO listening, local_addr: 127.0.0.1:56570

...

$ curl http://localhost:56570/dropshot/src/lib.rs
// Copyright 2025 Oxide Computer Company

//! Dropshot is a general-purpose crate for exposing REST APIs from a Rust
//! program.  Planned highlights include:
//!
//! * Suitability for production use on a largely untrusted network.
//!   Dropshot-based systems should be high-performing, reliable, debuggable, and
//!   secure against basic denial of service attacks (intentional or otherwise).
...

```

dropshot/examples/file_server.rs
```diff
@@ -122,10 +125,8 @@ async fn static_content(
             )
         })?;
 
-        let file_access = hyper_staticfile::vfs::TokioFileAccess::new(file);
-        let file_stream =
-            hyper_staticfile::util::FileBytesStream::new(file_access);
-        let body = Body::wrap(hyper_staticfile::Body::Full(file_stream));
+        let stream = ReaderStream::new(file).map_ok(hyper::body::Frame::data);
+        let body = Body::wrap(StreamBody::new(stream));
```
davepacheco, line 129: This short bit of code makes my head spin.  Writing it down:

- we start with a `File`, which impls `AsyncRead`
- we wrap that in a `ReaderStream`, which converts it into a `Stream<Item=Result<Bytes, Error>>`
- we use `.map_ok(Frame::data)` to turn that into a `Stream<Item=Result<Frame<Bytes>, Error>>`
- we use `StreamBody::new` to wrap that into something that impls `http_body_util::Body<Data=Bytes>` and also still impls `Stream<Item=Result<Frame<Bytes>, Error>>`
- we use `dropshot::Body::wrap` to convert that to a `dropshot::Body`.

I thought maybe some of these steps were unnecessary but now I'm really not sure.  We want to return a `dropshot::Body` (although it's conceivable we could bypass that? -- I wouldn't bother digging in unless you want to).  That requires either having an `Into<Bytes>` (which would involve having read all the data into memory at once) or an `HttpBody<Data = Bytes>`.  I cannot find an easy way to see what impls this across all of the crates in play here.

davepacheco (approved):

Issue #1607, opened 2026-05-14
Re-export `slog`


alilleybrinker:
In general I think it's a good practice to re-export types from dependencies that are part of a crate's public API. Currently, `slog::Logger` is part of the public API for `ConfigLogging::to_logger` and `ServerBuilder::new`, but is not re-exported from `slog`. So as a user of `dropshot`, if I want to make a function that wraps (in my case) `ConfigLogging::to_logger`, I have to specify `slog` as a direct dependency of my crate alongside `dropshot` and ensure I manually keep the versions of `<my_crate> > slog` synced with `<my_crate> > dropshot > slog`. It's not _impossible_, but it is tedious.

It looks like `dropshot` currently re-exports one dependency: `semver`. Could the same be done for `slog`?

Separate, but related: when re-exporting multiple dependent crates, you may want to put them in a module together. I've seen it called `deps` elsewhere. Your call!

Also, thanks for building `dropshot`! I'm really enjoying using it so far!

davepacheco:
Hm.  I don't think this would be bad.  But it seems kind of untenable for each crate that uses slog to export its own `slog`.  The idea is to thread a common logger from the top-level through various dependencies.  That can't work if they're not using the same `slog`.

Tangentially: I've long thought about putting the `ConfigLogging` (and related LogContext stuff from `test_util`) into a separate crate because we use that in a bunch of places that we don't otherwise need Dropshot.  That wouldn't help that much here though because we're still going to wind up with `ServerBuilder` accepting a logger.

hawkw:
> Hm. I don't think this would be bad. But it seems kind of untenable for each crate that uses slog to export its own `slog`. The idea is to thread a common logger from the top-level through various dependencies. That can't work if they're not using the same `slog`.

If I understand correctly, I think the goal here is to make it _easier_ to ensure that everyone is using the same `slog` crate version. The idea is that rather than having an explicit dependency on `slog` in the top-level application's `Cargo.toml`, which must be the same version of `slog` as the one that `dropshot` depends on (and which breaks the build if, say, renovate/dependabot updates the app's `slog` dependency), one would *not* depend explicitly on `slog` in the app's `Cargo.toml` and would instead get slog from `use dropshot::slog::...`, so that one does not have to independently manage a `slog` version.

[...]

PR #1610, opened 2026-05-22
logging: do not panic in the event of a failure to log a message

hawkw:
This is a quite sad situation. Currently, the logger configuration produced by `dropshot::ConfigLogging` will panic whenever an error writing a log line occurs. This crashes the entire process, which is often very much not what you want to have happen. Therefore, this branch changes the behavior to ignore errors returned here, resulting in the log line being silently eaten. This is *also* a behavior that I will be the first to admit is quite sad, but I believe it to be a less sad default than crashing the entire process.

As an example of why, consider the Oxide rack's sled-agent. The sled-agent will currently crash when it attempts to log something and the root filesystem to which it attempts to log is full, as described in oxidecomputer/omicron#4354. This is quite bad, as if we were to implement some way of freeing up additional storage space in such a situation, it is likely the sled-agent that would be responsible for doing so!

In an ideal world, we might wish for different behavior based on the particular error that occurred. We may wish to panic on errors to serialize a log line (as that would be indicative of a programmer error), while we may wish to not panic on I/O errors. Sadly, slog's current interfaces make this challenging, as all errors from the loggers we use, are presently coerced to `std::io::Error`, making it difficult to distinguish between I/O and serialization errors in a non-flaky way.

In the future, we might want to change the behavior from *silently* dropping log lines to instead recording when we have done so, perhaps by maintaining a count of logging errors. This could be broken down by error kinds, and track the timestamp of the last time a log message was lost. Reporting could report these error counts through other means, such as timeseries metrics, would allow the operator of the service to at least *know* that their log is incomplete. We could even imagine having a thingy that tracks if we have lost log messages, and once a log message is written successfully, tries to log a "hey, by the way, we also dropped however many log lines over the last however long!".

We might also want to make the panciking behavior configurable. This is sadly somewhat more annoying than it ought to be, as `slog::Drain::fuse()` and `slog::Drain::ignore_res()` change the _type_ of the drain, requiring duplicate code paths that construct almost-but-not-entirely identical loggers. But, this would let us have a config which, say, panics in tests but does not panic in production, which seems like a reasonable thing to want.

Nonetheless, this commit takes the shortest path to not panicking, and just turns all the `fuse()`s into `ignore_res()`es. I think the present behavior is bad enough in production that the quick fix feels fairly justified. We should consider making this nicer in the future though.

### DIFF
--- a/dropshot/src/logging.rs
+++ b/dropshot/src/logging.rs
@@ -158,13 +158,13 @@ where
             // as a full disk.  Although it is quite sad to silently eat a log
             // message, it is much sadder to crash the entire process,
             // particularly when the process is what is ultimately responsible
-            // for remediating the circumstances that causes logging to fail.
+            // for remediating the circumstances that cause logging to fail.
             //
             // In an ideal world, we might like some way of tracking a count of
             // how many log messages have been dropped, and for what reason.
             //
             // We may also wish to configure slog to panic if it attempts to log
-            // something that it cannot serialze, but not to panic if it
+            // something that it cannot serialize, but not to panic if it
             // encounters an I/O error.  Sadly, slog's interfaces for error
             // handling make this difficult, as everything is getting coerced to
             // `std::io::Error` anyway, even if it is actually a serialization
--- a/dropshot/src/logging.rs
+++ b/dropshot/src/logging.rs
@@ -149,27 +149,27 @@ where
     T: slog::Drain + Send + 'static,
     <T as slog::Drain>::Err: std::fmt::Debug,
 {
-    let level_drain = slog::LevelFilter(drain, Level::from(level))
+    let level_drain = slog::LevelFilter(drain, Level::from(level)).ignore_res();
+    let async_drain = slog_async::Async::new(level_drain)
+        .chan_size(1024)
+        .build()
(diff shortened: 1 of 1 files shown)


hawkw:
(as an aside, I feel like the fancier potential solutions, such as tracking logger errors, are probably best saved until the `ConfigLogging` stuff finds a new home in its own repo, as @davepacheco proposes in https://github.com/oxidecomputer/dropshot/issues/1607#issuecomment-4489803971)

davepacheco:
We've been discussing this in chat a bit and I don't think came to a conclusion but I figured I should note it here.  My current feeling is that this introduces new problems (we can be silently missing logs at any time) and doesn't meaningfully improve the situation it's intended to (it's hard to see the system running out of disk space _not_ becoming a major incident), and there are other relatively straightforward interventions with minimal downside that _would_ significantly improve this situation (e.g., oxidecomputer/omicron#7227).  I'm not positive about my skepticism here.  But it does make me very nervous to think that all future log files could be missing log entries with no indication about it.  Perhaps that feeling alone also warrants not doing this as a silent, non-configurable behavior change.

ahl:
Decided in the control plane meeting today not to proceed in this direction.

PR #1693, opened 2026-09-02
ci: GitHub Actions hygeine and MSRV 1.88

notpeter:
- Explicitly specify permissions for all GitHub Actions workflows
- Update Swatinem/rust-cache to v2.9.2 from v2.9.1
- Update actions/checkout to v7.0.1 (add comment with version)
- Add comment for actions/setup-node (`# v7.0.0`)
- Remove dtolnay/rust-toolchain dependency (`rustup` is in base runner)
- Switch from using `rustup default` to using explicit `RUSTUP_TOOLCHAIN`,
  as the former is overridden by the existence of a `rust-toolchain.toml`
- Add `CARGO_INCREMENTAL=0` to match the previous behavior provided by dtolnay/rust-toolchain. 
- Run `trybuild` as a dedicated ubuntu job, rather than as part of the matrix because rustc output differs between (stable, msrv).
  I think it's ok to not run this on every platform as the rustc output should be (has been) identical between platforms and if not, it's a rustc bug not a dropshot issue.
- Update MSRV to 1.88 (from 1.85)
- Tweak job names so they won't change when MSRV changes (allows stable repo config). E.g.
  Before: `build-and-test (macos-14, all, 1.85)`
  After: `build-and-test (macos-14, all, msrv)`
  Unchanged: `build-and-test (macos-14, all, stable)`

> [!NOTE] 
> Until this change we were not actually testing under our specified MSRV (1.85)
> Which meant we didn't realize that our MSRV was actually 1.88.

rust-not185

See example [run output](https://github.com/oxidecomputer/dropshot/actions/runs/33425722062/job/99598673902#step:5:10)

PR #1697, opened 2026-09-03
ci: GitHub Actions workflow improvements

notpeter:
- Supersedes https://github.com/oxidecomputer/dropshot/pull/1693
- Explicitly specify permissions for all GitHub Actions workflows
- Update Swatinem/rust-cache to v2.9.2 from v2.9.1
- Update actions/checkout to v7.0.1 (add comment with version)
- Pin actions/setup-node to a SHA (`# v7.0.0`)
- Drop dtolnay/rust-toolchain
- Add `CARGO_INCREMENTAL=0` to match the previous dtolnay/rust-toolchain behavior

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -58,8 +58,7 @@ jobs:
   build-and-test:
     runs-on: ${{ matrix.os }}
     env:
-      # Override rust-toolchain.toml, which otherwise takes precedence over the
-      # toolchain installed below.
+      # Override rust-toolchain.toml using explicit RUSTUP_TOOLCHAIN
       RUSTUP_TOOLCHAIN: ${{ matrix.toolchain }}
     strategy:
       matrix:
@@ -99,9 +98,8 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-      # Deliberately no RUSTUP_TOOLCHAIN here: the toolchain is pinned by
-      # rust-toolchain.toml, since the trybuild fixtures assert exact rustc
-      # diagnostics against committed .stderr files.
+      # toolchain is pinned by rust-toolchain.toml which is required as the trybuild
+      # fixtures assert exact rustc diagnostics against committed .stderr files.
       - name: Install Rust toolchain from rust-toolchain.toml
         run: rustup toolchain install
       - name: Report cargo version
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -101,11 +101,9 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       # Deliberately no RUSTUP_TOOLCHAIN here: the toolchain is pinned by
       # rust-toolchain.toml, since the trybuild fixtures assert exact rustc
-      # diagnostics against committed .stderr files. `rustup toolchain install`
-      # with no argument installs the toolchain named by that file; `rustup
-      # show` no longer installs it implicitly as of rustup 1.28.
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -93,17 +96,17 @@ jobs:
...
+      CARGO_INCREMENTAL: 0
     steps:
-      - uses: actions/checkout@0c366fd6a839edf440554fa01a7085ccba70ac98
-      # rust-toolchain.toml overrides dtolnay/rust-toolchain@stable, so in some
-      # sense using it is pointless. But it does a few other useful things such
-      # as disable incremental compilation, so we use it anyway.
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      # Deliberately no RUSTUP_TOOLCHAIN here, uses rust-toolchain.toml to pin version.
+      # trybuild fixtures assert exact rustc diagnostics in committed .stderr files.
```
sunshowers: I'd rather keep things as they are (I think turning off incremental compilation is important, and the color is a nice bonus as well).

notpeter: ~~Cool. I've reverted back to `dtolnay/rust-toolchain`.~~

notpeter: So here's actually my issue with `dtolnay/rust-toolchain` -- it actually makes implicit a bunch of things that would be better off to be explicitly defined.

Your comment here was incorrect:
https://github.com/oxidecomputer/dropshot/blob/8ef467ecca3eaa173214166567042e28e1f1cc4c/.github/workflows/rust.yml#L96-L101

See: [yesterday's trybuild job run on main](https://github.com/oxidecomputer/dropshot/actions/runs/33687675791/job/100438908110)
Screenshot 2026-09-04 at 10 01 24

In fact `dtolnay/rust-toolchain` never respects `rust-toolchain.toml`.  It requires specifying an explicit `toolchain` in your workflow, but when you pin to `dtolnay/rust-toolchain@stable` branch it automatically uses stable.  If you attempt to pin to a commit hash in main it shows its true colors and says `toolchain` is required. [job link](https://github.com/oxidecomputer/dropshot/actions/runs/33879715459/job/101045106088). [source link](https://github.com/dtolnay/rust-toolchain/compare/master...stable).

Sadly, the way this implemented also prevents you from pinning to explicit `stable` commits even if you want that behavior.  While the `v1` tag moves it points to commits in master (which is pretty common for GitHub Actions). But `stable` is implemented as a divergent branch.  Every time there is a release it is rebased to master and any previous commit becomes orphaned and subject to being garbage collected.

And these behaviors are intentional:
> Re: rust-toolchain.toml support
> I would prefer not to support it in this action. But someone else could maintain a different action that provides that pair of steps as one action.
> [dtolnay comment](https://redirect.github.com/dtolnay/rust-toolchain/issues/133#issuecomment-2591318675)

> In a workflow that pins the action using a full-length commit SHA (as opposed to something like @nightly or @1.89.0) it is required that you pick a SHA that is within the history of the master branch. Any commit that is not within the history of master will eventually get garbage-collected and your workflows will fail.
> [dtolnay/rust-toolchain README.md](https://github.com/dtolnay/rust-toolchain)

To me this sort of complexity is not worth it.  `CARGO_INCREMENTAL: 0` was the only env var we were getting implicitly from `dtolnay/rust-toolchain`, `RUST_BACKTRACE` and `CARGO_TERM_COLOR` were already being explicitly set:

https://github.com/oxidecomputer/dropshot/blob/8ef467ecca3eaa173214166567042e28e1f1cc4c/.github/workflows/rust.yml#L12-L15

sunshowers: Hmm, okay, thanks -- makes sense. (Though I think my comment agrees with what you said about precedence?)

